PR #119, opened 2018-04-07
Bug fixes to bootstrap.ps1 and other minor improvements

oabdellatif:
bootstrap.ps1 had a bug related to paths which resulted in an exception when trying to download any files. The error message was also very vague; the InnerException contained the specific details necessary to diagnose the first issue, so error handling was modified to look at the InnerException first. Build dependency versions were also updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 To build from source you have two choices.
 
-Automatic (mosty):
+Automatic (mostly):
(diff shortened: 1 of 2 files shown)


TanninOne:
good job! It's very possible the paths only worked correctly when running the script from the command line  so that the current working directory was set correctly.

PR #128, opened 2018-04-11
Minor bug fixes and overall polish

oabdellatif:
There are a couple of UI elements like text boxes and separators that were too bright as they were, so they've been tweaked to be darker. Also, spacing between buttons and labels in some places was too small and has been increased. CSS for hovering over tabs was implemented where there was none, which fixed the ugly white outlines when your cursor was on a tab. Finally, capitalization in different parts of the interface was made more consistent; for example, most buttons and actions in general are in title case, so buttons that weren't were changed to match. A bug fix or two have also tagged along for the ride.


locales/de/common.json
```diff
@@ -43,7 +43,7 @@
   "Beta": "Beta",
   "Add": "Hinzufügen",
   "Number of Mods enabled": "Anzahl aktiver mods",
-  "Search for games": "Nach Spielen suchen",
+  "Search for Games": "Nach Spielen suchen",
```
TanninOne, line 46: Sorry, but that's not correct. Games would only be capitalised at this point if it was a title, but it's not, it's a tooltip/button label.

locales/de/common.json
```diff
@@ -91,9 +91,9 @@
   "Creation Time": "Zeitpunkt d. Speicherns",
   "Delete": "Löschen",
   "Browse": "Suchen",
-  "Meta server": "Metaserver",
+  "Meta Servers": "Metaserver",
   "Servers to query for meta data.": "Server, auf denen nach Metadaten gesucht wird.",
-  "Add a metaserver": "Metaserver hinzufügen",
+  "Add a Meta Server": "Metaserver hinzufügen",
```
TanninOne, line 96: again, I don't think the capitalisation in the english text is correct.

locales/de/common.json
```diff
@@ -102,7 +102,7 @@
   "Disable Dependent": "Abhängige Mods deaktivieren",
   "Confirm deletion": "Löschen bestätigen",
   "Confirm": "Bestätigen",
-  "Third-party libraries": "Bibliotheken von Drittanbietern",
+  "Third-Party Libraries": "Bibliotheken von Drittanbietern",
```
TanninOne, line 105: Same here. Third-party was correct.

src/extensions/firststeps_dashlet/index.ts
```diff
@@ -23,7 +23,7 @@ function init(context: IExtensionContext): boolean {
     extTodos.push({ id, icon, type, props, condition, text, value, action, priority });
   };
 
-  context.registerDashlet('ToDo List', 2, 2, 150, Dashlet, state => {
+  context.registerDashlet('To Do List', 2, 2, 150, Dashlet, state => {
```
TanninOne, line 26: ToDo, To-Do or To-do are correct. "To Do" I believe is not.

src/extensions/gamemode_management/views/GamePicker.tsx
```diff
@@ -232,11 +232,11 @@ class GamePicker extends ComponentEx<IProps, IComponentState> {
                   <IconButton
                     id='start-discovery'
                     icon={discovery.running ? 'stop' : 'search'}
-                    tooltip={discovery.running ? t('Stop search') : t('Search for games')}
+                    tooltip={discovery.running ? t('Stop Search') : t('Search for Games')}
```
TanninOne, line 235: Same as before. "search" and "games" should not be capitalised in this spot.

src/extensions/mod_management/views/CheckModVersionsButton.tsx
```diff
@@ -41,7 +41,7 @@ class CheckVersionsButton extends ComponentEx<IProps, {}> {
         <ToolbarIcon
           id='check-mods-version'
           icon='refresh'
-          text={t('Check for mod updates')}
+          text={t('Check for Mod Updates')}
```
TanninOne, line 44: capitalisation

src/extensions/settings_interface/SettingsInterface.tsx
```diff
@@ -144,7 +144,7 @@ class SettingsInterface extends ComponentEx<IProps, IState> {
                 checked={customTitlebar}
                 onToggle={this.toggleCustomTitlebar}
               >
-                {t('Custom Window Titlebar')}
+                {t('Custom window titlebar')}
```
TanninOne, line 147: I can get behind _this_ change but it's inconsistent with everything else you corrected. Either button labels are "titles" or they aren't. If they are, this needs to be upper case. If they aren't, "Fix now" was correct.

PR #132, opened 2018-04-16
Completed currently translatable strings except...

dertrutinator:
Left out the following:

1. 

Mod names, as extension-translate currently adds them to _common.json_ like so:
`"Installing Better Dialogue Controls v1_2-27371-1-2": "Installing Better Dialogue Controls v1_2-27371-1-2"`
and
`"Better Dialogue Controls v1_2-27371-1-2 installed": "Better Dialogue Controls v1_2-27371-1-2 installed"`,
but they should probably be replaced using something like `{{modname}}`.

2.

`"Drop {{ accept }}": "Drop {{ accept }}"`, as I couldn't find that string anywhere in the UI and I don't know what it  means without context.

3.

I have also chosen not to translate the "Failed to initialize translation" message and the explanation of this message, as they will not show up in a translation anyway.

Regarding the _Feedback_ menu: I recommend asking users to write their reports in English, at least for those languages no one on the development team speaks. I have not added something like that for the German localisation, because @TanninOne speaks it, but I will add it if asked to do so.


M4lik:
`"Drop {{ accept }}": "Drop {{ accept }}"` is coming from a Dropzone (Drag & Drop). The `{{ accept }}` part gets replaced by `URL or File`  when the user is NOT "holding" a File or URL atm.

locales/de/common.json
```diff
@@ -274,6 +269,131 @@
...
+  "or": "oder",
+  "Attach Special File": "Spezielle Datei anhängen",
+  "System Information": "Systeminformationen",
+  "Vortex Log": "Vortex-Log",
+  "Application Settings": "Anwendungseinstellungen",
+  "Application State": "Status der Anwendung",
+  "Send anonymously": "Anonym versenden",
+  "Submit Feedback": "Feedback versenden",
+  "feedback-instructions": "Bitte<1></1><2><0>verwende Satzzeichen und Zeilenumbrüche,</0><1>sei präzise und themabezogen. Du musst keine ganzen Sätze bilden. Ein Bugreport ist ein technisches Dokument, keine Prosa,</1><2>melde nur einen Fehler pro Nachricht,</2><3>stelle keine eigenen Vermutungen auf, melde einfach, was Du siehst und was Du erwartet hast,</3><4>benutze, wenn möglich, Beispiele, wie man den Fehler nachbilden kann. Auch bei einem generellen Problem(\"Mod-Installer verwenden dieses Feature, wenn sie jenes verwenden sollten\"), beschreibe eine Möglichkeit, das Problem sichtbar zu machen.</4></2>Die meiste Zeit wird üblicherweise gebraucht, wenn wir versuchen, einen Fehler zu reproduzieren, um ihn fixen zu können. Je weniger Zeit wir darauf verschwenden, desto mehr Zeit haben wir, um tolle neue Features zu programmieren!",
```
dermalikmann: I personally would add something like "Bitte schreibe dein Feedback in Englisch", because although Tannin can speak it, other people who might read the Messages can not.

TanninOne: Good point, I may not be the only person reading that feedback

Rummskartoffel: In that case, how do I use those xml(?) tags in order to add another line to the list in `feedback-instructions`?

locales/de/common.json
```diff
@@ -395,5 +394,6 @@
   "Finish": "Fertigstellen",
   "Checking dependencies": "Überprüfe Abhängigkeiten",
   "Description": "Beschreibung",
-  "Write your own notes on this mod": "Mache Dir eigene Notizen zu dieser Mod"
+  "Write your own notes on this mod": "Mache Dir eigene Notizen zu dieser Mod",
+  "Drop {{ accept }}": "Hier {{ accept }} ablegen"
```
dermalikmann: I would rather suggest "{{ accept }} hier ablegen"

locales/de/common.json
TanninOne: Hmm, I'm not sure you can tbh.
In the original text those are html tags (br, ul, li) but the i18next library seems to translate them.
I wasn't able to find any documentation how that works but it appears the tags are mapped to numbers in the order they appear in the original which means adding a tag may not be possible at all, because it wouldn't know what html tag it's supposed to be.
I'll see if we can influence that mapping somehow

Rummskartoffel: I mean, we could just put it there in the English version as well. Or just translate it to an empty string in the English *commons.json*.

TanninOne: Yes, we could, though tbh. I've not received a single non-english feedback anyway so this at least seems to be self-explanatory to most. I'd still like to understand what's going on with the translation lib and html as this may get in our way in the future.

dermalikmann: My suggestion would be to add a new `<li>` in the feedback source code and then map the new element in the json file.

dertrutinator:
So what do we do with this now? I have tried @M4lik\'s [change](https://github.com/Nexus-Mods/Vortex/pull/132/commits/82f520c62a9abe20ac81cfbe5c26c0b509f19af0) to _common.json_ with 0.14.1 and it was visible, although without a bullet at the start of the line, so it _kind of_ works even without changing anything in _extension-feedback_. Or should I undo it, so that this can be merged?

PR #1026, opened 2018-09-17
Fixed vortex api failing to build on linux.

TerryRPatterson:
Fix failure during webpack build on non-windows platforms by replacing app.getPath("Home") with os.homedir.

### DIFF
--- a/src/util/Steam.ts
+++ b/src/util/Steam.ts
@@ -5,11 +5,11 @@ import * as fs from './fs';
 import { log } from './log';
 import { getSafeCI } from './storeHelper';
 
-import { app as appIn, remote } from 'electron';
 import * as path from 'path';
 import { parse } from 'simple-vdf';
 
-const app = (remote !== undefined) ? remote.app : appIn;
+
+import {homedir} from 'os';
 
 export interface ISteamEntry {
   appid: string;
@@ -75,7 +75,7 @@ class Steam implements ISteam {
           }
         });
     } else {
-      this.mBaseFolder = Promise.resolve(path.resolve(app.getPath('home'), '.steam', 'steam'));
+      this.mBaseFolder = Promise.resolve(path.resolve(homedir(), '.steam', 'steam'));
     }
   }
 


TanninOne:
Hi! Just to keep you updated: I'm not ignoring this pull request, I just feel like this change shouldn't be necessary.

I don't quite understand why it happens because whether or not app is defined will only be decided at runtime and it should never actually be undefined (app being a core part of electron).
It appears that for some reason webpack is convinced, at build time, that app will be undefined and tries to optimise away code, but that would be a mistake.

I think a better solution can be found, because app.getPath is used a lot. If it _were_ unavailable, nothing would work anyway. If it isn't unavailable, this change shouldn't be necessary.

codeman101:
@TanninOne 

I took at this commit myself out of interest. I agree with the request because although both methods achieve the same result the requested code is a simplification of the original code.

TanninOne:
As I said, app.getPath is used all over Vortex for different directories (not all can be accessed through other node libraries), if it doesn't work and webpack optimizes it away, this will break the entire application not just this one location.
And I find it very hard to believe that, because electron is primarily developed and tested by MacOS users, it's more likely for stuff to break on Windows that work on other platforms than the other way around.

The fact that it's one line less doesn't make it a simplification because it would be inconsistent with the rest of the application.

So until someone can explain (and ideally demonstrate) to me why&how this breaks on other platforms, I'm not merging it.

IDCs:
Hi @TerryRPatterson, I'm afraid this fix will not be merged in. As Tannin mentioned, we are currently using the app.getPath function throughout Vortex. Fixing it in the Steam script is simply not sufficient.

codeman101:
Ever since the vortex install script on lutris I don't really see the point of this anyway anymore since vortex is only used for Windows games.

@TanninOne I'm sorry for how I used to be months ago it should have clicked with me that making a native version of the app would not make sense since the games it would interact with are Windows only. I'm glad a lutris script was created for it. I now use it myself for Skyrim through lutris.

TanninOne:
Well, I wouldn't really say this is the case, there are a few supported games that exist natively for Linux, like Factorio, its simply that we can't (currently) afford the time to work on a native port ourselves.

PR #1908, opened 2018-10-30
Did a bunch of German translations again

dertrutinator:
I left out a few ones I couldn't translate without proper context as well as those for translate.json, as those are pointless.


TanninOne:
Thanks a lot!

PR #5352, opened 2019-08-30
feat: chinese

SettingDust:
Based on liuyanxi's text.

I notice there is some text cannot be translate.

Such as
```
Announcements
News
Changelogs
ToDo List
```

How can I translate them


TanninOne:
Hi! Please report those cases to this issue: https://github.com/Nexus-Mods/Vortex/issues/303
We will go through it at some point and make them translatable, for now unfortunately nothing can be done.

And please publish your translation here: https://www.nexusmods.com/site/mods/
In the next larger update users will be able to find those translations from inside Vortex so it will be more convenient but for now we don't want to include translations in the main Vortex application (the german one is basically a sample)

PR #5619, opened 2019-10-30
added french translation

obelouix:


TanninOne:
Thank you for your contribution!
But please upload translations to https://www.nexusmods.com/site , we don't bundle translations with the main repo (the german one was intended as a sample and will be removed from the repo as well).

We will soon implement an easier way for users to find and install extensions from the site.

PR #5796, opened 2019-12-05
Small fix to radio buttons

ffgriever-pl:
This is a small fix for radio buttons rendered within `showDialog` window, so that they do not render on top of each other if there is not enough space, but rather maintain usual vertical separation and show scrollbars instead. Checkboxes already make use of this fix (`overflow: auto` for divs within `.dialog-content-choices`).

### DIFF
--- a/src/views/Dialog.tsx
+++ b/src/views/Dialog.tsx
@@ -244,7 +244,9 @@ class Dialog extends ComponentEx<IProps, IComponentState> {
     } else if (content.choices !== undefined) {
       controls.push((
         <div key='dialog-content-choices' className='dialog-content-choices'>
-          {content.choices.map(this.renderRadiobutton)}
+          <div>
+            {content.choices.map(this.renderRadiobutton)}
+          </div>
         </div>
       ));
     } else if (content.input !== undefined) {


TanninOne:
Which dialog did you have this problem with? I feel like this is probably better fixed in css. As far as I can tell this change simply makes the css for .dialog-content-choices ineffective.

ffgriever-pl:
I've fixed it this particular way only because checkboxes had the exact same fix implemented earlier, so I've chosen the consistency over anything else.

I didn't see it in any message displayed by the vortex itself, but I can easily generate a dialog that causes the issue. I've run into this problem when working on an extension (I'm using "checkboxes" and "choices" as the simplest way to give a user a choice directly in showDialog).


```
      showDialog(
        'question',
        "test window", {
        bbcode: "test message",
        choices: choices,
      },
        [{ label: 'Install' }, { label: 'Cancel', action: () => dismiss() }]);
```
I used bbcode because it makes it easier to spot the issue, as there is much less vertical space left than when using simple text. `choices` contains simply 20 generated objects.

Here is the result when using radials (no patch):
screen1

Here is the result when using checkboxes (no patch):
screen2

Here is the result when using radials (with patch):
screen3

TanninOne:
Ok, I agree, this is consistent with the checkboxes, didn't see that before.

Thank you for the contribution!

PR #6102, opened 2020-04-17
Updated the "Open on Nexus Mods" icon

Pickysaurus:
This PR uses the Nexus Mods icon, but the "open external" icon may also be a good fit.

Ideally something to imply you're opening it externally and not using an icon that already appears in the menu 
image

### DIFF
--- a/src/extensions/nexus_integration/index.tsx
+++ b/src/extensions/nexus_integration/index.tsx
@@ -804,7 +804,7 @@ function guessIds(api: IExtensionApi, instanceIds: string[]) {
 
 function init(context: IExtensionContextExt): boolean {
   context.registerAction('application-icons', 200, LoginIcon, {}, () => ({ nexus }));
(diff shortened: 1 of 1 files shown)


Pickysaurus:
"open-in-browser" is the name of the alternative suggestion.

src/extensions/nexus_integration/index.tsx
```diff
@@ -804,7 +804,7 @@ function guessIds(api: IExtensionApi, instanceIds: string[]) {
 
 function init(context: IExtensionContextExt): boolean {
   context.registerAction('application-icons', 200, LoginIcon, {}, () => ({ nexus }));
-  context.registerAction('mods-action-icons', 999, 'open-ext', {}, 'Open on Nexus Mods',
+  context.registerAction('mods-action-icons', 999, 'nexus', {}, 'Open on Nexus Mods',
```
IDCs, line 807: "open-in-browser" is the name of the icon you're looking for

PR #6224, opened 2020-05-04
Build overhaul and refactor

mgziminsky:
Cleaned up the build process to DRY out the configurations, remove dead code, and simplify the custom build scripts. I've tried to split the commits up into specific manageable chunks, and highly recommend reviewing the individual commits rather than all at once.

## Summary of main changes
- switch to a single package.json setup
- single electron-builder config file 
- extract and import common configuration for the webpack configs
- minimize the copying of static files to different locations for builds. Instead use them in place whenever possible
- go back to using `postinstall` to handle native dep rebuilding, but use both `install-app-deps` and `electron-rebuild` for node-sass case
- cleaned up dependency tree using `yarn-deduplicate -s fewer`. Nice speed boost to installs

## Why?
Well, my original reason for doing this was a weird error I was getting in my production builds that seemed to imply something weird was happening with the build process. In the end, it was unrelated, but I've already done the work, and I believe the changes are big improvement in terms of maintainability and cognitive load.

## Verification
All the tests pass, and i've done some basic usage of the app in dev and prod. I didn't notice anything broken and the dev console doesn't show any errors in prod. In dev, there is a single bad reference to the roboto font somewhere during opening the settings for the first time that I just couldn't track down, but it isn't an issue in prod, and doesn't appear to break anything in dev.

### PS
One last thing. The only thing currently preventing this entire project from building with python 3 and vs2019 is the missing imports I mentioned in https://github.com/Nexus-Mods/node-permissions/pull/1 and adding `"resolutions": { "node-gyp": ">=5.0.0" }` to `package.json`. I'm currently building and running just fine using local patched clones of the direct deps from that list in the PR. I had sorta hoped you would update them all in one go since forking and submitting 10 PRs is a bit of a PIA for me...


TanninOne:
First of all: Thank you for the contribution.
As you may imagine this is a fairly significant change that I certainly won't merge into the 1.2.x cycle of versions.
So far I've only skimmed you're changes and identified a bunch of things that I'd like to use, others that I can't include, many where I can see the purpose but replacing a working and tested system just to prettify the build system makes me anxious.
E.g. that "flaky rebuild" has worked perfectly fine for two years and probably  over a hundred releases, is it really worth touching to save a few lines of code. I found both postinstall and electron-rebuild to be notoriously problematic in the past and postinstall in particular works differently between npm and yarn (at least it did when we switched) so I'd rather have a more manual solution that has been tested.
Same goes for the two package.json setup. Yes, it may be non-ideal to run a script to test for conflicts between them but it's working and we did many successful builds with it. For us it's more important that our releases go out with few issues for users than to make the build system easy to get into for external devs. Sorry, but that's simply the truth. With hundreds of thousands of users who could have a terrible update experience if any of this breaks, I'd rather be conservative (never touch a running system).

We are considering moving away from electron-builder long term because unfortunately it has at least one horrible bug that doesn't get fixed and as a result of that we'll have to reevaluate the build scripts and everything anyway. As such, this PR may be open for a long time and likely we will cherry pick individual commits at different times

mgziminsky:
That's understandable, thanks for at least looking. If you want, I can break out the parts you want to merge into a separate PR. 

I understand not wanting to mess too much with a working system. I'm not familiar with your build/release process or dev environment, but the thing that prompted me to make most of these changes was because I was having actual issues building a working app from a fresh checkout. My reasoning for a couple of the larger changes is below in case it is useful.

The module rebuilds, node-sass specifically, was one of the biggest issues I was having. From a clean checkout, running `yarn install` followed by `yarn dist` would result in a node-sass module runtime error. I couldn't find a consistent way to fix it, and once I somehow got it working, it would break again as soon as I updated and had to run `yarn install` again. When deciding to make that particular change, I did read through the history and see you had tried using the `postinstall` at one point. When I tried just doing `install-app-deps` by itself as you did, I had the same issues with node-sass not building you did. I dug into it a bit, and it appears to be because some sort of pre-build check that node-sass does that decides a build isn't necessary. Using electron-rebuild forces it to build somehow, and since setting the `postinstall` to run both, I haven't once had issues with the native module versions.

I switched to the single package.json setup because I was getting a react error saying it could be caused by having multiple version of react. Making this change was a last resort for me after weeks of frustration with this particular error. It turned out that this change didn't fix it, but while doing research and investigation leading up to it, everything said that using a split setup like this is discouraged now. Since it makes things cleaner and I had it working, I decided to leave it. Regarding the issue I was having, it was because I was attempting to run the app directly from the unpacked dist directory, which for some reason breaks react. I don't know if this is expected or if it's just something with my setup, but any attempt to run from a subdirectory of the project causes react to throw a fit. Even running it via a different path using symlinks or junctions doesn't work if the true path is nested in the project. My guess is react or something is doing some sort of path resolution then walking the parents looking for node_modules or something for some reason. Interestingly, if I place the unpacked contents outside the source tree and link to it from inside, things do work, and the build won't remove the linked directory, just its contents when building.

TanninOne:
Yeah, node-sass or native modules in general are a total pita, however in the past we had problems with postinstall.  For development builds (when you start Vortex from the ide with the environment variable NODE_ENV=development set) Vortex will rebuild the module on the fly on startup if they fail to load, for releases we create fresh build where it works fine the way it currently is. Putting the rebuild into postinstall would probably cause us problems during development, where native modules frequently get deleted by npm/yarn when you upgrade a module.

> I switched to the single package.json setup because I was getting a react error saying it could be caused by having multiple version of react.

This wouldn't have anything to do with the dual package.json setup. The dual setup simply means that we use a different package for release builds than for development, they are never combined. The react error means that different modules (maybe an extension?) bundles a react version that is different from the one used in Vortex.
The way it's currently set up, without any modifications or foreign extensions, I don't  see how Vortex could be loading multiple react instances, the require cache should ensure only one gets loaded.

> everything said that using a split setup like this is discouraged now

do you have a source/reasons for that?

> Regarding the issue I was having, it was because I was attempting to run the app directly from the unpacked dist directory, which for some reason breaks react. I don't know if this is expected or if it's just something with my setup, but any attempt to run from a subdirectory of the project causes react to throw a fit.

The way node.js require/import works (this is not specific to Vortex) is that it caches modules based on the resolved file name of the module, case sensitive.
Meaning that if you require('react') from two different locations (say from core vortex and once from an extension) and once it resolves to "c:\project\vortex\node_modules\react" and the next time it resolves to "c:\PrOjEcT\vortex\node_modules\react" because for whatever reason the module search algorithm goes through a junction point the one time and through the original path the next, node.js will create a second instance of react and even though it's the same exact version in the same physical location, that is enough to trigger react, because react uses the require cache to implement a single-instance prerequisite.
This is an inconvenience in node.js because they don't consider case-insensitive file systems, but from their perspective it's just a cache so it's no biggy if the same module gets loaded twice. They explicitly say it can't be relied upon for singleton behavior.

It's a proper bug in react because they use this cache to implement singleton behavior and their perspective is Homer
Which is to say:  Yes this is f*ed up, I did report it, they're not fixing it. So basically anything that could confuse the node.js module resolution, like links, can mess up react and the only real solution is to avoid them.
To be fair: react isn't alone in abusing the require-cache for singleton behavior.

PR #7752, opened 2020-10-22
Add internal nxm:// handling

basxto:


TanninOne:
@basxto What problem does this solve?

basxto:
With that [Builtin Mod Browser](https://www.nexusmods.com/site/mods/88) can handle `nxm://` links on vortex instances, that are not the standard handler for `nxm://`. That can be handy for a development version existing alongside a stable version. Or when it can’t be set as the standard handler yet.

Related to https://github.com/MaverickMartyn/builtin-mod-browser/issues/6

TanninOne:
Makes sense, thank you for the explanation.

PR #12003, opened 2022-06-03
fix “check for mod updates”

ExE-Boss:
Fixes GH-11994

--------------------------------------------------------------------------------

Note that I did not change the type of `IMod.attributes` to `IModAttributes` as that caused way too many type issues (so the type of `IMod.attributes.downloadGame` is still `any` for now).

### DIFF
--- a/src/extensions/mod_management/types/IMod.ts
+++ b/src/extensions/mod_management/types/IMod.ts
@@ -33,6 +33,14 @@ export interface IMod {
   fileOverrides?: string[];
 }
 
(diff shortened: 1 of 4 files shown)


TanninOne:
@ExE-Boss thank you for the contribution! I'm not really happy typing it like that because attributes.downloadGame is supposed to be a string, cases where is an array are bugs where they were inserted incorrectly.
I will introduce more checks to prevent invalid data to be inserted there in the first place

PR #12069, opened 2022-06-14
Visual indicator for running background processes

mxswat:


src/views/NotificationButton.tsx
```diff
@@ -124,12 +124,28 @@ class NotificationButton extends ComponentEx<IProps, IComponentState> {
       });
     }
 
+    const isRunningBgProcess =
```
TanninOne: Please don't adopt nonsense designer language into the code. A process is a well established term in programming and this has nothing to do with processes.

src/views/NotificationButton.tsx
```diff
@@ -124,12 +124,28 @@ class NotificationButton extends ComponentEx<IProps, IComponentState> {
       });
     }
 
+    const isRunningBgProcess = 
+      notifications
+      .filter((x) => x.type === 'activity' && x.progress === undefined)
+      .length >= 1 && combinedProgress.length == 0;
```
TanninOne: The way this is wrapped is super hard to read.
Also, it's needlessly complicated. If "combinedProgress.length" is 0 then you already know there is no activity notification with a progress. all the filter has to look for is type === activity. And then I would advice putting the check for combinedProgress.length first, that's the cheaper (performance-wise) condition to check.
Not very relevant here but a good habit to get into.

src/views/NotificationButton.tsx
```diff
@@ -124,12 +124,28 @@ class NotificationButton extends ComponentEx<IProps, IComponentState> {
...
+    
+    if (isRunningBgProcess) {
```
TanninOne: This is quite hacky, I really don't like it.
The look of the radial progress indicator is implemented by RadialProgress and _its_ css.
That's a fairly decent, reusable component.

The style of the "activity" indication you now have as part of the notification button and its css, making it not reusable and the styling is spread out over multiple files (the filled out ring being 1/4 of the circle is here, the animation speed in the css)
Please clean this up, make this spinner part a feature of the RadialProgress component

src/views/NotificationButton.tsx
```diff
@@ -124,12 +124,28 @@ class NotificationButton extends ComponentEx<IProps, IComponentState> {
...
+    }
+
     return (
       <div style={{ display: 'inline-block' }}>
         <Button id='notifications-button' onClick={this.toggle} ref={this.mButtonRef}>
           <Icon name='notifications' />
           <RadialProgress
-            className='notifications-progress'
+            className={
+              `${isRunningBgProcess ? 'notifications-spin ' : ''}notifications-progress`
```
TanninOne: The pattern we usually use is to construct an array of css classes and then have className={classes.join(' ')}
that is much easier to read and less error prone than string concatenation like this.

TanninOne (changes requested):

src/controls/RadialProgress.tsx
```diff
@@ -38,8 +45,17 @@ class RadialProgress extends React.Component<IProps, {}> {
   }
 
   public render(): JSX.Element {
-    const { className, data, offset, style, totalRadius } = this.props;
+    const { className, data, offset, style, totalRadius, spin } = this.props;
     const sideLength = (totalRadius + (offset || 0)) * 2;
+
+    if (spin) {
```
TanninOne: seems needlessly copy&paste, apart from the classname and data this is the same code as in the default case.

TanninOne: I'm also fairly certain this is a bug now. Now the "spin" flag takes precedence over the case where we actually know the progress.

src/controls/RadialProgress.tsx
```diff
@@ -38,8 +45,17 @@ class RadialProgress extends React.Component<IProps, {}> {
...
+      return (
+        <svg className={`notifications-spin ${className}`} viewBox={`0 0 ${sideLength} ${sideLength}`} style={style}>
+          {spinData.map(this.renderArc)}
```
TanninOne: Also: Why would you have spinData as an array if you don't intend to reuse the existing code? It's static data, you _know_ it's a single element

src/stylesheets/vortex/notification.scss
```diff
@@ -119,6 +119,12 @@
     width: 100% !important;
     height: 100% !important;
   }
+  .notifications-spin {
```
TanninOne: This doesn't belong to notifications, it should be in radial.scss

src/controls/RadialProgress.tsx
```diff
@@ -38,8 +45,17 @@ class RadialProgress extends React.Component<IProps, {}> {
   }
 
   public render(): JSX.Element {
-    const { className, data, offset, style, totalRadius } = this.props;
+    const { className, data, offset, style, totalRadius, spin } = this.props;
     const sideLength = (totalRadius + (offset || 0)) * 2;
+
+    if (spin) {
+      return (
+        <svg className={`notifications-spin ${className}`} viewBox={`0 0 ${sideLength} ${sideLength}`} style={style}>
```
TanninOne: It's still string concatenation and also: why is it still called **notifications**-spin?

src/views/NotificationButton.tsx
```diff
@@ -124,13 +124,18 @@ class NotificationButton extends ComponentEx<IProps, IComponentState> {
...
+    );
+
     return (
       <div style={{ display: 'inline-block' }}>
         <Button id='notifications-button' onClick={this.toggle} ref={this.mButtonRef}>
           <Icon name='notifications' />
           <RadialProgress
             className='notifications-progress'
             data={combinedProgress}
+            spin={pendingActivities.length >= 1}
```
TanninOne, line 138: As I said above: the spin flag now takes precedence over data, meaning that if we have notifications that have actual progress information we're not showing it.

TanninOne (changes requested):

PR #12088, opened 2022-06-17
Draggable tools

IDCs:


src/extensions/starter_dashlet/BoxWithHandle.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { CSSProperties, FC, memo } from 'react';
+import { useDrag } from 'react-dnd';
+import Icon from '../../controls/Icon';
+import { StarterInfo } from '../../util/api';
```
TanninOne, line 4: please don't import "api" files from the core application. There are at least two occurrences of this in this diff.

src/extensions/starter_dashlet/BoxWithHandle.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface IProps {
+  children?: any;
+  className?: string;
+  item: StarterInfo;
+}
+
+export const BoxWithHandle: FC<IProps> = memo((props: IProps) => {
```
TanninOne, line 12: I may be missing something here, what's the point of memo-izing this function?

src/extensions/starter_dashlet/Tools.tsx
```diff
@@ -130,17 +134,21 @@ class Starter extends ComponentEx<IStarterProps, IWelcomeScreenState> {
...
           <Media.Body>
-            {this.renderToolIcons(game, discoveredGame)}
+            <FlexLayout type='column'>
+              <FlexLayout type='row' className='starter-dashlet-tools-header'>
+                <h1>{t('Tools')}</h1>
+                <Toggle
+                  checked={addToTitleBar}
+                  onToggle={this.onToggle}
+                >
+                  {t('Enable toolbar')}
```
TanninOne: So afaict the "toolbar" is still provided by a "dynamic" extension whereas this is part of the base application. That's really not how extensions are supposed to interact.

src/extensions/starter_dashlet/BoxWithHandle.tsx
```diff
@@ -0,0 +1,32 @@
...
+      return {
+        isDragging: monitor.isDragging(),
+        opacity: monitor.isDragging() ? 0.4 : 1,
+      };
+    },
+  });
+  const children = Array.isArray(props.children)
+    ? props.children : [props.children];
+  return (
+    <div className='starter-drag-handle-container' ref={dragPreview} style={{ opacity }}>
```
TanninOne: "BoxWithHandle" is a fairly generic name and this component seems to be quite generic but these class names are very specific

src/extensions/starter_dashlet/Tools.tsx
```diff
@@ -445,17 +383,24 @@ class Starter extends ComponentEx<IStarterProps, IWelcomeScreenState> {
   }
 
   private makePrimary = (starter: StarterInfo) => {
```
TanninOne: this name doesn't seem accurate any more as the function now seems to act more like a toggle, right?

src/extensions/starter_dashlet/types.ts
```diff
@@ -0,0 +1,15 @@
+import { IDiscoveredTool } from '../../types/IDiscoveredTool';
+import { IRunningTool } from '../../types/IState';
+import { IDiscoveryResult } from '../gamemode_management/types/IDiscoveryResult';
+import { IGameStored } from '../gamemode_management/types/IGameStored';
+export interface IConnectedProps {
```
TanninOne: Not a huge fan of the naming here. We have a very generic interface name (IConnectedProps) in a super generic file (types.ts) that is intended for exactly one specific component. Makes the code hard to understand as there is no easy way to make the connection

src/extensions/starter_dashlet/useConnectedProps.ts
```diff
@@ -0,0 +1,30 @@
...
+import { activeGameId } from '../profile_management/selectors';
+
+import { useTranslation } from 'react-i18next';
+import { useSelector, useStore } from 'react-redux';
+import { getSafe } from '../../util/storeHelper';
+import { IConnectedProps } from './types';
+
+const emptyObj = {};
+
+export const useConnectedProps = (): IConnectedProps => {
```
TanninOne: Same problem as above. "useXYZ" implies a hook function, a piece of reusable functionality. But this isn't reusable, it's specifically for one component. I don't think moving this away from the Component class it's intended for makes the code better, on the contrary.

src/stylesheets/vortex/starter.scss
```diff
@@ -36,10 +59,33 @@
...
   }
+
+  .starter-drag-handle-container {
+    display: flex;
+    flex-direction: row;
+    margin-right: .5rem;
+    background-color: transparent;
+    .starter-drag-handle {
+      background-color: $brand-bg;
+      min-width: 20px;
```
TanninOne: pixel sizes

src/stylesheets/vortex/starter.scss
```diff
@@ -36,10 +59,33 @@
...
+      height: auto;
+      display: flex;
+      cursor: move;
+      border-right: 1px solid $brand-menu;
+      .starter-drag-handle-icon {
+        padding: 3px;
```
TanninOne: pixel sizes

TanninOne (changes requested):

PR #12612, opened 2022-08-04
do not hardcode thumbnail dimensions

alistair3149:
Came across this while working on a Vortex theme.

**Issue:**
The max dimensions of the thumbnail is hardcoded, and will break if the container size is changed.

**Solution:**
Instead of hardcoding the max dimensions, the image should adapt to the container size.
Using `object-fit: cover` will achieve the goal.
https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

### DIFF
--- a/src/stylesheets/vortex/gamepicker.scss
+++ b/src/stylesheets/vortex/gamepicker.scss
@@ -1,4 +1,5 @@
 $thumbnail-width: 256px;
+$thumbnail-height: 144px;
 
 #gamepicker-layout {
   text-align: right;
@@ -8,6 +9,7 @@ $thumbnail-width: 256px;
   display: grid;
   justify-content: center;
   grid-template-columns: repeat(auto-fill, $thumbnail-width + $gutter-width);
+  grid-auto-rows: $thumbnail-height + $gutter-width;
 
   > .game-thumbnail {
     margin: $half-gutter !important;
--- a/src/stylesheets/vortex/gamepicker.scss
+++ b/src/stylesheets/vortex/gamepicker.scss
@@ -112,8 +112,9 @@ $thumbnail-width: 256px;
   }
 
   .thumbnail-img {
-    max-width: 256px;
-    max-height: 144px;
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
   }
 }
 
--- a/src/stylesheets/vortex/gamepicker.scss
+++ b/src/stylesheets/vortex/gamepicker.scss
@@ -1,5 +1,4 @@
 $thumbnail-width: 256px;
-$thumbnail-height: 144px;
 
 #gamepicker-layout {
   text-align: right;
@@ -9,7 +8,6 @@ $thumbnail-height: 144px;
   display: grid;
   justify-content: center;
   grid-template-columns: repeat(auto-fill, $thumbnail-width + $gutter-width);
-  grid-auto-rows: $thumbnail-height + $gutter-width;
 
   > .game-thumbnail {
     margin: $half-gutter !important;
@@ -115,7 +113,7 @@ $thumbnail-height: 144px;
 
   .thumbnail-img {
     width: 100%;
-    height: 100%;
+    aspect-ratio: 16 / 9;
     object-fit: cover;
   }
 }


TanninOne:
But this then leads to inconsisting sizing of the thumbnails:

game_thumbnails_mess

Keep in mind that most games are supported through third party extensions, the images for those come from the site so we can't rely on data to be consistent. It's much easier to enforce consistent sizing in the client through css than to try to get all extension authors to size the images properly.
So I'd imagine you'd at least want to fixate the size of the container (game-thumbnail-body) but I kind of think it would be more flexible for themes if we have the container be sized relative to a normalized image size (which can still be overridden) than have the image forced to take up the entire panel size and have the container decide how big the image is.

TanninOne:
@alistair3149 is that reply satisfying and are you ok with me closing this PR or did I miss something?

Absolutely love the fluent theme btw, it's what I use for my own setup now. :+1:

alistair3149:
> But this then leads to inconsisting sizing of the thumbnails:
> 
> game_thumbnails_mess
> 
> Keep in mind that most games are supported through third party extensions, the images for those come from the site so we can't rely on data to be consistent. It's much easier to enforce consistent sizing in the client through css than to try to get all extension authors to size the images properly. So I'd imagine you'd at least want to fixate the size of the container (game-thumbnail-body) but I kind of think it would be more flexible for themes if we have the container be sized relative to a normalized image size (which can still be overridden) than have the image forced to take up the entire panel size and have the container decide how big the image is.

I agree with setting the size in the container through CSS. My end goal is to make the style more maintainable. In this case, using only the container to define the sizes instead of having to set it through both the image and the container, as that's two separate rules to maintain. I'll revise the PR with the changes.

> @alistair3149 is that reply satisfying and are you ok with me closing this PR or did I miss something?
> 
> Absolutely love the fluent theme btw, it's what I use for my own setup now. 👍

Thank you for the kind words! As I mentioned above, I'll revise the PR hopefully in the coming days and see if the solution is sufficient. If not, we can see how to tackle it.

alistair3149:
I realized that the `.game-group` grid already define the width of the thumbnail. Instead of using the container, the grid can be used to define the height of the thumbnail instead. @TanninOne what do you think of the solution?

TanninOne:
Sorry to be difficult but I believe this change would be make theming this screen less flexible.
256x144 is the intended image size, 16/9 the aspect ratio.
If we specify the size on the grid or on .game-thumbnail or something, this means themes _have_ to have the image fill the entire tile, otherwise the image gets distorted.
Imagine you wanted the tiles to look something like this:

game_tile_border

Where the size of the image doesn't match the size of the tile, neither in absolute size nor aspect ratio.
If the grid had a hard-coded width and height the image has to have the same aspect ratio as it's parent or it will look borked.

What do you think of dropping the grid-auto-rows and then in .thumbnail-img instead of specifying the height we use
```
width: 100%
aspect-ratio: 16 / 9;
object-fit: cover;
```

That is: we wouldn't specify the height at all, it would be derived from the width and scale correctly. The image would be then always have the correct aspect ratio and as long as the source image has the correct ratio, it would always show fully. If it's not the correct ratio, at least it would always show the same area, it wouldn't show a different region of the image depending on the theme.

alistair3149:
> Sorry to be difficult but I believe this change would be make theming this screen less flexible. 256x144 is the intended image size, 16/9 the aspect ratio. If we specify the size on the grid or on .game-thumbnail or something, this means themes _have_ to have the image fill the entire tile, otherwise the image gets distorted. Imagine you wanted the tiles to look something like this:
> 
> game_tile_border
> 
> Where the size of the image doesn't match the size of the tile, neither in absolute size nor aspect ratio. If the grid had a hard-coded width and height the image has to have the same aspect ratio as it's parent or it will look borked.
> 
> What do you think of dropping the grid-auto-rows and then in .thumbnail-img instead of specifying the height we use
> 
> ```
> width: 100%
> aspect-ratio: 16 / 9;
> object-fit: cover;
> ```
> 
> That is: we wouldn't specify the height at all, it would be derived from the width and scale correctly. The image would be then always have the correct aspect ratio and as long as the source image has the correct ratio, it would always show fully. If it's not the correct ratio, at least it would always show the same area, it wouldn't show a different region of the image depending on the theme.

No worries. It's insightful to learn the rationale of the original implementation and solution assumed that the text are within the thumbnail.

Setting aspect ratio is more versatile, it'll still allow the size of the thumbnail be changed through the grid only so it's perfect. Updating the PR now.

PR #12650, opened 2022-08-09
allow re-installing required dependencies if fomod options unfulfilled

IDCs:


src/extensions/mod_management/InstallManager.ts
```diff
@@ -2391,7 +2391,9 @@ class InstallManager {
                     .then(() => dep['reresolveDownloadHint']())
                     .then(() => doDownload(dep));
                 })
-            : Promise.resolve(dep.mod.id);
```
TanninOne, line 2394: This is fine but I would probably put this as a feature into findModByRef. Adding an optional parameter to findModByRef so the caller can pass in installer options so you could call the function going "find me the mod with this reference but only if it used these installer options".
If we want to elevate installer options as an "identifying criterium" of mods we should have that as part of the api.

PR #12702, opened 2022-08-13
use Roboto Flex VF instead of Roboto regular

alistair3149:
**Issue:**
Font weights are not rendering correctly for Roboto.
Since only Roboto regular is included as the font, only the 400 (normal) font is rendered correctly, 600 (bold) is simulated and does not render nicely. This is especially problematic as the default heading font weight in Vortex is 500, which is only rendered as 400 so it looks the same as regular body text.

**Solution:**
By using a [variable font](https://web.dev/variable-fonts/), more specifically [Roboto Flex](https://fonts.google.com/specimen/Roboto+Flex), any font-weight between 100 to 1000 will be rendered correctly.
With the additional benefits of having more control over font styles, and not needing to maintain multiple files.

Although this fix is specifically for Roboto because of the heading issue, I also suggest to do the same for Montserrat because of the benefits of having less maintenance and more flexibility.


TanninOne:
I forwarded this to the UI department but they decided to not go with the variable font option but instead to bundle further roboto variants.

PR #13109, opened 2022-09-17
begin linux packaging targets

NicBOMB:
- apply 'space savings' per-target-platform excludes all prebuilds unless it's for the target platform https://github.com/Nexus-Mods/Vortex/commit/f300cdd1fc1758c4986dfd029f8a9dbae0efc75d
- fix assets/images/vortex.png resaved vortex.ico as vortex.png since vortex.png was too small for node_modules/app-builder/pkg/icons/icon-converter.go (requires 256x256)
- fix 7z-bin linux exclusion to only include -> (basic files and module info, bin, AND per-target-platform binary)


TanninOne:
@NicBOMB node-native-errors is another one of those packages that you don't want to load at all on non-windows platform, it only exists because the way node maps native errors to "platform independent" error codes is so bad that they are borderline useless on windows.

regarding libxmljs: Vortex doesn't actually use that library any more because it's been a growing pain in the ass for us.
The only reason we still ship it is for the possibility that a third-party extension may be built on it, since we don't want to break backwards compatibility.
There is a fairly good reason you could drop it from the linux build and no one would care.

PR #13285, opened 2022-10-09
import Bluebird from 'bluebird', not Promise(BB)

NicBOMB:
This commit refactors all import statements for 'bluebird' to Bluebird.
Some import statements used PromiseBB or Promise previously.
This should help with consistency and further code cleanup.

An unused import statement for 'bluebird' was removed entirely from each of `src/control/DraggableList.tsx` and `src/extensions/nexus_integration/texts.ts`.
An additional unused import statement was also removed from `src/extensions/nexus_integration/texts.ts`.
[It hadn't been used for a while.](https://github.com/Nexus-Mods/Vortex/commit/b7db2f2b61c8f643c1ed638c2fc62bc3dbd08ee0#diff-f71910ff257547dd19dce2ee2027c5c2cf208b0a828c764e5521f61829461b04)

### DIFF
--- a/src/actions/notifications.ts
+++ b/src/actions/notifications.ts
@@ -6,7 +6,7 @@ import {truthy} from '../util/util';
 
 import safeCreateAction from './safeCreateAction';
 
-import Promise from 'bluebird';
+import Bluebird from 'bluebird';
 import { ipcMain, ipcRenderer } from 'electron';
 
 import * as reduxAct from 'redux-act';
@@ -120,7 +120,7 @@ export function addNotification(notification: INotification) {
     const noti = { ...notification };
 
     if ((noti.id !== undefined) && (suppressNotification(noti.id))) {
-      return Promise.resolve();
+      return Bluebird.resolve();
     }
 
     if (noti.id === undefined) {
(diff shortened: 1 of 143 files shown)


thetric:
As a longer term goal, Vortex should migrate to built-in Promises as suggested by Bluebird:

> ## ⚠️Note⚠️ 
>
>**Please use native promises instead if at all possible**. Native Promises have been stable in Node.js and browsers for around 6 years now and they have been fast for around 3. Bluebird still offers some useful utility methods and you can use it - but please consider native promises first.
>
> This is a good thing, the people working on Bluebird and promises have been able to help incorporate most of the useful things from Bluebird into JavaScript itself and platforms/engines. There are still missing things (.map/.filter are on their way with the iteration helpers proposal and async iterators!).
>
> If there is a feature that keeps you using bluebird. Please let us know so we can try and upstream it :)
>
>Currently - it is only recommended to use Bluebird if you need to support old browsers or EoL Node.js or as an intermediate step to use warnings/monitoring to find bugs.
>
> https://github.com/petkaantonov/bluebird

thetric:
Ah, thanks :) I'm just pocking around a bit

PR #13444, opened 2022-11-04
Added an option to "fix" the detected path for a game store

Pickysaurus:
Some games (Fallout 3, Oblivion, Morrowind) include multiple copies of the game at the top level so the path we detect will not be correct. Added an optional modifyPath option that will allow game extensions to fix the path this feature finds.

### DIFF
--- a/src/util/GameStoreHelper.ts
+++ b/src/util/GameStoreHelper.ts
@@ -23,6 +23,7 @@ export interface IStoreQuery {
     id?: string;
     name?: string;
     prefer?: number;
+    modifyPath?: (detectedPath: string) => string;
   }>;
 }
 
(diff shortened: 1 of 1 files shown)


Pickysaurus:
This applied to older Bethesda games on Epic games or Xbox game pass.

TanninOne:
@Pickysaurus Please clarify what this is good for. How can the game directory contain multiple copies of the game? And why doesn't the discovery function just return the correct path in the first place?

Pickysaurus:
The older Bethesda games install all locales as separate instances in Xbox/Epic installs. This applies to Fallout 3/Fallout New Vegas/Oblivion/Morrowind. 


image
image

Pickysaurus:
Fix rejected

PR #13961, opened 2023-02-05
new localization (cs)

f456ew:
https://www.nexusmods.com/site/mods/139?tab=description

by Wol0rd


TanninOne:
Your localization was added to the extension list, we don't merge localizations into the repo.

PR #14935, opened 2023-11-27
FBLO updated layout and styles

insomnious:


src/stylesheets/vortex/page-mod-load-order.scss
```diff
@@ -107,14 +167,80 @@
...
+      }
+
+      // bottom box (validation console)
+
+      > :last-child {
+        //background-color: #892be26b;
+        flex: unset; // shrinks to whatever space is needed, preferably none
+        max-height: 50%; // don't ever take up more than 50% of height
+
+        background-color: #2a2c2b;
```
IDCs: This should be set to a variable so that themes are applied correctly to it.

IDCs (approved):

PR #15044, opened 2023-12-20
Support updating mods with pre-release versions

madtisa:
Let's say we have mod with current version `1.0.0-pre1` and newest `1.0.0-pre2`. Coercing would convert both versions to 1.0.0 and update notification won't occur


IDCs:
I agree with you completely - but I'm afraid that most of our mod authors (not to mention some game developers) will not abide to semantic versioning. This update will fix the update functionality for some games, yet break it for others.

Until the website is in a position to enforce versioning, I'm afraid that specific line of code will remain as is.

Issue #15334, opened 2024-03-14
Underscores in archive names causing issues with mod installation


insomnious:
> I obtained this mod from afkmods (2553-wip-shalidors-maze). I reported the issue to the mod author but he insists that it is a Vortex bug and he won't do anything about it.  He says that I should report it here.

> My workaround is to add the mod back in every time using "Install from file" in Vortex after I hit "Apply Changes" in the above dialog box.  I have attached the mod below for you to try.

> The other odd thing about this mod is that the title of it doesn't show up in Vortex; only the file name appears.  image.png.0043b85f374a1ad36c24d209e0483f8b.png

> Don't know if this is related to the disappearing mod issue or not.

> This mod is for Skyrim Special Edition.

https://forums.nexusmods.com/topic/13474412-mod-changed-on-disk-with-restart-of-vortex/

https://www.afkmods.com/index.php?/files/file/2553-wip-shalidors-maze/

PR #15349, opened 2024-03-18
fixed incorrect detection of removed mods on startup

IDCs:
We're now less restrictive with the mods we install. The previous detection mechanism was too restrictive when attempting to avoid using the vortex staging tag, which was causing mods with  the '__' prefix being ignored during mod installation status detection.

fixes nexus-mods/vortex#15334


insomnious (approved):
Done

PR #15651, opened 2024-05-08
fixed potential crash when queryPath functor returns function

IDCs:
queryPath should only return:
string | Promise\<string\> | Promise\<IGameStoreEntry\>


insomnious:
Fixes #15648

Issue #15439, opened 2024-04-03
Remove libxmljs dependency


insomnious:
# Chore

The `libxmljs` library is somewhat unmaintained, and has been causing us issues with our build system for years as it keeps dropping assets.

Currently we only use this library in the Bannerlord extension which we aim to replace/stub in the future.

We should remove this dependency so that it no longer blocks our build process.

## Outline
(Describe the task).

## Supporting Information
(Include any supporting screenshots etc to support the above).

## Checklist

- [ ] Include a description of the task.
- [ ] Include relevant department labels.
- [ ] Include any supporting information.

Thaodan:
How was libxmljs replaced? I'm trying to understand the change as libxmljs is causing issues for me while packaging Vector as shouldn't bundle native dependencies.

insomnious:
Deprecated in 1.13.0

PR #15818, opened 2024-05-28
Finally removed libxmljs!

IDCs:
We must inform extension authors that libxmljs is being deprecated before this becomes live!

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -50,7 +50,6 @@
     "jsonwebtoken": "^9.0.0",
     "leveldown": "^5.6.0",
     "levelup": "^4.4.0",
-    "libxmljs": "Nexus-Mods/libxmljs#sambacha",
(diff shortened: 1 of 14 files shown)


insomnious:
Before this is merged in and 1.12 is released, we need to let extension authors know that they will no longer be able to use libxmljs within their extension as we are removing it. Being removed because it's harder to get assemblies that electron needs for it to work correctly. We had already moved nearly everything to xml2js (which is actively maintained) except for the old bannerlord game extension. Bannerlord extension has been replaced for the new BUTR one and so libxmljs can be removed safely for our core app and bundled extensions.

We are going to have a deprecation phase where we ASAP get in touch with authors and write scripts to find what extensions are using it. Changelog will mention it's being deprecated ASAP too.

insomnious:
Fixes #15439

PR #16169, opened 2024-08-05
Node events memory leak fixes

IDCs:
The rest of the leaks are within the collections extension; still trying to figure out how to deal with that without rocking the boat too much.

The functionality for the main app is still solid though.

- Ensures that the symlink activator is no longer "stuck" behind the elevation notification.
- Better error handling throughout the application, React errors will no longer block the entire page view when something fails to render, it will block the specific component that failed to render (which can indeed be a whole page in some cases)
- Dynamic divs are now wrapped in an ErrorBoundary and will provide a way for users to report any rendering issues that occur.
- General improvements to the DOM's consistency (removed anti-patterns) - this might screw up the stylesheets a bit (specifically for ECD)
- Ensured that we add and remove listeners correctly in the QuickLauncher and MainWindow components.

Please ignore until 1.13.x is worked on

### DIFF
--- a/src/controls/DynDiv.tsx
+++ b/src/controls/DynDiv.tsx
@@ -5,6 +5,8 @@ import { extend } from '../util/ExtensionProvider';
 import * as React from 'react';
 import { IExtensibleProps } from '../types/IExtensionProvider';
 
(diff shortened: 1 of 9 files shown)


IDCs:
No longer needed

Issue #16231, opened 2024-08-12
fallout london nexus mods update button in vortex goes to the mod ID of the FOLON mod but on the FO4 nexus


HerptyDerpoty:
**Describe the bug**
this button 
![image](https://github.com/user-attachments/assets/1265b235-47de-465d-9d98-d5da23e802b2)
 the update button redircts every mod, even ones marked 
![image](https://github.com/user-attachments/assets/76daa5be-e217-4b1b-9d93-931e5f5f170e)
to the regular FO4 nexus.

![image](https://github.com/user-attachments/assets/eaf8fb59-f81e-4bce-ab8a-8e506518cdeb)
this button (on the right) however does correctly go to the FOLON Nexus page for the mod

**To Reproduce**

1. Find a folon mod that needs updating
2. Click the update icon next to the version number

**Expected behavior**
Opens the mod on the correct nexus (FOLON nexus not FO4 nexus)

**Platform (please complete the following information):**
Vortex 1.12.1.0

HerptyDerpoty:
i should also add that it does correctly identify the mod for everything else, IE the update check is actually checking the mod on the FOLON nexus, its _just_ that the update button next to the version number takes you to the FO4 nexus.

HerptyDerpoty:
ok actually looking at it the above is not correct, it seems like its doing some super inconsistent behavior for checking if mods need updates or not if its a https://www.nexusmods.com/fallout4london/ mod.

PR #16237, opened 2024-08-13
fixed inability to update mods that are tagged as compatibleDownloads

IDCs:
FO4 for example, provides the ability to download FOLON mods.

The update check mechanism was attempting to check for an update using the FO4 domain name regardless of the domain name used when the mod was initially downloaded.

This would obviously result in Vortex installing a different mod as an update..

closes nexus-mods/vortex#16231


HerptyDerpoty:
Dunno if this is the place but this fixed _newly_ added mods that have an update when you check but if you already checked for an update the cached info doesn't get replace so you're stuck with a broken update button on folon mods

PR #16696, opened 2024-11-25
Improved error handling for history revert events

IDCs:
Suppressed ability to report history revert events when Vortex encounters a file system error (permissions/missing file/etc)

fixes nexus-mods/vortex#16694


IDCs:
This feels like a band-aid fix for now, as the history revert event should've never reported the file system error in the first place.

IDCs:
Closing as this doesn't fix the underlying issue. Additionally the error which this attempts to fix has not been seen since 1.12.6

Issue #17465, opened 2025-04-29
Tool won't run from the dropdown menu


ze0rs:
**Tool won't run from the dropdown menu**
Nothing happens when clicking on `Run` item on the menu. However, clicking on the button icon can run the tool successfully.

**To Reproduce**
1. Add a tool to any managed game
2. Click on the vertical 3-dot button
3. Click on the "Run" item
4. Nothing happens

**Expected behavior**
The tool should run after clicking on the "Run" menu item.

**Platform (please complete the following information):**
 - OS: Windows 11
 - Vortex Version: 1.13.7

PR #17466, opened 2025-04-29
Bugfix-17465: Fix ToolButton run menu item

ze0rs:
This PR is to fix issue #17465 where the `Run` menu item of the tool button won't start the tool.


IDCs:
Thank you for your contribution!

PR #17612, opened 2025-06-05
Adding ability to update all nexus sourced mods in one go

### LINKED ISSUE #17381: [FEATURE] - "Install All" Button for Mod Updates
**Is your feature request related to a problem? Please describe.**
It would be great to have a button to update all mods with available updates that can be downloaded directly in Vortex (cloud icon). Currently, the user must click to download each update individually. If the user does this too quickly, though, it can sometimes trigger "External Changes" popups because it seems that Vortex cannot keep up when downloading and installing multiple updates at once. The only way to avoid that issue is to disable "Deploy on Mods Enabled", which is a change from the default settings and not desirable in a lot of situations (i.e. that setting breaks FBLO mod position preservation).  

**Describe the solution you'd like**
Add a button that would have Vortex sequentially download and install (enable and deploy, based on user settings) all available mod updates for the active game profile. This would of course only apply to update files that are properly named and versioned (cloud icon). 

This could also enable a second button - "Check for Updates and Install", where Vortex runs the update check and downloads and installs mod updates automatically as it finds them. 

This feature is a somewhat glaring issue with Vortex, and a common user complaint. Other modern mod managers, such as R2ModMan from ThunderStore, can batch-update mods in this fashion for the user with a single click. 

**Describe alternatives you've considered**
Currently, the only alternative is to click one update download at a time. But you can't do it too fast because that can cause "External Changes" popups to occur.

IDCs:
- The check for updates drop down menu now has a new "Apply All Updates" entry which will run a full update check and install those updates in sequence.
- Other update check entries will raise a notification with an "Update All" button once the check completes.

Closes #17381

### DIFF
--- a/src/extensions/mod_management/InstallManager.ts
+++ b/src/extensions/mod_management/InstallManager.ts
@@ -1618,37 +1618,59 @@ class InstallManager {
   private queryIgnoreDependent(store: ThunkStore<any>, gameId: string,
                                dependents: Array<{ owner: string, rule: IModRule }>)
                                : Bluebird<void> {
+    const batchKey = 'remember-ignore-dependent-action';
+    let context = getBatchContext('install-mod', '', false);
+    const handleAction = (action: string, remember: boolean) => {
+      if (remember) {
+        context = getBatchContext('install-mod', '', true);
+        context?.set?.(batchKey, action);
+      }
+      if (action === 'Cancel') {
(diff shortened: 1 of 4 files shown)


ChemGuy1611:
Amazing! This is a huge improvement. Thanks for working on this 🙏

PR #17841, opened 2025-07-02
fixed file based override instructions not being applied correctly

IDCs:
previously the check for whether the user is actively installing a mod through a dependency was too strict and could cause the override instructions to never apply.

improved overall detection of whether the mod is being installed as a dependency, which intentionally does not allow override instructions


IDCs:
A better solution has been implemented in a different PR

PR #18016, opened 2025-08-05
Expanded tracking for collections

clarkey022:
Includes tracking for mod and collections enabled numbers at game launch.


src/views/QuickLauncher.tsx
```diff
@@ -238,6 +238,12 @@ class QuickLauncher extends ComponentEx<IProps, IComponentState> {
       return;
     }
     this.context.api.events.emit('analytics-track-click-event', 'Header', 'Play game');
+    const state: IState = this.context.api.store.getState();
+    const numberOfMods = Object.keys(state.persistent.mods[starter.gameId]).length;
```
IDCs: This is the total number of mods, not the total number of **enabled** mods. To retrieve the correct value, you would have to pull the active profile and filter based on the modState of each modId

IDCs (approved):

PR #18158, opened 2025-08-28
consolidated update functionality into a single button

IDCs:
- Premium users are able to update all mods in one click
- Removed pointless "full update check" notification.
- Simplified mods-update handler


ChemGuy1611:
This is a massive upgrade to the Vortex user experience. 

Thanks for working on this!

PR #18407, opened 2025-09-23
18399 remove notifications

IDCs:


insomnious:
I've checked 

- [x] remove notification that asks to resume if collection is removed
- [x] warning notifications about dependencies still around after a collection installed
- [x] remove windows system notifications - find ticket, toggle in settings disabled by default
- [x] dismiss notifications on game/profile change

PR #18721, opened 2025-11-04
fixed false positive download reference checks for fuzzy/bundled mods

IDCs:


IDCs:
closes nexus-mods/vortex#18719

PR #18841, opened 2025-11-20
Pull engine and package manager version from manifest

erri120:


windows_dev_setup.ps1
```diff
@@ -1,18 +1,29 @@
...
+# Installs: Git, Python, CMake, VS Build Tools + Windows SDK, NVM, Node, Yarn
 # Then clones/updates Vortex repository
 
 #Requires -RunAsAdministrator
 
+
 # Configuration
-$NODE_VERSION = "22.19.0"
+$NODE_VERSION = "22.19"
+$YarnVersion = "1.22.19"
```
Aragas, line 10: More of a discussion, but perhaps it's better to have a failfast approach here? If we're not able to extract the variables, fail instead of relying on in-script defaults

Aragas:
There's also a CI issue - `error vortex_devel@0.0.1: The engine "node" is incompatible with this module. Expected version "22.19". Got "22.14.0"`

Update 1:
It's because our workflows use Node `22.14`, hahaha

Update 2:
Perhaps we can also extract the node version for GitHub workflows from the package.json file?

windows_dev_setup.ps1
erri120: The script itself is weird as it can be used outside a cloned repo. Usually you'd clone the repo and then run some setup script inside the cloned repo for all the tools.

Aragas: Yea, see it now. Perhaps then we can first install git and clone the repo, then install Node? Not sure what @insomnious would think about this

PR #18887, opened 2025-11-26
Build with FOMOD on Linux

erri120:
Vortex now builds without issues and without any addition setup on Linux.


Aragas:
CI is failing, needs a fix

PR #18897, opened 2025-11-27
Fold submodules into main Monorepo

Al12rs:
Consolidate previously separate submodules into the main repository. 

Skipped submodules:
- `extensions/games`
- `extensions/fomod-installer`
- `api`

We tested:
- `yarn install`
- `yarn build`
- `yarn start`
- Added StardewValley and installed a collection.

We didn't test what CI does

EDIT:
- Removed `install-extensions.js` as `yarn build` should already take care of it inside `BuildSubprojects.js`


Al12rs:
Some of the submodule folders still have:
- `.github/workflows`
- `LICENSE`
- Empty `README.md`

I could go in and clean these up, potentially as a separate PR if that is preferrable.

Need to decide what to do with the CI workflows, if we want to import them at the top level, so maybe that can be a separate thing.

PR #18923, opened 2025-12-01
Port extensions workflows to monorepo

Al12rs:
Ported the following workflows from ex-submodules to the monorepo:

- `extensions/collections/.github/test.yml`
- `extensions/gamebryo-savegame-management/.github/main.yml`
- `extensions/mod-report/.github/node.js.yml`
- `extensions/modtype-bepinex/.github/node.js.yml`
- `extensions/mtframework-arc-support/.github/node.js.yml`

Also updated node.js version to use the one listed in the main `package.json` rather than using a hard coded version (16.x).

Some of these handled creation of artifact and releases on the old submodule repositories. 
That was removed in favour of just installing dependencies and building. 

## Questions:
1. Some of the old workflows were used to generate artifacts and release them on the respective repos. I'm assuming we no longer need that, would that be correct?

2. Only the collections workflow actually has tests, the other ones only do install and build. Are they still useful or should we remove them?

3. For now these only run on Windows, should we also run them on Linux?

### DIFF
--- a/.github/workflows/extensions-gamebryo-savegame-management-test.yml
+++ b/.github/workflows/extensions-gamebryo-savegame-management-test.yml
@@ -16,11 +16,7 @@ concurrency:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest]
-
-    runs-on: ${{ matrix.os }}
+    runs-on: windows-latest
 
     steps:
     - name: Checkout repository


Al12rs:
Ready for review

erri120:
I believe the only workflow worth keeping is the one for the collections extension which runs tests besides building. Since the other extensions are included in the normal build, we're doing double the work for nothing.

IDCs:
1. Correct, we no longer need the artifacts.
2. The build workflows are mainly used for early fault detection as the main repo's build script takes at least ~25 minutes to run (on Windows). I would keep the savegame-extension and collections workflows at the very least (savegame extension has a native module which has been known to fail regularly); other ones - I guess we can remove.
3. I think we should add linux workflows for the extensions that use native modules - currently just savegame and collections extensions.

@Al12rs

IDCs:
@insomnious - do we want to remove the regular build workflows? (apart from the collections and savegame ones)

Issue #18974, opened 2025-12-07
[BUG-ALPHA] Collections - Vortex Fails to Install Mods During Collection Downloads/Installs


VeryLastKiss:
**Your Nexus Mods username**
VeryLastKiss

**Describe the bug**
I'm on the latest beta version of Vortex (**v1.16.0-alpha.5**) and I'm trying to download my [Essential Foundation](https://www.nexusmods.com/games/newvegas/collections/caband) collection to test the alpha, but I'm getting stuck on these 4 mods for some unknown reason. I had to delete the archives of the mods multiple times until Vortex was finally able to properly install the mods.

<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/544d97a3-604f-4523-84c3-1aa04d9a7ae2" />

**To Reproduce**
1. Be on **v1.16.0-alpha.5** of Vortex.
2. Attempt to download my [Essential Foundation](https://www.nexusmods.com/games/newvegas/collections/caband) collection.

**Expected behavior**
All 4 of those mods in my screenshot are included in the default download and should automatically install.

**Platform (please complete the following information):**
 - OS: Windows 11
 - Vortex Version: **1.16.0-alpha.5**

[network.log](https://github.com/user-attachments/files/24017274/network.log)
[vortex.log](https://github.com/user-attachments/files/24017273/vortex.log)
[vortex1.log](https://github.com/user-attachments/files/24017278/vortex1.log)
[vortex2.log](https://github.com/user-attachments/files/24017277/vortex2.log)
[vortex3.log](https://github.com/user-attachments/files/24017276/vortex3.log)
[vortex4.log](https://github.com/user-attachments/files/24017275/vortex4.log)

PR #18986, opened 2025-12-08
Switched FOMOD to depend on .NET 9 isntead of self-contained

### LINKED ISSUE #18975: [BUG-ALPHA] Collections - The Mod Configuration Menu for Fallout New Vegas Often Fails to Install
**Your Nexus Mods username**
VeryLastKiss

**Describe the bug**
I'm on the latest beta version of Vortex (**v1.16.0-alpha.5**) and I'm trying to download my [Essential Foundation](https://www.nexusmods.com/games/newvegas/collections/caband) collection to test the alpha. A command prompt opens to install the Mod Configuration Menu, but the command prompt takes about 20-30 seconds to open the menu to install the mod, which is significantly slower than the public stable/beta versions of Vortex. The Mod Configuration Menu is also extremely small in size for a mod, so it should be nearly instantaneous. Furthermore, the Mod Configuration menu often fails to install entirely.

**To Reproduce**
1. Be on **v1.16.0-alpha.5** of Vortex.
2. Attempt to download my [Essential Foundation](https://www.nexusmods.com/games/newvegas/collections/caband) collection.

**Expected behavior**
The Mod Configuration Menu is a very small mod in size, and it should download and install nearly instantly.

 - OS: Windows 11
 - Vortex Version: **1.16.0-alpha.5**
 - 
[network.log](https://github.com/user-attachments/files/24017274/network.log)
[vortex.log](https://github.com/user-attachments/files/24017273/vortex.log)
[vortex1.log](https://github.com/user-attachments/files/24017278/vortex1.log)
[vortex2.log](https://github.com/user-attachments/files/24017277/vortex2.log)
[vortex3.log](https://github.com/user-attachments/files/24017276/vortex3.log)
[vortex4.log](https://github.com/user-attachments/files/24017275/vortex4.log)

Aragas:
Added back installer sandbox setting
Added awaitDotnetAssert call to make sure .NET 9 is installed when we run the installer

Might fix #18975 and #18974

Comment on the issue:  
> MCM is a trusted mod, so we don't invoke the Windows App Container sandboxing.
> The only thing that comes in mind is that we're distributing FOMOD via a self-contained single file. What it does is unzipping itself into a temp folder and then executing. Perhaps Windows Defender isn't liking it? Or something else is slowing it down?
> This way we also keep the executable within Vortex's folder

There's no point of doing self-contained without single-file and compression, since just self-contained will yield 157MB of dist size. Instead let us depend on the runtime Vortex provides, which gives us 17MB.  
I also don't want to do NativeAOT for compilation. Might work, but seems risky with trimming/compilation

Needs https://github.com/Nexus-Mods/fomod-installer/pull/29 to be merged first


Sewer56:
> What it does is unzipping itself into a temp folder and then executing. Perhaps Windows Defender isn't liking it?

This hasn't been the case for 5 years by default.
It was the case up until .NET 5, since then .NET 5 loads binaries from memory, sans native binaries.

(Due to no Windows API to load from memory, just `LoadLibrary`. There's still of course stuff like 'Manual Mapping', like in DLL Injection, but y'know)

[Details documented here](https://github.com/dotnet/designs/blob/main/accepted/2020/single-file/design.md#the-host)

For some unknown reason, checking the submodule referenced in this PR, the original FOMOD project *specifically opted into* extracting at startup, hurting boot time:
```xml
<IncludeAllContentForSelfExtract>true</IncludeAllContentForSelfExtract>
<IncludeNativeLibrariesForSelfExtract>true</IncludeNativeLibrariesForSelfExtract>
<IncludeRuntimeForSelfExtract>true</IncludeRuntimeForSelfExtract>
```

There isn't a reason you should do that with managed assemblies. The only exceptions is if you have code that checks for its own location, e.g. think `Assembly.Location`; with single file some of these APIs may return different results than the folder with the executable you're expecting.

If FOMOD was doing that- god knows why.

Sewer56:
Anyway, the history note aside; if we're shipping a runtime with Vortex; or installing it as part of Vortex install, then no point doing self contained.

Though *you can* do framework dependent + single file.
Just that for some benign reason nobody talks about it; and have not for the past 5 years (lol).

Aragas:
Thanks @Sewer56, those are good points!
Yea, it looks like frameworks dependent is just so simple there's no point on complicating things, especially since we don't plan on making the .NET 9 installer redundant

PR #19052, opened 2025-12-12
Fixed renderer script not being loaded

IDCs:
Hash changed, so renderer was not being loaded on app start.

(ignore the branch name, I wrongly assumed it's down to devtools initially)


erri120:
Why are we hardcoding script hashes in the first place? Hashes can be inserted by the build tool.

PR #19229, opened 2026-01-06
suppress ability to report UserCanceled errors during dependency install

IDCs:
fixes nexus-mods/vortex#19181


IDCs:
Linux built fine, which means Windows will too.

PR #19251, opened 2026-01-08
allow users to check for mod updates even when the mod is disabled

IDCs:
fixes nexus-mods/vortex#19250


erri120 (approved):

src/extensions/mod_management/views/ModList.tsx
Healz4Life, line 1: Thanks!
I respect you making this change off a request on discord instead of making me go through the 'official' request place that seems to slow down things making it too political of a process than helping directly make vortex better.
I am now seeing what I can do myself after seeing how this change was made!

PR #19303, opened 2026-01-14
Added: Support for opening files/directories on Linux

Sewer56:
## Summary

Fix file and directory opening on Linux by using `shell.openPath` for local paths instead of `shell.openExternal`.

To be more specific, we use `shell.openPath` for local paths (which in turn delegates to `xdg-open`) and keep using `shell.openExternal` for URI schemes like `nxm://` or similar.

## Changes

- Separate handling for URLs vs local file paths in `opn.ts`
- Use Electron's `shell.openPath` for local files/directories on non-Windows platforms
- Add path detection helpers to distinguish URLs from filesystem paths

## Purpose

The `shell.openExternal` API is supposed to be used for protocols, e.g. `nxm://`, the fallback we had however, used it for everything, if `ShellExecuteEx` was not available; this was technically incorrect.

Since within the codebase, this function is used for both files, and protocols, I haven't rewired that for the time being; as to avoid being too destructive. Instead, I updated the `else` case to handle both paths and protocols.

## Modified Files

- `src/util/opn.ts` - Route local paths through `shell.openPath` on Linux/macOS
- `package.json` - Update cross-env dependency

And this resolved my weird `file-roller` bug/quirk

- `flake.nix` - Update electron sandbox path
- `flake.lock` - Bump nixpkgs revision


src/util/opn.ts
```diff
@@ -68,12 +108,20 @@ function open(target: string, wait?: boolean): Promise<void> {
       }
     }
   } else {
+    const openPromise = isUrlTarget(target)
+      ? openExternal(target)
+      : openLocalPath(target);
+
+    openPromise.catch((err) => {
+      const error = (err as any)?.message ?? err;
```
erri120: Use `getErrorMessage` from `src/shared/errors`.

Sewer56: Wasn't in my branch, had to merge `master`.
Anyway, done in c91593c0dc15e457d1475b2b298219defbe9f6db

erri120 (approved):

0xra0:
It still doesn't work, it doesn't detect the directory if try to add the game manyally

Sewer56:
That could be using a different API, I haven't checked.
I fixed it for the case of right clicking the mods from the 'Mods' tab; and clicking `Open Folder` or `Open Archive`.

0xra0:
> That could be using a different API, I haven't checked. I fixed it for the case of right clicking the mods from the 'Mods' tab; and clicking `Open Folder` or `Open Archive`.

Ok

PR #19330, opened 2026-01-16
Collection Name character length set to 64

Healz4Life:
Changed MAX_COLLECTION_NAME_LENGTH from 36 to 64
Making it easier to fit a name that might list a few features. As shown need here.
Stim1
64 gives enough room to write Fidelity instead of Fix even :O
If there's a larger number that the devs want to use, feel free to make the change before merging! This is just all the character length I've found necessary so far.

### DIFF
--- a/extensions/collections/src/constants.ts
+++ b/extensions/collections/src/constants.ts
@@ -32,7 +32,7 @@ export const INSTALLING_NOTIFICATION_ID = "installing-collection-";
 
 // limits
 export const MIN_COLLECTION_NAME_LENGTH = 3;
-export const MAX_COLLECTION_NAME_LENGTH = 36;
+export const MAX_COLLECTION_NAME_LENGTH = 64;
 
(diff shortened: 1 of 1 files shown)


IDCs:
@insomnious @halgari,

I've checked with the BE team - there is no string limitation set for the collection name so the DB should handle it just fine. Vortex should handle anything below 68 characters within its current components.

I can't think of a reason to limit to 36 characters, bar forcing curators to find a meaningful name for their collection without adding functionality specific information (e.g. "INCLUDES XNVSE!" or whatever)

Edit: IMHO - 36 characters is usually enough for 4 words, 7 characters long each (excluding whitespace) - that should be enough.

IDCs:
After some internal discussion - we decided against raising the character limit. Collection names should be short and concise - it's not meant to depict features, that's what the collection's description is for.

Healz4Life:
last thought: it doesn't even take up two lines fully on the card display

PR #19353, opened 2026-01-19
Set shell to true if not explicitly set when we detect common script extensions

Aragas:
Related - #19127


Aragas:
Yea, I didn't want to add .vbs, since it's usually a malware. Not even sure when was the last time I've seen a legit .vbs script

PR #19363, opened 2026-01-20
Add eslint rule to fix bluebird Promise alias imports

erri120:
Adds a new custom eslint rule to prevent importing bluebird as `Promise` and fixes all occurrences in the repo. This doesn't change any runtime behavior, it just renames the import from `Promise` to `PromiseBB` so that it doesn't alias over the global `Promise`.

### DIFF
--- a/eslint-rules/no-bluebird-promise-alias.mjs
+++ b/eslint-rules/no-bluebird-promise-alias.mjs
@@ -7,7 +7,7 @@ export default {
     },
     messages: {
       renamePromise:
-        "Do not import Bluebird as 'Promise'. Rename to 'PromiseBB'",
+        "Do not import Bluebird as 'Promise'. Rename to 'Bluebird'",
     },
     schema: [],
   },


Pickysaurus:
Yes please! Bluebird removed entirely would be a great step 2!

IDCs:
> Yes please! Bluebird removed entirely would be a great step 2!

This will never happen unless the community extensions stop using bluebird.

Aragas:
We could polyfill Bluebird for the extensions, since BB only exposes a few new methods.
This way while they will use the Bluebird API, the backend will be our Promise

ChemGuy1611:
As a community extension developer, could someone explain to me what is undesirable about Bluebird? I've only used it for mod installers that re-zip the files into an archive with 7zip, as the documentation recommends.

Pickysaurus:
> As a community extension developer, could someone explain to me what is undesirable about Bluebird? I've only used it for mod installers that re-zip the files into an archive with 7zip, as the documentation recommends.

It's just an alternative library to the normal JavaScript `Promise` that Vortex used back when it was originally designed. Methods like `Promise.all()` and `Promise.allSettled()` didn't exist back then natively, so Bluebird allowed doing some "fancy" stuff with async code. It's pretty much redundant these days, but there may be a few places where a Bluebird-only method is called on a Promise, so it can't be changed without editing that code.

IDCs:
@ChemGuy1611
Bluebird was valuable when native Promises were slow and lacked features x years ago, but modern JavaScript has caught up and native promises are now highly optimized.

Additionally:
- Bluebird === ugly promise chains
- Bluebird doesn't play nice with async/await code blocks unless they're wrapped within a native promise.
- Mixing Bluebird and native promises can cause subtle bugs that are pretty difficult to debug at times.
- Maintenance - one extra dependency we have to keep up to date.

ChemGuy1611:
That makes sense. Thanks for the explanations.

PR #19364, opened 2026-01-20
TS strictNullChecks part 1

Aragas:
Done some strictNullChecks for src/main and other small files
Added electron.d.ts to override electron being stupid

### DIFF
--- a/src/actions/notifications.ts
+++ b/src/actions/notifications.ts
@@ -194,7 +194,7 @@ export function addNotification(notification: INotification) {
     if (noti.id !== undefined && noti.displayMS !== undefined) {
       const currentId = noti.id;
       const currentDisplayMS = noti.displayMS;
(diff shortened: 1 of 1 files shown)


typings.custom/electron.d.ts
erri120, line 1: I'm not a fan of custom package typings as they can get stale very quickly. Since this is only about nullability for one parameter, I suggest forcing non-null with the `!` suffix instead of adding custom typings.

Aragas: I don't mind replacing it with suppressions, but it will require to have at least 30 suppressions, from my raw estimation based on code search. Are we sure that suppressing will be better in this case?

erri120: Let's keep the custom typings, the real solution would be to get electron to fix their stuff...

PR #19394, opened 2026-01-22
Add Game Extension Prototype

Sewer56:
Not for merging/review. We still haven't got final designs etc. Do not review.
This is just to make sure the code doesn't get lost.


Sewer56:
Nyo way~!!

PR #19429, opened 2026-01-26
chore: refactor icon component

r1chm8:
## Summary

The Icon component previously accepted string values that mapped to MDI/NXM icon names or raw icon paths.

This PR removes support for string-based icons in favour of always importing icon paths directly. This simplifies the Icon component and significantly improves developer experience by providing proper type safety. Imported icons are now type-checked at compile time, whereas string values offered no guarantees and were prone to runtime errors.

## Changes

- Removed string-based icon support in favour of imported icon paths
- Updated all existing string icon usages to use the correct MDI imports
- Migrated icon sizing to Tailwind utility classes instead of custom CSS variables
- Removed unnecessary React. prefixing from related components

## Benefits

- Cleaner, simpler Icon API
- Stronger type safety for icons
- More consistent styling via Tailwind
- Reduced surface area for runtime errors

### DIFF
--- a/src/stylesheets/tailwind-v4.css
+++ b/src/stylesheets/tailwind-v4.css
@@ -24,15 +24,5 @@
 @import "./shared/theme";
 @import "./shared/elements";
 
-:root {
-    /* Icon sizes for consistent icon sizing, don't think this is needed anymore should be able to use `size-[x]` */
-    --icon-size-xs: 12px;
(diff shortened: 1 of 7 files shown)


insomnious:
Looks good to me. Icons are still displaying fine when built. The autocomplete works well now so we don't have to guess or manually look up the `mdiXXXXXX` strings.

image

PR #19432, opened 2026-01-26
Enable recommended ESLint rules with type information 

erri120:


erri120:
nvm, eslint has issues when typechecking is enabled. Turns the linting step from taking 1 second to 30+ minutes depending on the hardware, likely 60+ minutes on GitHub CI.

PR #19436, opened 2026-01-26
Remove unnecessary `safeCreateAction` wrapper

erri120:
`disableChecking` is a thing...


Aragas:
I'm not sure it's safe to remove the methods, named imports will break in extensions as I understand

erri120:
> I'm not sure it's safe to remove the methods, named imports will break in extensions as I understand

`safeCreateAction` isn't part of the public facing API. It's internal as far as I can see. You can verify by searching for `safeCreateAction` in the `etc/vortex_devel_api.md` API export.

PR #19438, opened 2026-01-26
Port ProcessMonitor to cross-platform using ps-list

Sewer56:
Note that this PR is basically the old code, ported, and some basic unit tests for sanity. The reason for the large diff is comments; I wanted to understand a bit of what it was doing, so I've done some labelling between myself and GPT-5.2. Otherwise it's a mostly straight port.

## Summary
- Port `ProcessMonitor` from Windows-only `winapi-bindings` to cross-platform `ps-list` library
- Add provider abstraction (`IProcessProvider`) for dependency injection and testability
- Add unit tests and documentation

### Cross-platform process enumeration
ProcessMonitor previously used `winapi-bindings.GetProcessList` and early-returned on non-Windows platforms. This change replaces it with `ps-list`, which works on Windows, Linux, and macOS.

### Provider abstraction
Introduced `IProcessProvider` interface and `IProcessInfo` type in a new `processProvider.ts` file. This allows injecting mock providers in tests whilst the default `PsListProcessProvider` handles production use.

### Path extraction from command line
Since `ps-list` doesn't always provide executable paths, added `getCommandPath()` to parse paths from the raw command line. Handles:
- Quoted paths: `"C:\Program Files\Game\game.exe" --args`
- Unquoted paths with spaces: `/home/user/Steam Games/game.exe --args`

### Signal-terminated process handling
Fixed `ExtensionManager.runExecutable()` to handle child processes terminated by signals (where exit code is `null`).
This was needed to avoid getting a nasty error handler dialog on Linux if game was unexpectedly killed. e.g. By sending a signal to it, or CTRL+C, or otherwise.

### Test coverage
Added 5 unit tests covering:
- Tool detection for child processes
- Tool stopped dispatch when process exits
- Detached game vs. child-only tool filtering
- Command-line path parsing with spaces
- Skipping redundant dispatch when PID unchanged

## Future work
This is an iterative improvement bringing xplat whilst preserving existing behaviour. Future improvements could include:
- Extracting and separately testing utility methods (`getCommandPath`, `isChildProcessOfVortex`)
- Additional platform-specific edge case handling

I'd honestly want to start with splitting apart the helper methods and doing testing on those, rather than integration tests with mocks; but there'll be another time for that.

## Testing
- [x] Linux - tested locally
- [x] Windows - basic tests
    - [ ] Needs additional regression testing (i.e. tools, edge cases, etc.)
- [ ] macOS - untested


Aragas (dismissed):

IDCs:
TIL that the process monitor never worked when executing through a game store (makes sense)

Sewer56:
Yup @IDCs , when IPC is involved.
And in some cases when invoking runs a shell, which then runs the game. (e.g. Stardew Valley on Linux)

PR #19460, opened 2026-01-28
Changed: limit Prettier formatting to code files only

Sewer56:
## Summary
- Limit Prettier to explicit code globs (no markdown/docs)
- Exclude `api/` submodule and document opt‑in formatting
- Add `yarn format:check` to CI workflow

## Details
- `package.json` `format`/`format:check` use explicit glob lists for code files only
- `.prettierignore` keeps exclusions only; formatting is opt‑in via scripts
- `.github/workflows/main.yml` runs `yarn format:check` after lint
- Repo‑wide code formatting applied across source and extensions

## Notes
While CI runs `yarn format:check`, it does not currently auto‑fix with `yarn format`. This is because of the transition to GitLab; I don't yet know how make auto commits over there (have not looked).

## Reminder
Check your "format on save" settings if you explicitly force‑disabled them.


IDCs:
@Nexus-Mods/vortex-developers this looks fine to me but it's a massive PR and could use an extra pair of eyes.

About the CI formatting check:
* it will ensure everyone builds and hopefully test their changes before submitting a PR.
* formatting errors will be reported around ~3 minutes after `main.yml` workflow kicks off, so it shouldn't hinder development too much.

IDCs:
@Sewer56 - formatting check failed

insomnious:
Can you remove the CI format check please and resolve the conflict, then we can merge. We will make sure that people are formatting on save.

PR #19481, opened 2026-01-30
goodbye @electron/remote

IDCs:
Important note: This PR will actively conflict with the extensionManager refactor PR which will be coming soon.

* appUni pattern backwards compatibility is maintained via monkeyPatching the electron module. See `src/util/electron.ts`
* electron `window`, `dialog`, `browserView`, `session`, `menu`, `contentTracing`, `clipboard`, `powerSavingBlock` functionality is now available via preload.
* I've yet to remove @electron/remote from the lock/package files - need a sanity check first. (are there any extensions out there that are importing @electron/remote directly?)


extensions/gamebryo-plugin-management/src/util/UserlistPersistor.ts
```diff
@@ -148,14 +145,20 @@ class UserlistPersistor implements types.IPersistor {
...
   }
 
-  private handleInvalidList() {
-    const dialog = process.type === 'renderer'
-      ? remote.dialog
-      : dialogIn;
+  private async handleInvalidList() {
+    const showMessageBox = async (options: Electron.MessageBoxOptions) => {
+      if (process.type === 'renderer') {
+        const result = await (window as any).api.dialog.showMessageBox(options);
```
Aragas, line 182: Perhaps we should declare the api on the Window type, if we can't propagate it?

```typescript
declare global {
  interface Window {
    api?: IExtensionApi
  }
}
```

src/shared/types/ipc.ts
```diff
@@ -26,6 +26,62 @@ export interface MainChannels {
...
+  temp: string;
+  home: string;
+  documents: string;
+  exe: string;
+  desktop: string;
+};
+
+/** Type containing all known channels for synchronous IPC operations (used primarily by preload scripts) */
+export interface SyncChannels {
+  // NOTE(erri120): These are synchronous IPC channels used during preload initialization.
```
erri120:
```suggestion
  // NOTE: These are synchronous IPC channels used during preload initialization.
```

PR #19498, opened 2026-02-01
Convert Application.ts from Bluebird chains to async/await

halgari:
I'm not really sure what to say, this file offends me. So I sent Claude to work replacing Bluebird with promises (just in this one file) so I can actually read and understand it for some code I'm working on next week. I have a lot of strong feelings about this code, so to keep it lighthearted I think I'll let Claude roast it since it was the one that did the work:

Ah yes, the old Application.ts — a 1,400-line love letter to callback hell's slightly more presentable cousin. Picture, if you will, a regularStart() method that read like a CVS receipt: `.then().tap().then().tap().then().catch().catch().catch().catch()` cascading into the abyss, each indentation level a cry for help that went unanswered. The code didn't just use Bluebird's .tap() — it collected them, nine precious side-effects scattered like breadcrumbs through a forest of broken dreams. And let's pour one out for the pièce de résistance on line 1043: `return new PromiseBB(() => null)` — a promise that intentionally never resolves, the programming equivalent of telling someone "I'll call you back" and then changing your phone number. The createStore() method alone was 350 lines of .then() chains so deeply nested that reading it required supplemental oxygen and a sherpa. This wasn't code; it was an archaeological dig site where each layer revealed new horrors, a monument to the dark days before async/await freed us from the tyranny of callback pyramids wearing a trench coat pretending to be modern JavaScript.

Converted patterns:
- .tap() -> inline statements
- .tapCatch() -> catch block with rethrow
- .catch(ErrorType, handler) -> instanceof checks
- .mapSeries() -> for...of with await
- Never-resolving promise -> await new Promise(() => {})

-- 

Everything compiles and the app still boots for me. This should be a pretty clean replacement, as it's just reorganizing the code.

### DIFF
--- a/src/main/Application.ts
+++ b/src/main/Application.ts
@@ -239,38 +239,36 @@ class Application {
     });
   }
 
-  private startUi(): PromiseBB<void> {
+  private async startUi(): Promise<void> {
     const MainWindow = require("./MainWindow").default;
     this.mMainWindow = new MainWindow(this.mStore, this.mArgs.inspector);
     log("debug", "creating main window");
-    return this.mMainWindow.create(this.mStore).then((webContents) => {
-      if (!webContents) {
-        return PromiseBB.reject(new Error("no web contents from main window"));
-      }
-      log("debug", "window created");
-      this.mExtensions.setupApiMain(this.mStore, webContents);
-      setOutdated(this.mExtensions.getApi());
-      // in the past we would process some command line arguments the same as we do when
-      // they get passed in from a second instance but that was inconsistent
-      // because we don't use most arguments from secondary instances and the
-      // rest get handled by the extension they are intended for.
-      // so now "applyArguments()" is only intended for forwarding messages from
-      // secondary instances
+    const webContents = await this.mMainWindow.create(this.mStore);
+    if (!webContents) {
+      throw new Error("no web contents from main window");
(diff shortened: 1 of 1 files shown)


erri120:
I unfortunately have similar changes on my branch https://github.com/Nexus-Mods/Vortex/compare/master...task/main-async to rework `src/main.ts` and `src/main/Application.ts` to use only native Promises. The only method I haven't ported to native Promises was the `createStore` method because it's going to disappear after @IDCs is done, otherwise the entire file got cleaned up. I was going to put up my branch as a PR after fixing up all files in `src/main` to use native Promises.

insomnious:
Leave this until @erri120 and @IDCs have finished what they've started, we can reassess if anything is needed from it then

IDCs:
@halgari - extension manager separation work is now in, so should be able to pick this up again. I'm afraid it kinda broke your PR (19 conflicts😬) might be simpler to just re-run Claude on it and submit a new PR 😆.

erri120:
`src/main/Application.ts` doesn't have Bluebird references anymore, we can close this PR.

IDCs:
@halgari - will let you close this PR if you agree that it's no longer required.

erri120:
Closing as obsolete.

PR #19509, opened 2026-02-02
Remove bluebird from `src/main`*

erri120:
Completely removes bluebird references from `src/main` besides the redux store related functionality which is going to be (re)moved with another PR.

### DIFF
--- a/src/main/Application.ts
+++ b/src/main/Application.ts
@@ -269,50 +269,52 @@ class Application {
       );
     });
 
+    const onReady = () => {
+      const vortexPath =
+        process.env.NODE_ENV === "development" ? "vortex_devel" : "vortex";
+
+      // if userData specified, use it
+      let userData =
+        args.userData ??
+        // (only on windows) use ProgramData from environment
+        (args.shared &&
+        process.platform === "win32" &&
+        process.env.ProgramData !== undefined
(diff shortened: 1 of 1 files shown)


src/main/devel.ts
```diff
@@ -0,0 +1,27 @@
+import installExtension from "electron-devtools-installer";
+import {
```
Aragas: Just to be sure, the require was previously lazy since it was in a function, it's not breaking anything in production?

src/main/Application.ts
```diff
@@ -296,57 +262,70 @@ class Application {
...
-        startupMode = this.handleSet(args.set, userData);
-      } else if (args.del) {
-        startupMode = this.handleDel(args.del, userData);
-      }
-      if (startupMode !== undefined) {
-        startupMode.then(() => {
-          app.quit();
+    app
+      .whenReady()
+      .then(() => {
```
Aragas: We could extract it to a separate function for better readability? Won't also have such nesting level in this case

src/main/devel.ts
erri120: Changed to async import to prevent hoisting.

Aragas:
Needs conflict resolve and should be ready to go

PR #19534, opened 2026-02-03
Upgrade `sass` from 1.32.* to 1.97.3 and run migrations

erri120:
We were running a 5 year old version of the `sass` package to compile SCSS files at runtime. Much has changed in those 5 years, especially on the CSS front. With the new version we now get warnings about deprecated features in our SCSS files. I ran the `sass-migrator` on all SCSS files in `src` which can auto-fix most issues but not all of them.

Going to be ripping out the StyleManager in a follow-up PR as part of the main/renderer separation.


Aragas:
Looks like it shouldn't affect external extensions 👍

PR #19538, opened 2026-02-03
Remove path aliases

erri120:
We can't have nice things with our current build pipeline.


IDCs:
Sad to see these go, but I did encounter issues with the aliases as well :(

PR #19749, opened 2026-02-11
Added: Core Support for Flatpak Packaging

Sewer56:
# Add Flatpak packaging support for Vortex

## Summary

This PR adds the first usable Flatpak path for Vortex: build, run, install,
and bundle.

This is the first major step toward an easy install experience on
SteamOS/Steam Machines, and distros further beyond.

<Image1> <Image2>

## What you should know first

- This is foundation work to make Flatpak support possible end-to-end.
  - We have made a best effort to align with
    [Flathub requirements][flathub-requirements] so we can publish on Flathub
    (the default Flatpak catalogue on SteamOS).
- It includes docs, helper scripts, manifest wiring, and offline dependency
  source generation.
- A key part of this PR is dependency source manifests plus hash-gated
  regeneration; the next section explains how and why.

## Flatpak dependency model

> [!important]
> Packaging for Flathub has two important requirements:
>
> - [No network access during build][flathub-no-network]
> - [Dependency manifest requirement][flathub-dependency-manifest]

Flatpak is a bit different here: the Flathub store (what SteamOS and most
distros default to) requires us to build Vortex offline.

The way building works is that a tool called `flatpak-builder` fetches every
dependency declared in `sources` up front
(see `flatpak/com.nexusmods.vortex.yaml`), then runs the build
(`build-commands`) without network access.

In this PR, we generate those dependency source lists into:

- `flatpak/generated-sources.json` (npm/yarn artefact sources)
- `flatpak/generated-nuget-sources.json`
  (NuGet artefact sources, C#/.NET FOMOD stuff)

Each entry inside is an artefact URL plus checksum. This makes builds
reproducible for review, and helps protect against supply-chain attacks.

## Developer workflow (summary)

This section is the short version. Full docs are in:

- `docs/flatpak-maintenance.md` - setup, commands, troubleshooting,
  maintainer loop
- `docs/flatpak-technical.md` - architecture and implementation details

### Fast development loop

Use this while iterating quickly.

```bash
python3 flatpak/scripts/flatpak_build.py
python3 flatpak/scripts/flatpak_run.py
```

### UX/distribution loop

Use this when testing install-like behaviour and software centre visibility.

```bash
python3 flatpak/scripts/flatpak_install.py --run
python3 flatpak/scripts/flatpak_bundle.py --skip-build
```

### Why Python scripts?

You may notice the above scripts use Python. This is a conscious decision.

Generating the aforementioned `flatpak/generated-sources.json` and
`flatpak/generated-nuget-sources.json` requires tools
***written in Python*** from `flatpak-builder-tools`
(`flatpak-node-generator` and `flatpak-dotnet-generator`).

So Python is required for this workflow.

The slightly tricky part is installing `flatpak-builder-tools` across Linux 
distros. Some environments disallow direct/global `pip install`, for instance,
in cases where `python` itself is shipped as part of the distro.

Sometimes this can catch people offguard; if they're used to e.g.
Python in Windows where this isn't the case.

To ensure a smooth experience that "just works" regardless of global Python
environment, the scripts in in `flatpak/scripts/` self-manage a `virtualenv`
(venv) and install the required generator tools there.

So anyway, I figured, since we need to do a bit of Python anyway;
I might aswell use it for helper scripts too; for consistency.

## Why generated files are committed?

`flatpak/generated-sources.json` and
`flatpak/generated-nuget-sources.json` are generated files, but they are still
committed on purpose.

Submitting to Flathub happens via a PR, and these dependency source files are
part of what reviewers check
([Submitting for Flathub][flathub-submission-pr]).

Keeping these files in our repo (just like `yarn.lock`) means what we test
locally is 1:1 with what we submit to Flathub.

It also gives us a clear history of dependency-source changes across
submissions.

## Source files auto-generation strategy

We may eventually want to have Flatpak builds running in CI on
PRs, in addition to typical Windows and Linux builds for testing. But with that
comes a caveat, the `generated-sources.json` and friends need to be kept up
to date!

And that would normally require manual updates, which are easy to miss!
If `yarn.lock` or NuGet inputs change and the manifests are not regenerated,
Flatpak can quietly drift to a different dependency set from non-Flatpak builds.
Not good!

At first I considered a dedicated PR check for this (i.e. `GitHub Actions workflow`),
to ensure our sources stay in sync with the lockfiles, (if `yarn.lock` changes,
error if sources not updated, etc).

However, I figured that would create developer frustration:

- PRs editing lockfiles (`yarn.lock`) would fail until someone reran generators
  - To auto-push a commit fix by bot, requires 2 workflows, due to write rules.
    - And with Gitlab in the mix, it's unclear how it'd work.
  - For PRs, not everyone also sets `allow edits by maintainers`. etc.

So this PR takes a quieter approach: *What if we don't update the sources until
we need to?*

1. make a hash of generator inputs (all lockfiles)
2. if hash does not match, regenerate sources

I added this neat logic to the build scripts. Meaning that you can get 
flatpak builds in CI and for testing without the risk of forgetting 
manual source updates.

## Key implementation changes

- `flatpak/com.nexusmods.vortex.yaml`
  - our Flatpak build configuration
- `preinstall.js`
  - adds `VORTEX_SKIP_SUBMODULES=1`
  - skips submodule updates in Flatpak builds while still running required
    FOMOD preinstall build steps
  - we skip this because Flatpak builds run without networking; submodules
    are updated outside the Flatpak build.
- `flatpak/scripts/flatpak_run.py` + `src/app/Application.ts`
  - adds `VORTEX_ENABLE_LOGGING=1` env var to enable debugging in Flatpak
    builds
  - enabled by passing `--log` to `flatpak_run.py`
- `package.json` + `yarn.lock`
  - updates `dir-compare/minimatch`; there were previously two entries in
    `yarn.lock`, which tripped up the source generator.
- `electron-builder-config.json`
  - unpacks `assets/dotnetprobe`; it was not being extracted during build,
    so it never ran.

### Linux Protocol Registration

Added dynamic protocol registration for `nxm://` links on Linux.

Previously, the protocol handler installed during initial repo setup
(`scripts/linux-protocol-registration.sh`) always took precedence over the 
Flatpak one due to name conflict. 

This made the Flatpak not fully testable end-to-end, as someone on a dev machine
couldn't start downloads, etc.

If you knew how to, switching between development and Flatpak builds required
manually running commands to change the default handler.
This was very painful for testing.

I ported the general protocol registration from [NexusMods.App][nma-protocol], 
with some tweaks:

- **Separate development handler** - Development builds now use `com.nexusmods.vortex.dev.desktop` while Flatpak/production builds use `com.nexusmods.vortex.desktop`, allowing both to coexist.
- **Automatic registration** - Vortex now auto-registers itself as the protocol handler on startup when appropriate.
  - This means no manual setup required when setting up development on a new machine.
  - And you can also seamlessly switch between the flatpak and development builds.
- **Dynamic desktop file generation** - The dev desktop entry is generated at runtime with the correct paths and environment

**Extra Notes:**

I learned something new today. Turns out, some browsers (e.g., Vivaldi)
may use a modified `LD_LIBRARY_PATH`. On my end, this actually caused Electron
to crash! So in the protocol handler, I unset this; just in case. Turns out
it's common practice to do so, same with `LD_PRELOAD`. TIL.

**Environment persistence** - Not strictly needed, but I decided to persist a
few env-vars (`XDG_DATA_DIRS`, `GIO_EXTRA_MODULES`, `GDK_PIXBUF_MODULE_FILE`,
`CHROME_DEVEL_SANDBOX`, `ELECTRON_OVERRIDE_DIST_PATH`, `NODE_ENV`) if a protocol
handler was created from a development build (`yarn start`). I needed this on
Nix somewhat, to avoid unexpected behaviours.

**xdg-utils workarounds** - Back when working on NMA, I found some bugs in
xdg-utils around 'generic' fallbacks for `.desktop` files. Although they've been
patched since, I still ported the workaround from NMA in case people are running older
OS versions.

[nma-protocol]: https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/NexusMods.Backend/OS/LinuxInterop.Protocol.cs
[xdg-issue]: https://gitlab.freedesktop.org/xdg/xdg-utils/-/issues/279

### Process monitor is stubbed for Flatpak

Host process monitoring is not possible in Flatpak. 
So we can't reliably tell when games close.

Short-term (80/20), ProcessMonitor in Flatpak will be stubbed, just pretending
process termination happened immediately.

Expected impact should be low (likely single-digit game integrations out of
~500). Normally we wait for process exit, because Windows file-locking can cause
issues; but on Linux we can mutate files as they're in use just fine.

There is a small edge case 'the likely single digit number of game plugins'
which do rely on post-exit triggers (for example BG3 load-order refresh).
For those cases, we will investigate workaround down the line. Be it wrapping
the game's executable with something that drops a file as means of IPC,
or otherwise.

### Window System

Right now this boots in X11.

When we update Electron to next version, it will default to Wayland on Wayland sessions.

## Misc Notes

FOMOD projects build and there are no missing .NET warnings, I haven't tested
whether .NET works beyond having Vortex running `dotnetprobe` at startup
successfully.

## Quick validation (Linux)

```bash
python3 flatpak/scripts/flatpak_build.py
python3 flatpak/scripts/flatpak_run.py --log
python3 flatpak/scripts/flatpak_install.py --run
python3 flatpak/scripts/flatpak_bundle.py --skip-build
```

Expected:

- build succeeds offline
- FOMOD/.NET preinstall components build
- app installs and appears in Discover
- bundle generation succeeds

### Packaging

- `flatpak/com.nexusmods.vortex.yaml` - main Flatpak manifest
- `flatpak/run.sh` - Flatpak entrypoint launcher
- `flatpak/com.nexusmods.vortex.desktop` - desktop launcher metadata
- `flatpak/com.nexusmods.vortex.metainfo.xml` - AppStream metadata

### Documentation

- `docs/flatpak-maintenance.md` - maintainer guide
- `docs/flatpak-technical.md` - technical reference

### Scripts

- `flatpak/scripts/flatpak_build.py` - build Flatpak
- `flatpak/scripts/flatpak_run.py` - run build output
- `flatpak/scripts/flatpak_install.py` - install from local repo
- `flatpak/scripts/flatpak_bundle.py` - create `.flatpak` bundle
- `flatpak/scripts/flatpak_sources.py` - generate/sync npm sources
- `flatpak/scripts/_flatpak_sources_hash.py` - npm lockfile hashing
- `flatpak/scripts/flatpak_nuget_sources.py` - generate/sync NuGet sources
- `flatpak/scripts/_flatpak_nuget_hash.py` - NuGet input hashing
- `flatpak/scripts/_flatpak_env.py` - venv + generator pinning helpers

### Generated source manifests

- `flatpak/generated-sources.json` - npm/yarn dependency sources
- `flatpak/generated-sources.hash` - npm/yarn input hash
- `flatpak/generated-nuget-sources.json` - NuGet dependency sources
- `flatpak/generated-nuget-sources.hash` - NuGet input hash

### Supporting code changes

- `preinstall.js` - `VORTEX_SKIP_SUBMODULES=1` support
- `src/app/Application.ts` - `VORTEX_ENABLE_LOGGING=1` path
- `package.json` + `yarn.lock` - offline resolver alignment (`minimatch`)
- `electron-builder-config.json` - unpack `assets/dotnetprobe`

### Protocol registration

- `src/util/protocolRegistration/index.ts` - platform-agnostic protocol registration facade
- `src/util/protocolRegistration/electron.ts` - Electron-backed registration (Windows/macOS)
- `src/util/protocolRegistration/types.ts` - shared type definitions
- `src/util/protocolRegistration/linux/index.ts` - Linux route dispatcher
- `src/util/protocolRegistration/linux/nxm.ts` - nxm:// protocol implementation for Linux
- `src/util/protocolRegistration/linux/common.ts` - Linux desktop database helpers
- `src/util/protocolRegistration/linux/desktopFileEscaping.ts` - desktop file escaping utilities
- `src/util/ExtensionManager.ts` - refactored to use new protocol registration system
- `__tests__/desktopFileEscaping.test.ts` - tests for desktop file escaping

### Removed files

- `scripts/linux-protocol-registration.sh` - replaced by new TypeScript implementation
- `scripts/com.nexusmods.vortex.desktop.in` - replaced by dynamic generation

[flathub-requirements]: https://docs.flathub.org/docs/for-app-authors/requirements
[flathub-no-network]: https://docs.flathub.org/docs/for-app-authors/requirements#no-network-access-during-build
[flathub-dependency-manifest]: https://docs.flathub.org/docs/for-app-authors/requirements#dependency-manifest
[flathub-submission-pr]: https://docs.flathub.org/docs/for-app-authors/submission#submission-pr


Sewer56:
Will resolve conflicts next morning.
I also have to figure out why I see no start menu entry, despite protocol handler working.

caniko:
I have been working on something similar on the `nixpkgs` side, [init PR](https://github.com/NixOS/nixpkgs/pull/491344).

It compiles just fine! Mostly working except for #20439; I am using the same strategy that you are using in this flatpak definition to interact with the filesystem, `native` mode.

insomnious (approved):

PR #19828, opened 2026-02-12
Populate pluginList on collection installation

Aragas:


extensions/gamebryo-plugin-management/src/types/IStateWithGamebryo.ts
IDCs, line 1: This is probably the best example for the need of a new standard of type overloading/extension. 

```
declare module 'path_to_IState_declaration' {
    export interface ThisIsALeafOrBranch {
        something: string;
    }
}
```

PR #20270, opened 2026-02-16
Download Issues Fixes

Aragas:
* Truncate file if the size is bigger than expected
* Avoid compressed chunks
* Content-range was only capturing the last digit
* Reset write offset on non chunkable response
* Fixed file used by another process
* Additional validations on file before performing actions
* Retry after pause on empty chunks
* Rotate through our available downland urls on retries


IDCs:
This PR is a bit too significant for a 1.16 patch - it also fails to tend to the corruption issue.

Closed in favour of:
https://github.com/Nexus-Mods/Vortex/pull/20300

There's good things in here so going to keep the branch for future enhancements.

PR #20402, opened 2026-02-17
feat: replace LevelDB with DuckDB as persistence backend

halgari:
## What this PR does

Vortex is gradually moving business logic out of the renderer process (React/Redux) and into the main process. This PR replaces the LevelDB persistence backend with DuckDB (via the \`level_pivot\` extension), and adds the infrastructure for main-process code to write state and sync it back to the renderer.

### Persistence backend

DuckDB attaches a LevelDB file as a typed relational database via the \`level_pivot\` extension. State is stored as \`path -> JSON value\` pairs, so the renderer's Redux store and hydration flow are unchanged. A single \`DuckDBSingleton\` owns the DuckDB instance and connection pool; each \`LevelPersist\` database attaches to it under a unique alias so all hives share one engine.

### Bidirectional sync

State previously only flowed renderer -> Redux -> \`persist:diff\` IPC -> main -> LevelDB. Main-process code can now also initiate writes:

1. Call \`pushStateToRenderer(hive, operations)\` with a list of diff operations
2. Main writes them to DuckDB and waits for the write to flush
3. Main sends a \`persist:push\` IPC message to all renderer windows
4. Renderer applies the operations via a \`__persist_push\` Redux action
5. \`persistDiffMiddleware\` skips \`__persist_push\` actions to avoid a feedback loop

### SQL query files and the parser

Named SQL queries live in \`src/queries/\` as plain \`.sql\` files annotated with a comment-based DSL:

```sql
-- @type select
-- @name recently_managed_games
-- @description Games the user has managed, sorted by last activation date
-- @param profile_id VARCHAR
SELECT game_id, last_activated FROM ...
```

The \`@type\` tag at the top of each file controls how the query is handled. \`setup\` queries create pivot tables at startup; \`view\` queries create SQL views; \`select\` queries are the named, parameterised queries called by name at runtime.

\`queryParser.ts\` scans the \`src/queries/\` tree at runtime, parses each file extracting annotations and SQL body, validates name uniqueness across all files, and returns a list of \`ParsedQuery\` objects.

### Code generation

\`scripts/generate-query-types.ts\` is a build-time script that:

1. Spins up a temporary DuckDB instance and attaches an empty LevelDB for schema-only work
2. Runs all \`setup\` queries, then \`DESCRIBE\`s them to read column types
3. Runs all \`view\` queries, then \`PREPARE\`s each \`select\` query to read result column metadata
4. Maps DuckDB column types to TypeScript types (\`VARCHAR -> string\`, \`INTEGER -> number\`, etc.)
5. Writes \`src/main/store/generated/queryTypes.ts\` with typed \`Params\` and \`Row\` interfaces for every query, a \`Models\` interface with typed \`Table<T>\`/\`View<T>\` accessors, and a \`createModels(db)\` factory

SQL schema changes propagate to TypeScript automatically with a single script run.

### Live query updates

After each write transaction commits, \`LevelPersist\` reads \`level_pivot_dirty_tables()\` to find which tables were modified. Those table names are forwarded to \`QueryInvalidator\`, which debounces rapid writes (~16 ms) and batches them before acting.

\`QueryRegistry\` maintains a reverse index built at startup by calling \`getTableNames(sql)\` on each registered query, producing a map of \`tableName -> Set<queryName>\`. \`QueryInvalidator\` uses this index to resolve dirty tables into affected query names.

\`QueryWatcher\` holds a set of active subscriptions. When \`QueryInvalidator\` flushes, it calls \`QueryWatcher.onQueriesInvalidated(affectedQueries)\`. For each matching subscription the watcher re-executes the query, JSON-compares the result to the previous snapshot, and fires the callback only if something changed, delivering \`{ previous, current }\` diffs to subscribers.

This is all main-process only. The renderer continues using Redux; live query results feed back to Redux via \`pushStateToRenderer\` when main-process code decides to act on them.

## Why DuckDB

As we move code to the main thread, we want main-process logic to query and write application state using real SQL. DuckDB gives us joins, aggregations, window functions, and typed schemas on top of the same LevelDB files already on disk.

### DIFF
--- a/src/main/src/store/LevelPersist.ts
+++ b/src/main/src/store/LevelPersist.ts
@@ -64,6 +64,7 @@ class LevelPersist implements IPersistor {
 
   #mConnection: DuckDBConnection;
   #mAlias: string;
+  #mInTransaction: boolean = false;
 
   constructor(connection: DuckDBConnection, alias: string) {
     this.#mConnection = connection;
@@ -147,11 +148,35 @@ class LevelPersist implements IPersistor {
   }
 
   public async setItem(statePath: string[], newState: string): Promise<void> {
-    await this.#mConnection.run(
-      `INSERT INTO ${this.#mAlias}.kv VALUES ($1, $2)
-         ON CONFLICT (key) DO UPDATE SET value = excluded.value`,
-      [statePath.join(SEPARATOR), newState],
-    );
+    const key = statePath.join(SEPARATOR);
+    // level_pivot tables don't support UNIQUE indexes, so ON CONFLICT
+    // upserts aren't possible. Try UPDATE first; INSERT only if the key
+    // didn't exist. Wrap in a transaction when called outside of one to
+    // ensure atomicity.
+    const ownTransaction = !this.#mInTransaction;
+    if (ownTransaction) {
+      await this.beginTransaction();
+    }
+    try {
+      const result = await this.#mConnection.runAndReadAll(
+        `UPDATE ${this.#mAlias}.kv SET value = $2 WHERE key = $1 RETURNING key`,
+        [key, newState],
+      );
+      if (result.getRows().length === 0) {
+        await this.#mConnection.run(
+          `INSERT INTO ${this.#mAlias}.kv VALUES ($1, $2)`,
+          [key, newState],
+        );
+      }
+      if (ownTransaction) {
+        await this.commitTransaction();
+      }
+    } catch (err) {
+      if (ownTransaction) {
+        await this.rollbackTransaction();
+      }
+      throw err;
+    }
   }
 
   public async removeItem(statePath: string[]): Promise<void> {
@@ -166,6 +191,7 @@ class LevelPersist implements IPersistor {
    */
   public async beginTransaction(): Promise<void> {
     await this.#mConnection.run("BEGIN TRANSACTION");
+    this.#mInTransaction = true;
   }
 
   /**
@@ -173,6 +199,7 @@ class LevelPersist implements IPersistor {
    */
   public async commitTransaction(): Promise<void> {
     await this.#mConnection.run("COMMIT");
+    this.#mInTransaction = false;
   }
 
   /**
@@ -180,6 +207,7 @@ class LevelPersist implements IPersistor {
    */
   public async rollbackTransaction(): Promise<void> {
     await this.#mConnection.run("ROLLBACK");
+    this.#mInTransaction = false;
   }
 
   /**
--- a/scripts/generate-query-types.ts
+++ b/scripts/generate-query-types.ts
@@ -282,7 +282,7 @@ function generateTypeScript(
     for (const col of columns) {
       lines.push(`  ${col.name}: ${col.tsType};`);
     }
-    lines.push("}");
+    lines.push("};");
     lines.push("");
   }
 
--- a/src/main/src/store/generated/queryTypes.ts
+++ b/src/main/src/store/generated/queryTypes.ts
@@ -9,7 +9,7 @@ export type ModsPivotRow = {
   name: string;
   version: string;
   state: string;
-}
+};
 
 /** Row type for the 'profiles_pivot' pivot table */
 export type ProfilesPivotRow = {
@@ -17,7 +17,7 @@ export type ProfilesPivotRow = {
   name: string;
   gameId: string;
   lastActivated: bigint;
-}
+};
 
 /** Parameters for the 'recently_managed_games' query */
 export interface RecentlyManagedGamesParams {


halgari:
Don't merge this, we're likely going to do a lot of talks about this and rework it before a merge.

src/shared/types/ipc.ts
```diff
@@ -163,6 +163,9 @@ export interface MainChannels {
 
   // Menu click events (main -> renderer)
   "menu:click": (menuItemId: string) => void;
+
+  // Query system: notify renderer that queries have been invalidated
+  "query:invalidated": (queryNames: string[]) => void;
```
erri120: You can use the `QueryName` type here:

```suggestion
  "query:invalidated": (queryNames: QueryName[]) => void;
```

src/shared/types/ipc.ts
```diff
@@ -315,6 +318,15 @@ export interface InvokeChannels {
 
   // Compile stylesheets
   "styles:compile": (filePaths: string[]) => Promise<string>;
+
+  // Query system: execute a named query
+  "query:execute": (
+    queryName: string,
```
erri120: You can use the `QueryName` type here:

```suggestion
    queryName: QueryName,
```

src/main/store/DuckDBSingleton.ts
```diff
@@ -0,0 +1,168 @@
...
+ * Manages a single shared in-memory DuckDB instance with level_pivot loaded.
+ * All LevelPersist instances attach their databases to this shared instance.
+ */
+class DuckDBSingleton {
+  private static sInstance: DuckDBSingleton | undefined;
+
+  private mDuckDB: DuckDBInstance | undefined;
+  private mInitialized: boolean = false;
+  private mAttachedDatabases: Map<string, string> = new Map(); // alias -> path
+  private mConnections: DuckDBConnection[] = [];
```
erri120: Prefer JavaScript private over TypeScript private.

src/main/store/DuckDBSingleton.ts
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * Initialize the shared DuckDB instance, installing and loading level_pivot.
+   * Safe to call multiple times -- only initializes once.
+   */
+  public async initialize(): Promise<void> {
+    if (this.mInitialized) {
+      return;
+    }
+
+    const { DuckDBInstance: DuckDBInstanceCtor } = require("@duckdb/node-api");
```
erri120: Use ESM import syntax instead of CJS require syntax.

src/main/store/LevelPersist.ts
```diff
@@ -1,30 +1,28 @@
...
-function repairDB(dbPath: string): PromiseBB<void> {
-  return new PromiseBB<void>((resolve, reject) => {
+function repairDB(dbPath: string): Promise<void> {
+  return new Promise<void>((resolve, reject) => {
     log("warn", "repairing database", dbPath);
-    const leveldown: typeof leveldownT = require("leveldown");
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const leveldown = require("leveldown") as {
+      repair: (path: string, cb: (err: Error) => void) => void;
+    };
```
erri120: Use ESM import syntax instead of CJS require syntax.

src/main/store/QueryInvalidator.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (!window.isDestroyed() && window.webContents !== undefined) {
+        window.webContents.send("query:invalidated", affectedQueries);
+      }
+    }
+
+    // Notify watcher (fire-and-forget)
+    this.mWatcher?.onQueriesInvalidated(affectedQueries).catch((err) => {
+      log("warn", "QueryWatcher notification failed", {
+        error: (err as Error).message,
+      });
```
erri120:
```suggestion
      log("warn", "QueryWatcher notification failed", err);
```

insomnious:
Closes APP-67

src/main/src/store/LevelPersist.ts
```diff
@@ -48,111 +26,195 @@ class LevelPersist implements IPersistor {
...
+    cb: T,
+  ): (...args: Parameters<T>) => ReturnType<T> {
+    return ((...args: Parameters<T>): ReturnType<T> => {
+      const stackErr = new Error();
+      return cb(...args).catch((unknownErr: unknown) => {
+        const err = unknownToError(unknownErr);
+        err.stack = stackErr.stack;
+        throw err;
+      }) as ReturnType<T>;
+    }) as (...args: Parameters<T>) => ReturnType<T>;
```
erri120: I removed this restacking function from the project a while ago. Did you need this explicitly for something? Otherwise we should remove this, our error handling and reporting is already a mess, don't want custom error mutations in random pieces of code 😬

Aragas:
Looks really cool!

src/main/src/store/LevelPersist.ts
halgari: removed

erri120:
I'm getting this error when launching Vortex:

```
Failed to store application state: Binder Error: The specified columns as conflict target are not referenced by UNIQUE/PRIMARY KEY CONSTRAINT or INDEX
```

Stack shows `ReduxPersistorIPC.ts:253:11`.

PR #20530, opened 2026-02-18
Fix vortex-run path in app/package.json

erri120:


Sewer56:
I pointed this out myself, and yup, checks out.

Issue #19746, opened 2026-02-11
unresolved files conflict


Road-spec:
i have arledy resolved everything but with update i cant play skyrim because of this

**Your Nexus Mods username**
Will help us debug your issue.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Platform (please complete the following information):**
 - OS: [e.g. Windows 10]
 - Vortex Version [e.g. 0.16.12]

**Additional context**

<img width="1120" height="870" alt="Image" src="https://github.com/user-attachments/assets/1d5a6e5a-21f6-476c-b18a-4921e56dc35d" />

KyrianKroon:
Disabling and re-enabling the mods in question solved it for me. But definitely a bug.

Road-spec:
Thanks 😊

iggykg-bit:
> Disabling and re-enabling the mods in question solved it for me. But definitely a bug.

Hm,didnt solve it for me.You disabled both conflicting mods?To me its showing more then 100 mods,it would took an hour to disable and enable them all >.<

Road-spec:
For me i reinstall all mod in the conflict problems (so all the name you saw above)

IDCs:
Anyone that is still affected by this bug - are you able to verify a fix for this?

Will require us to replace a file in Vortex's installation directory.

1. Close Vortex

2. Download this file [index.js](https://github.com/user-attachments/files/25478123/index.js)

3. Go to the `mod-dependency-manager` folder in your Vortex installation folder, by default this is located in: `C:\Program Files\Black Tree Gaming Ltd\Vortex\resources\app.asar.unpacked\bundledPlugins\mod-dependency-manager`

4. Backup the `index.js` file in that folder so that you can restore it after testing.

5. Replace that file with the one you downloaded in step 2.

Run Vortex and see if you're still getting cyclical dependencies and/or non-resolveable conflicts.

Keep in mind that Vortex checksums its files on startup - it will complain about corruption, ignore it as that's expected (we changed the code)

PR #20645, opened 2026-02-19
fixed inconsistent rule checks between unsolved notif and rules dialogs

IDCs:
conflict rules are bidirectional, the dialog rule lookups were unidirectional and could potentially not display b->a conflicts.

additionally we had a duplication issue where a conflict pair could potentially pass through the filter twice (once from each side) and show up as two separate unsolved entries (when they were actually solved)

fixes nexus-mods/vortex#19746

Funny fact - this bug was in there since before 1.15.2, not sure why it's popping its head now.


jerbr08:
My original post#19669
I updated to 1.16.4 and am still having the same problems.

PR #20669, opened 2026-02-19
Design system restructure, demos & docs

r1chm8:
### Restructured component library
- Moved all components from `src/renderer/tailwind/components/next/` to `src/renderer/ui/components/`, removing the misleading "tailwind" naming
- Moved icon paths from `tailwind/lib/icon-paths/` to `ui/lib/icon_paths/`
- Extracted shared utilities (`joinClasses`, `getTabId`, `XOr`, `ResponsiveScreenSizes`) into `ui/utils/` with proper barrel exports
- Moved stylesheets from `stylesheets/shared/` to `stylesheets/ui/` to match  the new structure
- Removed the old `src/renderer/tailwind/` directory, `Tailwind` namespace  export, and `TailwindTest` component
- Removed the unused `Tailwind` namespace export from `src/renderer/api.ts`
- Steamlined the why typograhy is managed to make use of default tailwind var instead of doing this ourself
- Updated component to use new tailwind text vars so it's easier to see what typograhy style is being applied

### New & updated demo files
- Created demo components for all major components: `ButtonDemo`, `IconDemo`, `TypographyDemo`, `TabsDemo`, `InputDemo`, `SelectDemo`, `DropdownDemo`, `ListingDemo`, `PaginationDemo`, `PickerDemo`, `CollectionTileDemo`
- Each demo showcases all variants, states, and sizes of its component

### Design System dev page
- Renamed the `tailwind_dev` extension to `design_system_dev`
- Renamed the view from `TailwindPage` to `DesignSystemPage`, nav label from "Tailwind" to "Design System"
- All demo components are accessible via tabbed interface on this page (dev mode only)
- Added `mdiPalette` icon for the modern menu and `highlight-ui` as the legacy SVG sprite fallback

### Import cleanup
- ~~Updated 34 files to use barrel index imports (`join_classes` and `get_tab_id`) instead of importing internal module files directly~~
- Remove all barrel files

### Documentation
- Rewrote `src/renderer/ui/README.md` with current directory structure, import patterns, component API docs with defaults, and CSS conventions


erri120:
> Updated all imports across 34 files to use barrel index exports (get_tab_id and join_classes) instead of importing the internal module file directly

We should have less barrel files instead of more. While they might seem like a nice thing to have, it adds more boilerplate and more complexity to the build step. There are very few cases where we want to re-export something. Instead of having `src/renderer/ui/utils/get_tab_id/getTabId.ts` and `src/renderer/ui/utils/get_tab_id/index.ts` just have `src/renderer/ui/utils/getTabId.ts` and import it directly. Same for `src/renderer/ui/utils/join_classes/*`.

r1chm8:
> > Updated all imports across 34 files to use barrel index exports (get_tab_id and join_classes) instead of importing the internal module file directly
> 
> We should have less barrel files instead of more. While they might seem like a nice thing to have, it adds more boilerplate and more complexity to the build step. There are very few cases where we want to re-export something. Instead of having `src/renderer/ui/utils/get_tab_id/getTabId.ts` and `src/renderer/ui/utils/get_tab_id/index.ts` just have `src/renderer/ui/utils/getTabId.ts` and import it directly. Same for `src/renderer/ui/utils/join_classes/*`.

@erri120 I couldn't agree more, I hate barrel files, the only reason I put them in was because I thought that's how Vortex does it, I'll remove them all from in the UI directory.

insomnious (dismissed):

PR #20824, opened 2026-02-21
Theme switcher fixes

### LINKED ISSUE #19323: Cloning default themes doesn't work as expected
**Your Nexus Mods username**
wxMichael

**Describe the bug**
When cloning the default theme, the newly created theme looks like the old Vortex theme with lighter colors.

**To Reproduce**

1. Set theme to default
3. Click on Clone
4. Click on Apply at the bottom-right
5. Primary color will shift to light gray
**Expected behavior**
There should be no visible difference between a theme and its clone.

**Screenshots**

default

default-clone

 - OS: Windows 11
 - Vortex 1.16.0-beta.3

**Additional context**
This has been an issue for quite a while, so it's not specific to the beta.

wxMichael:
Fixes a number of issues with theme-switcher:
- Variables file reading now accounts for CRLF & LF line endings
- Theme defaults updated to current design, so cloning doesn't revert colors to Classic
- Default values for Dashlet Height & Titlebar Rows were unused
- Comments in variables.scss are now ignored. Commented variables in Compact were being read as non-comment.
- Fixed lag in color picker when dragging the selector
- Fixed variables displaying invalid hex codes (themes were using SCSS functions that the editor can't parse)
- Applied the current theme to the color picker
- Fixed incorrect tool icon border color when cloning themes like Classic

Fixes #19323 

Themed color picker:

color-picker

Error when cloning Contrast theme:

contrast-clone-error


IDCs:
Hey @wxMichael - going to merge this in as it can't hurt, but please note that I'm unsure of how this extension is going to develop in the future given UIV2 changes - I'm assuming it will get re-written entirely.

PR #21480, opened 2026-03-10
changes for linux to support at least launching Skyrim SE

killzoms:
This fixes

- [x] Steam Game Detection
- [x] Proton Game Launching under the correct App ID
- [x] Staging Directory changes
- [x] Path Suggestion
- [x] Hardlink Deployment
- [x] Move Deployment

This has not yet been tested on Windows


killzoms:
What is not fixed

1. ini parsing/editing
2. Some Fomod features: Failed to install Faster HDT-SMP Properly
3. Case sensitivity for deployment

Maybe possible?
Games added to Steam via Non-Steam Game might be able to be launched via Vortex? Not sure havent tested

PR #21547, opened 2026-03-14
Cleanup for game-pillarsofeternity2

wxMichael:
- Moved `game-pillarsofeternity2/` from `extensions/` to `extensions/games/`
- Added config so the extension displays as `Game: Pillars of Eternity 2: Deadfire` instead of `Game Pillarsofeternity2`
- Made package.json entries consistent with other extensions like Stardew Valley.  
  The only change I'm unsure of is `"main"`: `./out/index.js` ->  `dist/index.js`


wxMichael:
It seems like package.json's main variable is no longer used. Guessing it's from an old build setup?
If so I'd like to clean all those up.

**Edit:** @erri120 confirmed main isn't used in the extensions as they aren't typical packages, so I removed them in #21557

erri120 (approved):

PR #21606, opened 2026-03-17
removed bluebird from `mod_management` extension

IDCs:
## Summary

Removes all Bluebird imports from the `mod_management` extension and its dependent types, replacing them with native Promises and `async/await`. This is the largest single Bluebird consumer in the codebase (26+ import sites). Cleaning it first unblocks incremental migration of the remaining extensions.

### Core mod_management (27 files)

- **InstallManager.ts**: Replaced all `Bluebird`/`toBluebird` usage with native Promises. Added `mapWithConcurrency` helper to replace `Bluebird.map` with concurrency. Converted `Bluebird.each`/`mapSeries` to sequential `for...of` + `await` loops. Converted `Bluebird.delay` to `setTimeout`-based Promise.
- **LinkingDeployment.ts**: Replaced `PromiseBB.map` with `mapWithConcurrency` (concurrency: 50) for `removeDeployedFile` calls in `finalize`, preventing IPC overload with the elevated symlink activator. Added duck-typing for ENOENT checks on IPC-deserialized error objects. Converted all Bluebird return types to `Promise`/`PromiseLike`.
- **InstallContext.ts**: `PromiseBB.delay()` → `setTimeout` in native Promise.
- **eventHandlers.ts, index.ts, listInstaller.tsx, modActivation.ts, modMerging.ts, preStartDeployHook.ts, stagingDirectory.ts**: Bluebird → native Promise throughout. `PromiseBB.each`/`mapSeries` → `for...of` + `await`. Predicate catches (`.catch(ErrorType, handler)`) → standard `.catch()` with `instanceof` checks.
- **modActivation.ts**: Fixed error handling scope — `method.prepare()` is now inside the `try/catch` that triggers `method.cancel()`, matching the original `.tapCatch()` coverage. Removed inner `try/catch` around `method.activate()` that would have silently continued on async failures (the old sync-only `try/catch` didn't catch promise rejections).
- **modMerging.ts**: Converted from deeply nested Bluebird chains to `async/await`. `PromiseBB.each` for archive merges → sequential `for...of` loops.
- **views/Settings.tsx**: `suggestPath` now stats the volume root (`path.parse(...).root`) instead of the full mod path, fixing ENOENT when the mod directory doesn't exist yet (e.g. first-time setup). Converted to `async/await`.
- **views/DeactivationButton.tsx, ModList.tsx**: Bluebird → native Promise. `dispatch(showDialog(...))` wrapped in `Promise.resolve()` since the thunk returns Bluebird (cannot change without breaking external extensions).
- **types/IDependency.ts, IDeploymentMethod.ts, InstallFunc.ts, TestSupported.ts**: `PromiseBB` → `PromiseLike` in type signatures.
- **util/activationStore.ts, basicInstaller.ts, dependencies.ts, deploy.ts, externalChanges.ts, queryGameId.ts, refreshMods.ts, removeMods.ts, sort.ts**: Bluebird → native Promise.

### Shared types & utilities (8 files)

- **types/IExtensionContext.ts**: `PromiseBB` → `PromiseLike` for public API types (`CheckFunction`, `AttributeExtractor`, `GameInfoQuery`, `MergeFunc`, `registerStartHook`, `once`). Uses `PromiseLike` (not `Promise`) because `Bluebird<T>` lacks `[Symbol.toStringTag]` required by `Promise<T>`, and external extensions still return Bluebird.
- **types/ITestResult.ts**: `PromiseBB` → `PromiseLike`.
- **util/api.ts**: Updated export types.
- **util/walk.ts**: Replaced broken `Promise.allSettled` (result.status was undefined in this environment) with `Promise.all` + per-item `.catch(() => null)`. Switched from `fsOrig.lstat` to `fs.lstatAsync` for retry support.
- **util/getFileList.ts**: `PromiseBB` → native Promise.
- **util/calculateFolderSize.ts**: Fixed pre-existing bug where manual directory recursion duplicated `walk`'s own recursion, producing doubled absolute paths (`basePath/basePath/subdir`). Removed the redundant recursion.
- **util/util.ts**: `makeQueue` internals converted to native Promise. Return type widened to `Promise<T | void>` to handle the `tryOnly` early-resolve path.
- **util/fs.ts**: Callback type update (`PromiseBB` → `PromiseLike`).
- **actions/notifications.ts**: Added comment explaining `showDialog` thunk returns Bluebird for backwards compatibility with external extensions.

### Adjacent extensions (8 files)

- **symlink_activator/index.ts**: Fixed return value mismatches where methods declared `Promise<void>` but returned `PromiseBB.resolve()`/`PromiseBB.reject()`.
- **symlink_activator_elevate/index.ts**: Same return value fixes. **remoteCode.ts**: Removed `@vortex/shared` import — this code is serialized via `.toString()` and runs in an elevated process outside webpack where module imports are unavailable.
- **hardlink_activator/index.ts**: Fixed `PromiseBB.reject()` in `linkFile` to `Promise.reject()` matching its `Promise<void>` return type.
- **move_activator/index.ts**: Type alignment.
- **gamemode_management/index.ts, util/queryGameInfo.ts**: Type alignment, `PromiseBB` → `Promise`.
- **profile_management/index.ts**: Type alignment.
- **test_runner/index.ts**: `PromiseBB` → `Promise`.

### Out-of-scope files touched minimally

- **download_management/FileAssembler.ts**: Updated `mQueue` field type and `addChunk`/`close` return types from `PromiseBB` to `Promise` to match `makeQueue`'s new return type.
- **download_management/types/IDownloadJob.ts**: `dataCB` return type `PromiseBB<boolean>` → `Promise<boolean>`.

## Bug fixes

- **refreshMods.ts**: Fixed inverted condition (`downloads[archiveId]` instead of `=== undefined`) that caused every mod with a valid archive to be flagged as orphaned on startup, triggering unnecessary archive reassignment.
- **calculateFolderSize.ts**: Fixed doubled path bug where `path.join(absolutePath, absolutePath)` produced invalid paths during post-install size calculation.
- **modActivation.ts**: Fixed error handling scope where `method.cancel()` would not be called if deployment failed after `prepare()` succeeded.
- **Settings.tsx**: Fixed ENOENT in `suggestPath` when mod directory doesn't exist yet.
- **remoteCode.ts**: Fixed `shared_1 is not defined` ReferenceError in elevated process.

## Please note

- `Promise.allSettled` does not work reliably in this environment (`result.status` is `undefined`). Replaced with `Promise.all` + per-item `.catch(() => null)` in `walk.ts`.
- `remoteCode.ts` (symlink_activator_elevate) is serialized via `.toString()` and runs in an elevated process outside webpack. Module imports are not available there — must use only inline code.
- `IDeploymentMethod` interface uses `PromiseLike<T>` (not `Promise<T>`) because `Bluebird<T>` lacks `[Symbol.toStringTag]` required by `Promise<T>`. External extensions still return Bluebird from these methods.
- `IExtensionContext` public API types (`GameInfoQuery`, `CheckFunction`, `MergeFunc`, etc.) also use `PromiseLike<T>` for the same reason. Callers wrap with `Promise.resolve()` where `.catch()` is needed.
- `showDialog` thunk returns Bluebird — cannot change without breaking external extensions. Call sites wrap in `Promise.resolve(dispatch(...))`.
- `LinkingDeployment.removeDeployedFile` uses duck-typing `(err as any)?.code` for ENOENT checks because `symlink_activator_elevate` rejects with deserialized IPC objects that are not `Error` instances.
- `calculateFolderSize` had a pre-existing bug where it manually recursed into subdirectories that `walk` already recurses into, producing doubled absolute paths. Removed the redundant recursion.
- `modActivation.deployMods`: the inner `try/catch` around `method.activate()` was removed — in the old Bluebird code the `try/catch` only caught synchronous errors, but with `await` it would also catch async rejections, changing abort-on-failure behavior.
- `finalize` in `LinkingDeployment` now uses `mapWithConcurrency(..., 50)` for `removeDeployedFile` calls to avoid overwhelming the elevated process IPC with hundreds of concurrent operations.
- `refreshMods.ts` had an inverted condition (`downloads[archiveId]` instead of `=== undefined`) that caused every mod with a valid archive to be flagged as orphaned.

please note:
- not all paths are using async/await - this will be done later in other MRs as I didn't want to add even more difficulty to the task.
- LinkingDeployment uses duck-typing - there's no way around this thanks to how the symlink elevation scripts are functioning over IPC.
- mapWithConcurrency function was added to replace Bluebird's concurrency options - could remove this but am unsure of performance impact. (will most definitely overwhelm the symlink IPC component)

closes https://linear.app/nexus-mods/issue/APP-39/phase-out-bluebird-dependency-in-mod-management-extension fixes https://linear.app/nexus-mods/issue/APP-52/fix-remotecode-ipc-for-symlink-deployment-elevation fixes https://linear.app/nexus-mods/issue/APP-55/suggestpath-breaks-when-modpath-becomes-unavailable

### DIFF
--- a/src/renderer/extensions/mod_management/InstallManager.ts
+++ b/src/renderer/extensions/mod_management/InstallManager.ts
@@ -4623,14 +4623,14 @@ class InstallManager {
     unattended: boolean,
     details: IInstallationDetails,
   ) {
-    if (result.instructions == null) {
+    if (result.instructions === null) {
       // this is the signal that the installer has already reported what went
       // wrong. Not necessarily a "user canceled" but the error handling happened
       // in the installer so we don't know what happened.
       return Promise.reject(new UserCanceled());
     }
 
-    if (result.instructions == null || result.instructions.length === 0) {
+    if (result.instructions === undefined || result.instructions.length === 0) {
       return Promise.reject(
         new ProcessCanceled("Empty archive or no options selected"),
       );


Aragas:
One issue with `PromiseLike` is that it only has the `.then` signature, but not `.catch`

Aragas:
Perhaps it might be a good idea to introduce a new interface that extends PromiseLike and adds the catch function, then expose it? I believe duck typing should still work in this case

src/renderer/extensions/mod_management/util/activationStore.ts
```diff
@@ -338,51 +345,51 @@ export function fallbackPurgeType(
...
       api,
       activator,
       gameId,
       typeId,
       modPaths[typeId],
       stagingPath,
-    ),
-  ).then(() => undefined);
+    );
+  }
```
Aragas, line 386: Are they equivalent? I assume we want to wait for them in serial order, not have a concurrent action going on?

IDCs: Yeah `fallbackPurge` should definitely be sequential. (which it is as part of the for loop - they won't run concurrently)

src/renderer/extensions/mod_management/InstallManager.ts
```diff
@@ -4610,15 +4623,15 @@ class InstallManager {
...
       // wrong. Not necessarily a "user canceled" but the error handling happened
       // in the installer so we don't know what happened.
-      return Bluebird.reject(new UserCanceled());
+      return Promise.reject(new UserCanceled());
     }
 
-    if (result.instructions === undefined || result.instructions.length === 0) {
-      return Bluebird.reject(
+    if (result.instructions == null || result.instructions.length === 0) {
+      return Promise.reject(
```
Aragas: Wrong semantics, you check here both null and undefined, we had previously differenct code path for them

src/renderer/extensions/mod_management/eventHandlers.ts
```diff
@@ -980,44 +964,29 @@ export function onRemoveMods(
               error: getErrorMessageOrDefault(error),
             });
           }
-        },
-        { concurrency: 5 },
```
Aragas, line 984: We dropped the concurrency limit, is this intentional?

src/renderer/extensions/mod_management/util/deploy.ts
```diff
@@ -330,53 +324,53 @@ function purgeModsImpl(
...
-    .finally(() => {
-      api.dismissNotification(notificationId);
-      api.store.dispatch(stopActivity("mods", "purging"));
-    });
+        }
+        if (err instanceof ProcessCanceled) {
+          return null;
+        }
+        throw err;
+      } finally {
```
Aragas, line 356: I think there a mistake - aren't we still executing code on finally().then()? Because in that case the new code will not execute `api.emitAndAwait("did-purge", profile.id))`

src/renderer/extensions/mod_management/views/DeactivationButton.tsx
```diff
@@ -96,46 +96,51 @@ class DeactivationButton extends ComponentEx<IProps, {}> {
...
+            "You need to select a deployment method in settings",
+            undefined,
+            false,
+          );
+          return;
         }
 
-        if (err.code === undefined && err.errno !== undefined) {
+        const errCode = getErrorCode(err);
+        if (errCode === null && (err as {errno?: number}).errno !== undefined) {
```
Aragas: why not undefined?

erri120:
> Perhaps it might be a good idea to introduce a new interface that extends PromiseLike and adds the catch function, then expose it? I believe duck typing should still work in this case

The goal is to move to native `Promise` the `PromiseLike` type is just a stepping stone and values of type `PromiseLike` should always be used with `Promise.resolve` to convert to native promises. Then `then` method on `PromiseLike` shouldn't be used at all.

src/renderer/extensions/mod_management/views/DeactivationButton.tsx
IDCs: brain fart moment

src/renderer/extensions/mod_management/eventHandlers.ts
IDCs: Yep, I noticed that the `removeMods` functionality was unnecessarily slow in 1.16 so thought that it makes sense to drop the limit and see how it goes; tested this quite a bit and it's much faster without any noticeable performance hit.

src/renderer/extensions/mod_management/util/deploy.ts
IDCs: Good catch

src/renderer/extensions/mod_management/InstallManager.ts
IDCs: Different code path for null specifically? can you link me the code?

I specifically tried to check for nullish (both null and undefined) across the file as I noticed that in some instances we were checking for one and not the other (when both were possible)

PR #21616, opened 2026-03-17
Bundle main project as ESM using rolldown

erri120:
Closes APP-12. Part of APP-15.

- Removes webpack from the main project
- Uses rolldown for bundling
- Changes the output from CJS to ESM


Aragas:
Does it mean that shared doesn't need CJS? And can we make the extension use ESM?
I know that we have a require monkey patching that I believe forces us to use CJS?

erri120:
> Does it mean that shared doesn't need CJS? And can we make the extension use ESM? I know that we have a require monkey patching that I believe forces us to use CJS?

We still need to use CJS on the renderer for now thanks to `nodeIntegration`. We also have to keep `type: commonjs` on the main `package.json` file because of that.

PR #21633, opened 2026-03-17
Spine Fixes

Aragas:
Part of APP-53
Default page should be active now when a new game is added
Also fixed the game spine not being active when selected and the default page not being shown Also fixed selecting the home icon in the spine does not switch it to the active state

### DIFF
--- a/src/renderer/views/components/Spine/SpineContext.tsx
+++ b/src/renderer/views/components/Spine/SpineContext.tsx
@@ -14,9 +14,9 @@ import { useDispatch, useSelector } from "react-redux";
 
 import type { IMainPage } from "../../../types/IMainPage";
 
-import { setNextProfile } from "../../../extensions/profile_management/actions/settings";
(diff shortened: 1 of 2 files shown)


src/renderer/views/components/Spine/SpineContext.tsx
```diff
@@ -146,9 +159,20 @@ export const SpineProvider: FC = ({ children }: { children: ReactNode }) => {
...
 
+  const selectGlobalPage = useCallback(
+    (pageId: string) => {
+      const actions: Action[] = [setOpenMainPage(pageId, false)];
+      if (activeProfileId !== undefined) {
+        actions.push(setNextProfile(undefined));
+      }
+      batchDispatch(api.store, actions);
+    },
+    [activeProfileId, api.store],
```
IDCs, line 170: The `api.store` dependency is harmless here but unnecessary.

PR #21669, opened 2026-03-18
Allow loading index.cjs for extensions

erri120:
Closes APP-83.

- Allow loading extension entry files named `index.cjs` instead of just `index.js`
- Output `index.cjs` for our bundled extensions instead of `index.js`


Aragas:
Might be a good idea to add in the migration guide that we support index.cjs

PR #21673, opened 2026-03-18
Clean up Stardew Valley support and add Linux SMAPI installation support

Sewer56:
https://github.com/user-attachments/assets/ceac802b-154a-498d-a097-9c9622ede8cc

## What changed

This PR cleans up the Stardew Valley extension so it is easier to understand, test, and maintain.

- Split the large Stardew Valley extension file into smaller modules for game setup, installers, SMAPI handling, config file syncing, UI, state, and runtime events
- Added support for installing SMAPI on Linux
- Fixed Linux-specific behaviour for launching Stardew Valley and opening SMAPI logs
- Added the first automated tests for the Stardew Valley extension, using Vitest to cover SMAPI installer behaviour across platforms
- Fixed all lint errors in the Stardew Valley extension
- Added stricter type checks for the extension
- Added contributor docs so the extension is easier to follow and change safely
- Updated extension logging to use the current renderer logger

## Why

The Stardew Valley extension had grown into one large file that was hard to follow and risky to change.

This change aims to keep the existing behaviour where possible, while making the code easier to work on and adding better coverage around the most fragile SMAPI install paths.

## Testing

- `pnpm --dir extensions/games/game-stardewvalley run typecheck:strict`
- `pnpm --dir extensions/games run test:stardew`
- `pnpm --dir extensions/games/game-stardewvalley run build`

## Draft status

This PR should stay in draft until general and platform specific regression testing has been completed on Windows.

## Known issue

- Related issue: [APP-88 - Disabled SMAPI still appears in Tools and Play](https://linear.app/nexus-mods/issue/APP-88/disabled-smapi-still-appears-in-tools-and-play)

### DIFF
--- a/extensions/games/game-stardewvalley/package.json
+++ b/extensions/games/game-stardewvalley/package.json
@@ -18,6 +18,7 @@
     "game": "Stardew Valley"
   },
   "dependencies": {
+    "@vortex/shared": "workspace:*",
     "react": "catalog:",
     "react-bootstrap": "catalog:",
     "react-i18next": "catalog:",
@@ -35,7 +36,6 @@
     "@types/react-redux": "catalog:",
     "@types/relaxed-json": "catalog:",
     "@types/semver": "catalog:",
-    "@vortex/shared": "workspace:*",
     "copyfiles": "catalog:",
     "electron": "catalog:",
     "modmeta-db": "catalog:",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2256,9 +2256,9 @@ importers:
 
   extensions/games/game-stardewvalley:
     dependencies:
-      bluebird:
-        specifier: 'catalog:'
-        version: 3.7.2
+      '@vortex/shared':
+        specifier: workspace:*
+        version: link:../../../src/shared
       react:
         specifier: 'catalog:'
         version: 16.14.0
@@ -2287,9 +2287,6 @@ importers:
       '@nexusmods/nexus-api':
         specifier: 'catalog:'
         version: https://codeload.github.com/Nexus-Mods/node-nexus-api/tar.gz/4192c0c9f34306c2167e258dd4fef773af406161
-      '@types/bluebird':
-        specifier: 'catalog:'
-        version: 3.5.20
       '@types/node':
         specifier: 'catalog:'
         version: 22.19.11
@@ -2308,9 +2305,6 @@ importers:
       '@types/semver':
         specifier: 'catalog:'
         version: 7.7.1
-      '@vortex/shared':
-        specifier: workspace:*
-        version: link:../../../src/shared
       copyfiles:
         specifier: 'catalog:'
         version: 2.4.1


Sewer56:
Checklist:

# Game is Discovered (Steam)

Steps:
1. Install Stardew Valley from Steam.
2. Open Vortex and go to `Games`.
3. Confirm Stardew Valley is detected.
4. Click `Manage`.
5. Open the game page.

- [x] Windows 11
- [x] Linux (Native)

# Game Version is Shown

Steps:
1. Install the game.
2. Open `Games`.
3. Find Stardew Valley entry.
4. Click the little `3 dot` icon top right for game info

- [x] Windows 11
- ***N/A*** ❌

On Windows, we pull game version from binary but no such string here exists.
May be worth adding a ticket for game specific game detection thing

# Game version refreshes after a game update

Not tested.

# Install SMAPI via notification

Steps:
1. Manage Stardew Valley with no SMAPI deployed.
2. Open the Stardew Valley page.
3. Look for the warning that SMAPI is required.
4. Install SMAPI via `Get SMAPI` on notification.

- [x] Windows 11
- [x] Linux (Native)

# Install a mod via `nxm://` protocol

Steps:
1. Start with Stardew Valley already managed.
2. Download a mod from Nexus using `Mod Manager Download`.
3. Confirm the download appears in Vortex.
4. Install the mod.
5. Deploy if needed.
6. Launch the game.

Expected:
- The mod downloads into Vortex.
- The mod installs cleanly.
- The game launches with the mod active.

- [x] Windows 11
- [x] Linux (Native)

# Install a collection

Steps:
1. Start with Stardew Valley already managed.
2. Add a Stardew Valley collection from Nexus.
3. Continue the collection flow in Vortex.
4. Install the collection.
5. Launch the game.

Expected:
- The collection installs without unexpected errors.
- The game launches successfully afterward.

- [x] Windows 11
- [x] Linux (Native) ⚠️

Tested with `Stardew Valley Very Expanded`.

⚠️

image

Needs path library for patching files.
Outside of patching, mods in collection install fine.

# Launch through SMAPI and primary-tool behavior

What this covers:
- SMAPI becomes the primary tool after deploy

Steps:
1. Install and deploy SMAPI.
2. Launch Stardew Valley from Vortex.
3. Confirm the game starts through SMAPI.

- [x] Windows 11
- [x] Linux (Native)

# Removing SMAPI hides SMAPI Tool from UI

❌ Broken in Vortex itself.

Reported as separate issue, check Linear.

# Stardew-specific 'Compatibility' Column

Steps:
1. Make `Stardew Valley` the active game.
2. Open the `Mods` settings page for the game.
3. Check the `Compatibility` column.

Expected:
- Compatibility column shows ✅ or ℹ️ (info)

- [x] Windows 11
- [x] Linux (Native)

⚠️ TODO: Detect game version accurately here, per above. Needed for this column to work well.

# Stardew-specific 'Game preferences' Left Menu Item

Steps:
1. Make `Stardew Valley` the active game.
2. Open the `Game Preferences` view

Expected:
- Shows 2 options
    - `Use recommendations from mod manifests`
    - `Manage SDV mod Configuration files`

- [x] Windows 11
- [x] Linux (Native)

# SMAPI log button

Steps:
1. Make `Stardew Valley` the active game.
2. Run it with SMAPI.
3. Click `SMAPI Log`.
4. Confirm the log text is shown.
5. Log text is in clipboard

- [x] Windows 11
- [x] Linux (Native)

extensions/games/game-stardewvalley/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.3.9",
   "description": "Support for Stardew Valley",
   "scripts": {
-    "_assets": "copyfiles -u 1 -f ./assets/* dist",
+    "_assets": "copyfiles -u 1 -f ./assets/* dist/assets && copyfiles -u 2 -f ./src/ui/sdvstyle.scss dist/ui",
     "_build": "node build.mjs && pnpm run _assets && pnpm extractInfo",
+    "typecheck:strict": "pnpm exec tsc -p tsconfig.strict.json",
```
erri120: No need to have a separate typescript that's different from everything else. Rename `tsconfig.strict.json` to just `tsconfig.json` and change the command name to be `typecheck`.

extensions/games/tsconfig.test.json
erri120, line 1: Should be removed.

extensions/games/tsconfig.json
erri120, line 1: Should be removed.

extensions/games/package.json
erri120, line 1: Should be removed. Each extensions should be standalone with its own package.

extensions/games/game-stardewvalley/tsconfig.json
erri120, line 1: Rename to `tsconfig.json`

extensions/games/game-stardewvalley/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+  "include": [
+    "./src/**/*.ts",
+    "./src/**/*.tsx"
+  ],
+  "exclude": [
+    "node_modules",
+    "dist",
+    "out"
+  ]
+}
```
erri120, line 9:
```suggestion
{
  "$schema": "https://www.schemastore.org/tsconfig.json",
  "extends": "../../tsconfig.extensions.json",
  "compilerOptions": {
    "strictNullChecks": true,
    "noUncheckedIndexedAccess": true
  },
  "include": [
    "src"
  ]
}
```

extensions/games/game-stardewvalley/vitest.config.ts
```diff
@@ -0,0 +1,17 @@
...
+    alias: {
+      'vortex-api': path.resolve(__dirname, '__mocks__/vortex-api.ts'),
+    },
+  },
+  test: {
+    name: 'game-stardewvalley',
+    environment: 'node',
+    include: ['__tests__/**/*.test.ts'],
+    exclude: ['node_modules', 'dist', 'out'],
+  },
```
erri120:
```suggestion
  resolve: {
    alias: {
      'vortex-api': path.resolve(import.meta.dirname, '__mocks__/vortex-api.ts'),
    },
  },
  test: {
    environment: 'node',
    include: ['__tests__/**/*.test.ts'],
  },
```

extensions/games/game-stardewvalley/__mocks__/vortex-api.ts
erri120, line 1: File should be empty. With the alias in the config file you want the actual mocks to be in the unit test files themselves.

extensions/games/game-stardewvalley/src/helpers.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Provides small shared helpers for Bluebird wrapping and error normalization.
+ */
+import Bluebird from 'bluebird';
+
+/** Wraps an async function and returns a Bluebird-compatible function. */
+export function toBlue<T>(func: (...args: any[]) => Promise<T>): (...args: any[]) => Bluebird<T> {
+  return (...args: any[]) => Bluebird.resolve(func(...args));
+}
```
erri120: As mentioned in the Vortex refactoring workshop, we never want to convert from Bluebird to native Promises.

extensions/games/game-stardewvalley/src/helpers.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+  if (typeof err === 'function') {
+    return err.toString();
+  }
+  try {
+    return (err as { toString: () => string }).toString();
+  } catch (_err) {
+    return Object.prototype.toString.call(err);
+  }
+}
```
erri120: Use methods provided in the `@vortex/shared` package instead.

extensions/games/game-stardewvalley/src/smapi/index.ts
erri120, line 1: No barrel files.

extensions/games/game-stardewvalley/__mocks__/vortex-api.ts
Sewer56: You want me to duplicate this over every single file? Curious here.

extensions/games/game-stardewvalley/src/helpers.ts
Sewer56: @erri120 Open a separate ticket for Bluebird fixup (even if trivial).
Can do that in a follow-up.

This is pre-existing code; I didn't want to touch it as doing so was out of scope.

I've only moved it around; in order to make the extension more maintainable; as the original state was a bit difficult to develop and test on.

extensions/games/game-stardewvalley/__mocks__/vortex-api.ts
erri120: The vitest config makes sure that `vortex-api` gets resolved to _something_ but we don't actually care what that is. What we do care about are mocks for functions that get called. So we can have a `vortex-api.ts` file used for resolving the module itself and then each test file declares their mocks for that module at the top:

```typescript
vi.mock(import("vortex-api"), () => {
  return {
    log: vi.fn()
  }
});
```

This way the `vortex-api` module gets resolved without vitest complaining and we can mock whatever we want on this and keep the mocks local to the tests.

extensions/games/package.json
Sewer56: Some clarification is wanted here.
CC. @insomnious 

There's 2 ways to go here:
- Each extension is a full, standalone, self contained project
- Each extension is part of the codebase

The way you've explained it to me before is that in our current extensions ecosystem; all extensions inherit the dependencies from the `renderer` process. And right now at runtime, this is true. i.e. What's available in the `renderer` process; is what the extensions get.

Moving the extensions to a game specific `package.json` would essentially make the game like a more standalone project. Normally that would be fine, but at that point we should strive for a model where each extension is truly isolated and a 'real' project of its own.

So I did a bit of a dig to find out the answers. Currently I see `extensions/copy-extension.mjs`, which copies the `dist` folder but not the `node_modules`. So, dependencies are only deployed if their code happens to be bundled into `index.cjs`.

After some more digging I found `scripts/extensions-rolldown.mjs`
```typescript
/** @returns {<string[]} */
export function getExternals() {
  const rawPackageJson = readFileSync(packageJsonPath, "utf8");
  const packageJson = JSON.parse(rawPackageJson);

  const injectedExternals = ["electron", "vortex-api"];
```
It appears this was merged *very* recently (last week); I only got it after merging main.

If I'm understanding this right, this recent change made it so any dependency that's not in:
- `src/main/package.json`
- `electron`
- `vortex-api`
- `node.js`

Will be embedded inside the generated `index.cjs`.
[Also on a sidenote, why does this code not have any documentation/comments in the file etc.? This is very important build stuff!! Edit: I added a tiny bit myself]

That's fine. And seems to lean towards the `Each extension is a full, standalone, self contained project`. Can you confirm?

Now, in the current case, I added this `package.json` stub such that the LSP, typechecks etc. work as expected for the extensions in the `games` module. It was as a temporary measure; from before games module was inlined. 

Given we now support embedding dependencies; I will add only the Stardew required dependencies to Stardew (presumably this is the request). But ***I'd like for the 'strategy' to be documented somewhere***. Because this stuff is changing from week to week and it's unclear.

Maybe also a ticket to fix lints/deps in other extensions, down the road.

extensions/games/game-stardewvalley/package.json
```diff
@@ -16,6 +18,193 @@
...
+    "vortex-parse-ini": "catalog:",
+    "vortexmt": "catalog:",
+    "wholocks": "catalog:",
+    "winapi-bindings": "catalog:",
+    "winston": "catalog:",
+    "write-file-atomic": "catalog:",
+    "ws": "catalog:",
+    "xml2js": "catalog:",
+    "xxhash-addon": "catalog:",
+    "zod": "catalog:"
```
erri120: That's a lot of dependencies, this should only be what stardew valley actually needs.

Sewer56: Wait, one sec. I had a more proper file setup- seems I dropped it when I reverted a commit to strip it down a bit.

Thanks for catching it. I missed-

Sewer56: Sorted.

extensions/games/game-stardewvalley/src/helpers.ts
Sewer56: I just nuked as much of Bluebird as I can for the time being, mm.

Sewer56:
All resolved; even adjacent stuff I suggested putting in future ticket. If anything else shows up, please raise an issue.

Issue #17848, opened 2025-07-03
[BUG] - Executable does not set properly when switching a game in Vortex from one game store to another with different executable names


ChemGuy1611:
**Describe the bug**
When switching a game install from a location with one executable name to one with a different exe name (i.e. moving from Game Pass to Steam), Vortex does not seem to accept the new, dynamically set executable name and "sticks" to the one associated with the original install location, even though the gamePath is updated. This causes the Vortex launch button to fail with the below error.

Workaround: I am able to resolve this by renaming the new installation folder briefly, then restarting Vortex, which will cause the game to no longer be found. I can then rename the folder back to its original name and restart Vortex again. It now detects the correct executable. 

This example screenshot came as a result of me changing the structure of an in-development extension, but it is identical to the errors users of other exensions have reported. 
![Image](https://github.com/user-attachments/assets/d162c1c0-c5d4-440b-87e3-1574c074b661)

Forum report here: https://forums.nexusmods.com/topic/13518607-error-executable-doesnt-exist-gamelaunchhelperexe-expedition-33-steam-version/#comment-130517598

I have seen reports of this issue on the forum for several games, all with basically identical means of setting the executable: STALKER 2, Clair Obscur, Palworld.

Note that this is NOT exclusive to Game Pass versions. Those are just convenient examples. 

**To Reproduce**
1. Start managing the Game Pass version of any game with multiple different exes depending on the game store (any recent UE game on Game Pass, such as Clair Obscur).
2. Close Vortex and Uninstall the Game Pass version of the game, then install the Steam version. 
3. Restart Vortex, new discovery location is detected. 
4. Try to launch the game and see the error that the executable is not found because it is still looking for "gamelaunchhelper.exe" rather than the Steam executable. 
5. Unmanaging the game and restarting Vortex does NOT resolve the issue.

**Expected behavior**
game.executable should update dynamically when context.registerGame is called during Vortex startup. 

**Platform (please complete the following information):**
 - OS: Windows 11 24H2
 - Vortex Version 1.14.6

**Additional Context**

Taking a look at the source code, I see there's both a `discovery.executable` and `game.executable`. Could it be that one of these is not updating properly?
https://github.com/Nexus-Mods/Vortex/blob/9222a27583a45c7cf76db172f487ff2d25802e0d/src/extensions/gameversion_management/util/getGameVersion.ts#L19

Here is the code that is used to set the executable in the context.registerGame call. Note this is basically identical to Palworld. 

```
[...]

PR #21675, opened 2026-03-18
fixed stale discovery data when setting game location manually

IDCs:
Note that some users want to switch between different game stores when managing their game (for some reason) so we intentionally do not purge the old game directory as they may want to switch back.

fixes https://linear.app/nexus-mods/issue/APP-85/mods-are-not-installed-in-manually-set-locations fixes https://github.com/Nexus-Mods/Vortex/issues/17848

### DIFF
--- a/src/renderer/extensions/gamemode_management/reducers/settings.ts
+++ b/src/renderer/extensions/gamemode_management/reducers/settings.ts
@@ -1,4 +1,10 @@
+import * as _ from "lodash";
+
+import type { IDiscoveredTool } from "../../../types/IDiscoveredTool";
 import type { IReducerSpec } from "../../../types/IExtensionContext";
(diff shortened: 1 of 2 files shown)


src/renderer/extensions/gamemode_management/reducers/settings.ts
```diff
@@ -53,15 +62,21 @@ export const settingsReducer: IReducerSpec<ISettingsGameMode> = {
       return deleteOrNop(state, ["discovered", id, "path"]);
     },
     [actions.setGamePath as any]: (state, payload) => {
-      const input = {
+      const input: any = {
```
Aragas: Feeling risky today? :joy:

IDCs: Fair point, I'm rushing through these tickets and am missing small bits.

PR #21684, opened 2026-03-18
Use Thumbnail API

Aragas:
Part of APP-53
Use the new nexus mods thumbnail API
Persist the fetched images in a local cache
Use a "letter avatar" if nothing is found

Not sure if you guys will like the cache implementation, but my idea was that we should try to avoid fetching this data each time.

Another idea would be to distribute the thumbnail images locally with Vortex instead

### DIFF
--- a/src/renderer/views/components/Spine/utils.ts
+++ b/src/renderer/views/components/Spine/utils.ts
@@ -1,3 +1,4 @@
+import { getErrorCode } from "@vortex/shared";
 import { readFileSync } from "fs";
 import * as path from "path";
 import { useCallback, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


src/renderer/views/components/Spine/utils.ts
```diff
@@ -1,26 +1,229 @@
...
+const downloadImage = async (imageUrl: string, gameId: string): Promise<string> => {
+  const response = await fetch(imageUrl);
+  if (!response.ok) {
+    throw new Error(`HTTP ${response.status}`);
+  }
+  const ext = path.extname(new URL(imageUrl).pathname) || ".jpg";
+  const buffer = Buffer.from(await response.arrayBuffer());
+  const filePath = cachedImagePath(gameId, ext);
+  await ensureDirWritableAsync(iconCacheDir());
+  await writeFileAsync(filePath, buffer);
```
IDCs: You need to handle EEXIST errors here.

Aragas: Done

PR #21698, opened 2026-03-19
fixed native frame specific values not being read correctly from state

IDCs:
- readPersistedValue now reconstrcuts objects from leaf keys when a direct key lookup fails
- fixed native frame menu bar not appearing when customTitlebar is disabled

fixes https://linear.app/nexus-mods/issue/APP-100/custom-window-title-bar-not-working-in-200


IDCs:
I was being too optimistic about direct key lookups 😋

PR #21810, opened 2026-03-23
Playwright E2S Tests

Aragas:
Part of APP-82

Playwright e2e tests are now in a package, added kinda a headless mode
Using CommonJS for the output due to Playwright
Converted from BrowserView to WebContentsView

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,10 +1,6 @@
 name: E2E Tests
 
 on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
   workflow_dispatch:
 
 concurrency:
@@ -23,7 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     # E2E aren't stable yet. Windows takes too long to execute, Linux is experimental
     continue-on-error: true
-    timeout-minutes: 30
+    timeout-minutes: 60
 
     steps:
       - uses: actions/checkout@v6
@@ -80,13 +76,13 @@ jobs:
 
       - name: Run E2E tests (Windows)
         if: runner.os == 'Windows'
-        timeout-minutes: 10
+        timeout-minutes: 30
         continue-on-error: true
         run: pnpm e2e
 
       - name: Run E2E tests (Linux)
         if: runner.os == 'Linux'
-        timeout-minutes: 10
+        timeout-minutes: 30
         continue-on-error: true
         run: xvfb-run pnpm e2e
 


packages/e2e/scripts/patch-playwright-electron.mjs
erri120, line 1: Use a [PNPM patch](https://pnpm.io/cli/patch) instead of DIY script. `pnpm patch` exists for exactly this reason.

packages/e2e/playwright.config.ts
```diff
@@ -0,0 +1,21 @@
...
+export default defineConfig({
+  testDir: './tests',
+  timeout: 60_000,
+  expect: {
+    timeout: 10_000,
+  },
+  retries: 0,
+  // Each test file runs in its own worker. Tests within the same file share
+  // the Electron instance when using test.describe.serial or the same fixture.
+  workers: 1,
```
erri120: Why are we disabling parallelization for tests?

packages/e2e/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "strict": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "outDir": "dist",
+    "rootDir": "."
+  },
+  "include": ["tests/**/*.ts", "fixtures/**/*.ts", "playwright.config.ts"]
+}
```
erri120, line 14:
```suggestion
{
  "$schema": "https://www.schemastore.org/tsconfig.json",
  "extends": "../../tsconfig.strict.json",
  "include": [
    "tests",
    "fixtures",
    "playwright.config.ts"
  ],
  "compilerOptions": {
    "composite": true,
    "noEmit": true,
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "bundler",
    "lib": [
      "ESNext"
    ],
    "types": [],
  }
}
```

packages/e2e/playwright.config.ts
Aragas: That was a leftover before I've added custom user directories, good catch

packages/e2e/scripts/patch-playwright-electron.mjs
Aragas: That's a cool feature

packages/e2e/playwright.config.ts
Aragas: Actually, might still not work, we have that mutex that will break parallel execution

packages/e2e/playwright.config.ts
```diff
@@ -0,0 +1,26 @@
...
+export default defineConfig({
+  testDir: "./tests",
+  timeout: 60_000,
+  expect: {
+    timeout: 10_000,
+  },
+  retries: 1,
+  // Each worker launches its own Electron instance with isolated user data.
+  // CI: Windows runners are slower (1 worker), Linux can handle 2.
+  // Local: 4 workers.
```
erri120, line 12: Magic number? Default according to the docs is half of the number of CPU cores.

packages/e2e/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "type": "module",
+  "scripts": {
+    "test": "playwright test",
+    "test:headed": "playwright test --headed",
+    "test:debug": "playwright test --debug",
+    "test:report": "playwright show-report"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.58.2",
+    "playwright": "^1.58.2",
```
erri120: Not sure how these got here but they should be in the catalog instead.

packages/e2e/fixtures/vortex-app.ts
```diff
@@ -0,0 +1,222 @@
...
+  type ElectronApplication,
+  type Page,
+} from '@playwright/test';
+import path from 'node:path';
+import fs from 'node:fs';
+import os from 'node:os';
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
erri120: No need for this, use `import.meta.dirname` directly.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,105 @@
+name: E2E Tests
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
erri120: I'd like to remove this for now. I don't think we should run E2E tests on ever commit of every PR.

packages/e2e/playwright.config.ts
Aragas: Default is not scaling properly. I have 16 cores, so 8 workers is a death sentence
CI scaling is also not good. Windows exposes 4 cores, yet even with 1 it struggles. For Linux it's 4, so the default value is the same

.github/workflows/e2e.yml
Aragas: I've made them optional

packages/e2e/package.json
Aragas: Good catch, fixed

packages/e2e/fixtures/vortex-app.ts
Aragas: Fixed

.github/workflows/e2e.yml
erri120: Only the Linux step is marked as optional. E2E tests are slow, flaky, and resource-intensive compared to unit tests. Running them on every commit of every PR means we end up with slow feedback loops, flakiness frustration, and high CI costs.

We can figure out a E2E strategy for PRs later, maybe run a small suit of critical smoke tests but not the full suit.

Aragas: I do see the point here. Tbh that was something I remember suggesting myself - like run the E2S tests at the end of the day, since we expect way more tests. Okay, will disable the triggers

PR #21830, opened 2026-03-24
fixed UIV2 related profile page issues

IDCs:
- profile page now enabled by default for UIV2
- enhanced profile switch logic for newly created profiles as part of a collection installation

fixes https://linear.app/nexus-mods/issue/APP-131/enable-profiles-by-default-in-vortex-20 fixes https://linear.app/nexus-mods/issue/APP-123/uiv2-profile-management-page-does-not-appear-automatically

### DIFF
--- a/extensions/collections/src/views/InstallDialog/InstallStartDialog.tsx
+++ b/extensions/collections/src/views/InstallDialog/InstallStartDialog.tsx
@@ -419,7 +419,7 @@ function mapStateToProps(
     userInfo,
     nextProfileId: state.settings.profiles.nextProfileId,
     collectionsInstallWhileDownloading,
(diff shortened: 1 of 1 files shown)


extensions/collections/src/views/InstallDialog/InstallStartDialog.tsx
```diff
@@ -416,6 +419,7 @@ function mapStateToProps(
     userInfo,
     nextProfileId: state.settings.profiles.nextProfileId,
     collectionsInstallWhileDownloading,
+    useModernLayout: state.settings.window.useModernLayout ?? false,
```
Aragas: default value should be true

PR #21839, opened 2026-03-24
feat: add type-safe path system with resolver pipeline

halgari:
## Overview

This PR introduces a new **type-safe path system** for Vortex, extracted into two standalone pnpm workspace packages:

- **`packages/paths`**: core path types, resolver interfaces, and platform-agnostic logic
- **`packages/paths-node`**: Node.js-specific filesystem implementation

The goal is to make path handling in Vortex safe-by-construction: paths are branded types that can only be created through validated constructors, and resolution (turning a logical anchor + relative path into an OS-specific absolute path) is explicit and pipeline-driven.

### Branded Path Types (`packages/paths/src/types.ts`)

Three core types prevent mixing up path kinds at compile time:

- **`RelativePath`**: forward-slash normalized, sanitized relative paths (no `..`, no drive letters). Created via `RelativePath.of(str)` with Zod validation.
- **`ResolvedPath`**: absolute OS-specific paths (output of resolution). Only produced by resolvers, not constructed directly.
- **`Anchor`**: interned Symbols representing named resolution starting points (e.g. `"mods"`, `"game"`, `"appData"`).

### Resolver Pipeline (`packages/paths/src/resolvers/`)

Resolvers map `(Anchor, RelativePath)` to `ResolvedPath`. They compose into chains:

```
MappingResolver("mods")        <- knows "mods" anchor, maps to parent-relative path
    └── UnixResolver / WindowsResolver   <- terminal: produces final OS absolute path
```

- **`BaseResolver<ValidAnchors>`**: abstract base with type-safe `PathFor<A>()` for compile-time anchor validation
- **`MappingResolver`**: maps logical anchors to paths relative to a parent resolver
- **`UnixResolver`** / **`WindowsResolver`**: terminal resolvers that produce POSIX or Win32 absolute paths
- **`IResolverBase`**: non-generic interface used for parent references, avoiding TypeScript contravariance issues with `IResolver<A>`

**Forward resolution**: A resolver only handles anchors it owns, with no parent delegation. Unknown anchors throw immediately.

**Reverse resolution** (`tryReverse`): Converts an absolute OS path back to an `(Anchor, RelativePath)` pair by checking each resolver's registered anchors. Preserves original filesystem casing.

### FilePath (`packages/paths/src/FilePath.ts`)

`FilePath` is the primary user-facing value object. It wraps `(Anchor, RelativePath)` and resolves lazily to an OS path via its attached resolver. Provides helpers for joining, getting extensions, changing extensions, etc.

### IFilesystem (`packages/paths/src/IFilesystem.ts`)

Platform-aware filesystem interface. OS detection logic lives here rather than in resolvers, keeping resolvers as pure mapping logic. `NodeFilesystem` (`packages/paths-node`) implements this against `node:fs`.

## Package Structure

```
packages/
  paths/
    src/
      types.ts          <- RelativePath, ResolvedPath, Anchor branded types + Zod schemas
      IResolver.ts      <- IResolverBase / IResolver<A> interfaces
      IFilesystem.ts    <- Platform-aware filesystem interface
      FilePath.ts       <- Primary path value object
      pathUtils.ts      <- Cross-platform path module helpers
      resolvers/
        BaseResolver.ts     <- Abstract base with type-safe PathFor
        MappingResolver.ts  <- Logical anchor -> parent-relative mapping
        UnixResolver.ts     <- POSIX terminal resolver
        WindowsResolver.ts  <- Win32 terminal resolver
    __tests__/          <- Full test suite
  paths-node/
    src/
      NodeFilesystem.ts <- Node.js fs-backed IFilesystem implementation
```

## Test Coverage

Comprehensive test suite covering:
- Branded type construction and validation (`types.test.ts`)
- `FilePath` API (`FilePath.test.ts`)
- Unix and Windows resolver behavior (`UnixResolver.test.ts`, `WindowsResolver.test.ts`)
- Reverse resolution round-trips (`reverse-resolution.test.ts`, `multi-resolver-roundtrip.test.ts`)
- Path normalization edge cases (`path-normalization.test.ts`)
- Filesystem integration (`filesystem.test.ts`, `integration.test.ts`)

### DIFF
--- a/packages/paths-node/package.json
+++ b/packages/paths-node/package.json
@@ -19,10 +19,11 @@
   },
   "dependencies": {
     "@vortex/paths": "workspace:*",
-    "fs-extra": "^11.2.0"
+    "fs-extra": "catalog:"
   },
   "devDependencies": {
-    "@types/fs-extra": "^11.0.4",
+    "@types/fs-extra": "catalog:",
+    "@types/node": "catalog:",
     "vitest": "catalog:"
   }
 }
--- a/packages/paths/package.json
+++ b/packages/paths/package.json
@@ -24,7 +24,8 @@
     "zod": "^3.24.2"
   },
   "devDependencies": {
-    "typescript": "^5.8.3",
+    "@types/node": "catalog:",
+    "typescript": "catalog:",
     "vitest": "catalog:"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3816,12 +3816,15 @@ importers:
         specifier: ^3.24.2
         version: 3.25.76
     devDependencies:
+      '@types/node':
+        specifier: 'catalog:'
+        version: 22.19.11
       typescript:
-        specifier: ^5.8.3
+        specifier: 'catalog:'
         version: 5.9.3
       vitest:
         specifier: 'catalog:'
-        version: 4.1.0(@opentelemetry/api@1.9.0)(@types/node@25.5.0)(happy-dom@20.8.3)(jsdom@20.0.3)(vite@8.0.0(@types/node@25.5.0)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0))
+        version: 4.1.0(@opentelemetry/api@1.9.0)(@types/node@22.19.11)(happy-dom@20.8.3)(jsdom@20.0.3)(vite@8.0.0(@types/node@22.19.11)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0))
 
   packages/paths-node:
     dependencies:
@@ -3829,15 +3832,18 @@ importers:
         specifier: workspace:*
         version: link:../paths
       fs-extra:
-        specifier: ^11.2.0
-        version: 11.3.3
+        specifier: 'catalog:'
+        version: 9.1.0
     devDependencies:
       '@types/fs-extra':
-        specifier: ^11.0.4
+        specifier: 'catalog:'
         version: 11.0.4
+      '@types/node':
+        specifier: 'catalog:'
+        version: 22.19.11
       vitest:
         specifier: 'catalog:'
-        version: 4.1.0(@opentelemetry/api@1.9.0)(@types/node@25.5.0)(happy-dom@20.8.3)(jsdom@20.0.3)(vite@8.0.0(@types/node@25.5.0)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0))
+        version: 4.1.0(@opentelemetry/api@1.9.0)(@types/node@22.19.11)(happy-dom@20.8.3)(jsdom@20.0.3)(vite@8.0.0(@types/node@22.19.11)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0))
 
   packages/vortex-api:
     dependencies:
@@ -16393,14 +16399,6 @@ snapshots:
     optionalDependencies:
       vite: 8.0.0(@types/node@22.19.11)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0)
 
-  '@vitest/mocker@4.1.0(vite@8.0.0(@types/node@25.5.0)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0))':
-    dependencies:
-      '@vitest/spy': 4.1.0
-      estree-walker: 3.0.3
-      magic-string: 0.30.21
-    optionalDependencies:
-      vite: 8.0.0(@types/node@25.5.0)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0)
-
   '@vitest/pretty-format@4.1.0':
     dependencies:
       tinyrainbow: 3.1.0
@@ -23218,36 +23216,6 @@ snapshots:
     optionalDependencies:
       '@opentelemetry/api': 1.9.0
       '@types/node': 22.19.11
-      happy-dom: 20.8.3
-      jsdom: 20.0.3
-    transitivePeerDependencies:
-      - msw
-
-  vitest@4.1.0(@opentelemetry/api@1.9.0)(@types/node@25.5.0)(happy-dom@20.8.3)(jsdom@20.0.3)(vite@8.0.0(@types/node@25.5.0)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0)):
-    dependencies:
-      '@vitest/expect': 4.1.0
-      '@vitest/mocker': 4.1.0(vite@8.0.0(@types/node@25.5.0)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0))
-      '@vitest/pretty-format': 4.1.0
-      '@vitest/runner': 4.1.0
-      '@vitest/snapshot': 4.1.0
-      '@vitest/spy': 4.1.0
-      '@vitest/utils': 4.1.0
-      es-module-lexer: 2.0.0
-      expect-type: 1.3.0
-      magic-string: 0.30.21
-      obug: 2.1.1
-      pathe: 2.0.3
-      picomatch: 4.0.3
-      std-env: 4.0.0
-      tinybench: 2.9.0
-      tinyexec: 1.0.4
-      tinyglobby: 0.2.15
-      tinyrainbow: 3.1.0
-      vite: 8.0.0(@types/node@25.5.0)(esbuild@0.27.4)(jiti@2.6.1)(sass@1.97.3)(terser@5.46.0)
-      why-is-node-running: 2.3.0
-    optionalDependencies:
-      '@opentelemetry/api': 1.9.0
-      '@types/node': 25.5.0
       happy-dom: 20.8.3
       jsdom: 20.0.3
     transitivePeerDependencies:


packages/paths/__tests__/FilePath.test.ts
```diff
@@ -0,0 +1,324 @@
+/**
+ * Tests for FilePath class
+ */
+
+import { describe, test, expect, beforeEach, jest } from '@jest/globals';
```
Sewer56: We're meant to be putting the tests next to the source code now, after the vitest merge.
Don't worry, I'll fix it up myself 😉

Sewer56:
Resolved merge conflict and CI fail- sorting out the rest as I review.

src/renderer/jest.config.mjs
```diff
@@ -12,8 +12,16 @@ const config = {
     "^../util/ComponentEx$": "<rootDir>/src/__mocks__/ComponentEx.js",
     "^@/(.*)$": "<rootDir>/src/$1",
   },
-  testRegex: "(/src/__tests__/.*)\\.(js|jsx|ts|tsx)$",
-  testPathIgnorePatterns: ["/node_modules/", "/dist/", "/out/"],
+  testRegex: "(/__tests__/.*)\\.(js|jsx|ts|tsx)$",
```
Sewer56: No longer valid with vitest based setup; with files alongside source files. Cleaning up.

pnpm-workspace.yaml
```diff
@@ -2,6 +2,7 @@
 
 packages:
   - ./src/shared
```
Sewer56, line 4: Duplicate array item here, fixing.

Sewer56 (commented):
Fixing as I find stuff.
I just accidentally started review, mhm

packages/paths-node/src/NodeFilesystem.ts
```diff
@@ -0,0 +1,193 @@
...
+  // ========================================================================
+  // Write Operations
+  // ========================================================================
+
+  async writeFile(
+    filePath: ResolvedPath,
+    data: string | Uint8Array,
+    encoding?: string,
+  ): Promise<void> {
+    await fs.ensureDir(path.dirname(filePath as string));
```
Sewer56, line 56: Should ideally have an opt out- but I'll leave it be for simplicity until we determine it's a bottleneck down the road.

packages/paths-node/src/NodeFilesystem.test.ts
```diff
@@ -0,0 +1,498 @@
...
+    });
+
+    test("sep matches path.sep", () => {
+      expect(nodeFs.sep).toBe(sep);
+    });
+  });
+
+  describe("normalizePath", () => {
+    test("uses Windows or Unix separators when normalizing", () => {
+      const input =
```
Sewer56, line 173: The two tests here only worked on Unix paths. (Previously failed CI)
I fixed it so it accounts for both Unix and Windows.

packages/paths/src/types.ts
```diff
@@ -0,0 +1,540 @@
...
+ * - Must start with a dot
+ * - Cannot contain path separators
+ * - Normalized to lowercase
+ */
+export const ExtensionSchema = z
+  .string()
+  .refine((s) => s.startsWith("."), {
+    message: "Extensions must start with .",
+  })
+  .refine((s) => !s.includes("/") && !s.includes("\\"), {
```
Sewer56, line 78: On Linux, you can technically have a backslash inside the extension. Just not forward slash.
No sane file would have it in either case, but I'll just leave a comment here in the code.

packages/paths/src/types.ts
```diff
@@ -0,0 +1,543 @@
...
+ * - Rejects parent directory references (..)
+ * - Rejects absolute paths and drive letters
+ */
+export const RelativePathSchema = z
+  .string()
+  .transform((s) => s.replaceAll("\\", "/")) // Normalize to forward slashes
+  .transform((s) => s.replace(/^\/+/, "")) // Remove leading slashes
+  .transform((s) => s.replace(/\/+$/, "")) // Remove trailing slashes
+  .transform((s) => s.replace(/\/+/g, "/")) // Collapse multiple slashes
+  .refine((s) => !s.startsWith(".."), {
```
Sewer56: This rejects paths like `../path`, but accepts paths like `./../path` or even `folder/../../path`.
I will need to fix, as existing docs and function says 'rejects parent directory references', that presumably means at any level.

Sewer56: b67e8b15913e4e29075611b4ba8f78da251671f4

packages/paths/src/pathUtils.ts
```diff
@@ -0,0 +1,469 @@
...
+
+// ============================================================================
+// Platform detection helper
+// ============================================================================
+
+/**
+ * Detect whether a path is Windows-style (drive letter) or POSIX-style
+ * and return the appropriate path module.
+ */
+export function detectPathModule(p: string): PathModule {
```
Sewer56, line 539: There's ways in which a UNC path could be passed into this function.

While we shouldn't deliberately use UNC paths ourselves; it's not improbable that external code may hand us one. After all, Vortex had to write some logic for UNC handling prior.

Made a small tweak to add UNC parsing here.

Sewer56: efab45316145460945fa007380ce3c050dc9c713

packages/paths/src/FilePath.ts
```diff
@@ -0,0 +1,380 @@
...
+    const baseWithSep = baseNorm.endsWith(fs.sep)
+      ? baseNorm
+      : baseNorm + fs.sep;
+
+    if (!childNorm.startsWith(baseWithSep) && childNorm !== baseNorm) {
+      return null; // This path is not under basePath
+    }
+
+    // Extract relative path
+    const relative = pathMod.relative(resolvedBase, resolvedChild);
```
Sewer56, line 320: This call could misbehave if `basePath` ends with `/` or `\`.

Normally this wouldn't be a problem, as you *normally* wouldn't create paths that have a trailing separator from within the library... except...

Roots, `C:\` and `/` are valid cases where you would have a trailing slash.

Sewer56: I added tests that validate the `C:\` and `/` case.
I'll treat the remaining case as 'not possible', as our code shouldn't produce it.

Sewer56: Actually, hm. `win32Split()` handles the trailing slash, but our `posixRelative` does not; so the code's inconsistent in the case we did somehow pass in a trailing slash. I'll fix it just in case. 👌

Sewer56: Resolved in b4a5e0af858f4c01fc1b2c291d46a955253d28e8 and 63bcd3c2202586e12bfce3baf9616f368ad44f84

packages/paths/src/pathUtils.ts
```diff
@@ -0,0 +1,483 @@
...
+  }
+
+  // Remove trailing slashes from result
+  while (i > 1 && p.charCodeAt(i - 1) === 0x2f) i--;
+
+  return p.slice(0, i);
+}
+
+function posixBasename(p: string, ext?: string): string {
+  if (p.length === 0) return "";
```
Sewer56, line 113: There's a bug here. `/` would have the basename resolve to `/`.
Likewise in the Windows one, `C:\` would resolve to `C:`

Sewer56: Resolved in ffb53578780f347ce17c9dfd8b1a9f800cf430b1

packages/paths/src/integration.test.ts
```diff
@@ -0,0 +1,239 @@
...
+  describe("cross-platform path handling", () => {
+    test("RelativePath normalizes separators", () => {
+      const windowsStyle = RelativePath.make("mods\\skyrim\\data");
+      const unixStyle = RelativePath.make("mods/skyrim/data");
+
+      expect(windowsStyle).toBe(unixStyle);
+      expect(windowsStyle).toBe("mods/skyrim/data");
+    });
+
+    test("ResolvedPath preserves OS separators", () => {
```
Sewer56, line 155: Added assert here on full path instead of `toContain()`. The substring doesn't assert edge cases, e.g. folder order, separators, trailing separators, etc.

Sewer56: f459e182b7123767d226f2cd8b39671814c48bb9

packages/paths/src/test-helpers/MockFilesystem.ts
```diff
@@ -0,0 +1,469 @@
+/**
+ * MockFilesystem - In-memory filesystem for testing
+ *
+ * Provides a complete filesystem implementation in memory with configurable
+ * platform behavior (case-sensitive, path separators, etc.).
+ */
+
+// Note(sewer): Few small bugs in this file. But not blocking for current tests.
```
Sewer56, line 8: A few subtle bugs here. I left a note. 72a93b4a88d47d5b772b3b4e5544c81a88fd95b2 on the important ones. I didn't patch them, as they're not blocking. An LLM will most likely read my comment and patch it as needed in the future, should a test that runs into one of the edge cases be encountered.

packages/paths/src/resolvers/BaseResolver.ts
```diff
@@ -0,0 +1,391 @@
...
+
+  /**
+   * Check if path is under base path
+   */
+  private isUnder(
+    childPath: string,
+    basePath: string,
+    preNormalized = false,
+  ): boolean {
+    const fs = this.getFilesystem();
```
Sewer56, line 351: Similar as earlier comment regarding valid paths ending in slashes, i.e. `/` and `C:\` breaking. Will fix here too.

Sewer56: Sorted in fbf9acbfaf2849e4ba7d49bda5f19731ad2b5300

packages/paths/src/resolvers/BaseResolver.ts
```diff
@@ -0,0 +1,403 @@
...
+ * @template ValidAnchors - Union of string literals for valid anchor names
+ */
+export abstract class BaseResolver<
+  ValidAnchors extends string = string,
+> implements IResolver<ValidAnchors> {
+  #filesystem?: IFilesystem;
+
+  constructor(
+    public readonly name: string,
+    public readonly parent?: IResolverBase,
```
Sewer56, line 52: Per erri120's refactor guide. (Which should be codified as text somewhere-)
We should opt into JavaScript's native private fields, so I updated all our private use in 16ebebf8358297467b85db1de91d42c2ac307c84

Sewer56:
Okay. This was a massive one. Finally done. whew...
All downhill tomorrow- as far as SDV integration goes.

Just the one random CI failure that occasionally pops remains- I think it's cause of mixed node versions; instead of inheriting from workspace. So I just fixed that.

Doing 3 runs to verify... then I merge and sleep.

PR #21854, opened 2026-03-25
Fixed: make renderer Jest tests self-declare their runner

Sewer56:
Renderer Jest tests were relying on ambient globals (`describe`, `it`, `expect`, etc.), which caused TypeScript/LSP errors in editors for files outside the main `tsconfig.json` include scope.

This switches those tests to explicitly import from `@jest/globals`, fixing LSP errors for files that weren't cleanly moved over to Vitest.

Changes:
- Added `@jest/globals` dependency to the renderer workspace
- Updated 4 renderer Jest test files with explicit imports
- Replaced bare `jest` global access in `NotificationAggregator.ts` with a `globalThis`-based check
- Ran `oxfmt` on all changed files
- Validated: focused Jest run passes (4 suites, 107 tests), renderer typecheck passes

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -39,6 +39,9 @@ catalogs:
     '@iarna/toml':
       specifier: ^2.2.5
       version: 2.2.5
+    '@jest/globals':
+      specifier: ^29.7.0
+      version: 29.7.0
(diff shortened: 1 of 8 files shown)


erri120:
Obsolete after #22081.

Issue #21861, opened 2026-03-25
"Unresolved file conflicts" popup missing mods & unsaved rules auto-clearing


Healz4Life:
### Healz4Life

**Describe the bug**
Two issues for managing file empty/conflicts in the "Unresolved file conflicts" summary popup:

Recently started failing to list all mods. I no longer can set empty rules in one place, which was the most useful way to set rules, even if it is from a somewhat strange pop-up.
It allowed me to set any rules I need to after, as I prefer using only set after for easier rule management, helping figure out mod cycles much easier.

When setting rules "before/after" in an empty conflict, the pending, unsaved rules will automatically clear and reset if a background mod finishes installing, or if a locked rule is removed.

**To Reproduce**
**Incomplete List**
1. Install and enable 2 mods with overlapping file conflicts (e.g., Modern Minimap alongside Companion Map Icons).
2. Open the "There are unresolved file conflicts" notification.
3. With an unset rule, you will see only one mod is listed on the left rather than both as it had previously.

**Auto-clearing rules**
1. Open the "There are unresolved file conflicts" notification.
2. Set a few rule dropdowns to "before" or "after" without saving.
3. Allow a separate background mod installation to complete, OR remove a locked rule.
4. The unsaved rule selections auto-clear, interrupting the workflow.

**Expected behavior**
The unresolved conflicts popup should list all conflicting mods involved so rules can be set efficiently from one screen.

Unsaved rule selections should persist and not auto-clear when a background installation finishes or a locked rule is removed. Vortex is strange for not showing any sign of stuff like this happening.

**Screenshots**

<img width="733" height="531" alt="Image" src="https://github.com/user-attachments/assets/faae2b98-b0f6-4c0c-be9c-8a75bb3b9425" />

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/cdcdf2cf-602e-428b-9650-bb3018d67198" />

<img width="857" height="474" alt="Image" src="https://github.com/user-attachments/assets/2c5e99b5-ca74-4c0e-ac86-21fcf16a6203" />

**Platform (please complete the following information):**
 - OS: Windows 11
 - Vortex Version: 1.16.7

**Additional context**
The main unresolved conflicts button has historically been the most useful way to sort rules, even if it is a bit of a strange pop-up. It allowed me to efficiently set any rules I needed with my preference of only using "set after" for easier rule management and avoiding mod cycles. Missing mods in this list breaks that UX flow.

[...]

PR #21871, opened 2026-03-25
Unresolved conflicts dialog redundant conflict entry de-duplication and better persistence for local changes

IDCs:
Also made sure we avoid messing up the component's local rules state when the Redux state updates.

fixes https://linear.app/nexus-mods/issue/APP-145/unresolved-conflicts-popup-omits-mods-and-clears-unsaved-rules
fixes https://github.com/Nexus-Mods/Vortex/issues/21861


Healz4Life:
💚

PR #21888, opened 2026-03-26
fixed data not persisting correctly

IDCs:
level_pivot's EmitRowCount unconditionally emits a single-row chunk (the "rows affected" count), which DuckDB's pipeline
surfaces as the query result — so UPDATE … RETURNING always reports 1 row regardless of whether any row was actually matched.

This caused setItem to skip the INSERT fallback on every write, silently discarding all persisted state.

Replace the UPDATE … RETURNING existence check with a SELECT, then UPDATE or INSERT accordingly.

DELETE ... INSERT would also work here.


IDCs:
@halgari sorry for the tag - wanted to make sure you're aware of this PR

halgari:
Lgtm, and long term I'll find a better way to do these updates

PR #21904, opened 2026-03-26
FOMOD is now sourced from NPM

Aragas:
Closes APP-14
Added @nexusmods prefix to FOMOD packages
Removed build:fomod step
Removed FOMOD submodule
Launcher was broken, fixed it additionally

### DIFF
--- a/src/renderer/src/extensions/news_dashlet/rss.ts
+++ b/src/renderer/src/extensions/news_dashlet/rss.ts
@@ -4,8 +4,8 @@ import { get } from "https";
 import * as url from "url";
 
 export interface IEnclosure {
-  length: string;
-  type: string;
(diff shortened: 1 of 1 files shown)


pnpm-workspace.yaml
```diff
@@ -21,7 +21,6 @@ allowBuilds:
   electron: true
   esbuild: true
   esptk: true
-  fomod-installer-native: true
```
erri120, line 24: `@nexusmods/fomod-installer-native` needs to be approved instead.

Aragas: It doesn't do actually anything atm, install does echo, so I decided to remove it. Should it still be kept?

PR #22007, opened 2026-03-30
Tools Page

Aragas:
Closes APP-65
Moved some type declarations to renderer
Also fixes some issues we have in general

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -133,6 +133,12 @@ catalogs:
     '@types/encoding-down':
       specifier: ~5.0.5
       version: 5.0.5
+    '@types/enzyme':
+      specifier: ~3.10.19
+      version: 3.10.19
+    '@types/enzyme-adapter-react-16':
+      specifier: ~1.0.9
+      version: 1.0.9
     '@types/feedparser':
       specifier: ~2.2.8
       version: 2.2.8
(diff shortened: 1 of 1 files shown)


src/main/package.json
erri120, line 1: Looks like a broken merge. These dependencies were removed in #20402.

Aragas: Renderer still had leftovers. When I was doing pnpm package:nosign it wouldn't have worked otherwise. Thanks for noticing!

Aragas:
ESLint had an issue, will have to change the logic after it's working now

erri120:
> ESLint had an issue, will have to change the logic after it's working now

I saw you added `tsconfigRootDir: import.meta.dirname` to the configs. What issue did you ran into?

Aragas:
> > ESLint had an issue, will have to change the logic after it's working now
> 
> I saw you added `tsconfigRootDir: import.meta.dirname` to the configs. What issue did you ran into?

ESLint's VSCode extension wasn't able to understand where to look at the tsconfig
```
Parsing error: No tsconfigRootDir was set, and multiple candidate TSConfigRootDirs are present:
 - D:\Git\NexusMods\Vortex\src\renderer
 - D:\Git\NexusMods\Vortex\src\main
You'll need to explicitly set tsconfigRootDir in your parser options.
See: https://typescript-eslint.io/packages/parser/#tsconfigrootdireslint
```

PR #22011, opened 2026-03-30
Move game store discovery into main-process scanners

halgari:
## Summary
This PR replaces the old renderer-side game store extension model with a main-process discovery system backed by DuckDB query data.

The new flow is:
- main process initializes store scanners for Steam, GOG, Epic, Xbox, Origin, Uplay, and registry
- scanner results are written into a `store_games` DuckDB table
- the generic query IPC layer exposes `all_store_games` to the renderer
- renderer-side discovery and launch helpers consume query-backed store data instead of store-specific extension objects
- explicit dirty-table invalidation keeps query consumers up to date after discovery writes

### 1. Main-process discovery and query plumbing
- Adds `DiscoveryCoordinator`, `IStoreScanner`, and store-specific scanners under `src/main/src/games/scanners`
- Adds command IPC support for `discovery.start`
- Adds generic query IPC (`query:execute` and `query:dirty`) and wires discovery init into main persistence startup
- Adds `store_games` setup/select SQL and generated query types
- Stores discovery results in DuckDB and invalidates `store_games` after each successful run

### 2. CLI support
- Adds a `cli:list-games` entry point and bundle in `@vortex/main`
- The CLI runs discovery, queries `all_store_games`, and prints store, id, name, and install path
- Uses a CLI-safe logging shim so the bundle does not depend on Electron logging setup

### 3. Renderer migration to query-backed stores
- Adds a small query client for renderer-side cached query execution with dirty-query invalidation
- Reworks `GameStoreHelper` to load store data from `all_store_games` rather than calling legacy store implementations
- Reworks `StarterInfo` and game mode discovery to use the new query-backed helper methods
- `startQuickDiscovery()` now triggers a main-process scan and then consumes the query results

### 4. Standardized game finders and store-specific cleanup
- Removes bundled store-specific extensions for GOG, Origin, Uplay, and Xbox
- Updates game extensions to use standardized `queryArgs` / `GameStoreHelper.findBy*()` lookups instead of store-specific utility modules and direct launcher objects
- This simplifies the game finder path and standardizes store detection across extensions instead of each store shipping its own independent lookup/cache logic
- Also removes renderer API exports and hooks that only existed to support the old store-extension model

## Breaking Changes
This branch changes extension-facing behavior in ways that can break extensions which rely on the old store registration model.

Notable breakages:
- `registerGameStore` is removed from `IExtensionContext`
- bundled `extensions/gamestore-*` packages are removed for GOG, Origin, Uplay, and Xbox
- renderer-side code should no longer assume store-specific launcher objects exist or can be fetched/registered dynamically
- extensions that imported or depended on old store-specific utilities/exports need to migrate to the standardized query-backed `GameStoreHelper` and `queryArgs` flow

In practice, any out-of-tree/community extension that still expects to register a custom game store or talk to the removed bundled store extensions will need an update.


insomnious:
Do we have any data on how many extensions (if any) are affected? Maybe @erri120 extension's analysis?

halgari:
Would be good to find out. All the ones in our codebase are very old, and the small changes I included fixed those. Any extension that uses "query args" is fine and I think most do.

halgari:
I considered updating the extensions that used the `util.*` endpoints to use queryArgs, we can still do that, but it seems like a bit of scope creep. I could be convinced otherwise however.

erri120:
> Do we have any data on how many extensions (if any) are affected? Maybe @erri120 extension's analysis?

Based on all extensions (data snapshot from last month):

- `registerGameStore` got removed and is not called by any extension,
- `util.steam` got removed but is used 33 times by extensions, see [export.csv](https://github.com/user-attachments/files/26370722/export.csv) for usage,
- `util.epicGamesLauncher` got removed but is used 7 times by extensions, see [export.csv](https://github.com/user-attachments/files/26371164/export.csv) for usage,
- `util.GameNotFound` got removed but is used 3 times by extensions, see [export.csv](https://github.com/user-attachments/files/26370663/export.csv) for usage.

Extensions are using `util.steam.findByName` and `util.steam.findByAppId`, to keep compatibility I recommend having a `util.steam` stub that redirects `findByName` and `findByAppId` to an actual implementation. Annotate it with `@deprecated` and make a note in the API changelog that this will be removed soon ™️. We can also let the extension authors know directly about this change. Same with `util.epicGamesLauncher` which has `util.epicGamesLauncher.findByAppId`, `util.epicGamesLauncher.findByName`, and `util.epicGamesLauncher.isGameInstalled` usages in extensions.

Queries used for verifying API changes:

```sql
-- checks for usage of 'registerGameStore'
SELECT
  extensionId,
  json_extract_string(calls.value, '$.method') AS method
FROM extensions,
  json_each(value, '$.analysis') AS file,
  json_each(json_extract(file.value, '$.apiUsage.calls')) AS calls
WHERE method = 'registerGameStore'
```

```sql
-- checks for explicit steam usage
SELECT
  extensionId,
  file.key AS file,
  json_extract_string(component.value, '$.component') AS component
FROM
  extensions,
  json_each (extensions.value, '$.analysis') AS file,
  json_each (json_extract(file.value, '$.vortexApiUsage.components')) AS component
WHERE
  extensionId NOT LIKE 'github-Nexus-Mods-vortex-games%' AND
  json_extract_string(component.value, '$.component') ILIKE '%steam%'
ORDER BY
  component ASC,
  extensionId ASC,
  file ASC;
```

src/shared/src/types/preload.ts
```diff
@@ -409,3 +415,30 @@ export interface TelemetryApi {
...
+}
+
+/** Store game entry as returned from main-process discovery */
+export interface IStoreGameRow {
+  store_type: string;
+  store_id: string;
+  install_path: string;
+  name: string | null;
+  store_metadata: string | null;
+}
```
erri120: Is this misplaced? Doesn't seem to be used as a preload type.

src/renderer/src/extensions/gamemode_management/util/discovery.ts
```diff
@@ -159,11 +159,62 @@ function updateManuallyConfigured(
...
+  storeGames?: IStoreGameRow[],
 ): Bluebird<IGameStoreEntry> {
-  return GameStoreHelper.find(game.queryArgs)
+  const findResults =
+    storeGames !== undefined
+      ? Bluebird.all([
+          Bluebird.resolve(matchStoreGames(game.queryArgs, storeGames)),
+          game.queryArgs.registry !== undefined
+            ? GameStoreHelper.find({ registry: game.queryArgs.registry })
+            : Bluebird.resolve([]),
```
erri120: `matchStoreGames` returns `IGameStoreEntry[]` the method isn't async so we don't need Bluebird here.

src/renderer/src/extensions/gamemode_management/GameModeManager.test.ts
erri120, line 1: New tests should use vitest.

src/renderer/src/extensions/gamemode_management/util/discoveryQueries.test.ts
erri120, line 1: New tests should use vitest.

src/renderer/src/util/queryClient.ts
```diff
@@ -0,0 +1,137 @@
...
+  inFlightPromise?: Promise<unknown>;
+  queryName: string;
+  stale: boolean;
+}
+
+interface EnsureQueryOptions {
+  force?: boolean;
+}
+
+function toQueryError(err: unknown): Error {
```
erri120: Should have custom `QueryError` type at this point.

src/renderer/src/util/GameStoreHelper.ts
```diff
@@ -435,106 +615,107 @@ class GameStoreHelper {
...
-            }
+  private rowToEntry(row: IStoreGameRow): IGameStoreEntry {
+    const metadata = this.parseStoreMetadata(row);
+    const entry = {
+      appid: row.store_id,
+      gamePath: row.install_path,
+      name: row.name ?? "",
+      gameStoreId: row.store_type,
+      priority: this.storePriority(row.store_type),
+    } as IGameStoreEntry & Record<string, unknown>;
```
erri120: All keys are known since they come from `IStoreGameMetadata`.

```suggestion
    const entry: IGameStoreEntry & Partial<Record<keyof IStoreGameMetadata, unknown>> = {
      appid: row.store_id,
      gamePath: row.install_path,
      name: row.name ?? "",
      gameStoreId: row.store_type,
      priority: this.storePriority(row.store_type),
    };
```

src/renderer/src/util/GameStoreHelper.ts
```diff
@@ -402,26 +403,205 @@ class GameStoreHelper {
...
-      return [];
-    }
+    })();
+  }
+
+  private launchXboxGame(api: IExtensionApi, appInfo: any): Bluebird<void> {
+    return toBlue(async (): Promise<void> => {
+      if (!appInfo) {
+        throw new ProcessCanceled("appInfo is undefined/null");
+      }
```
erri120: No bluebird.

src/renderer/src/util/GameStoreHelper.ts
```diff
@@ -402,26 +403,205 @@ class GameStoreHelper {
...
+  private getOriginURI(appId: string): string {
+    return `origin2://game/launch?offerIds=${appId}`;
+  }
+
+  private getUplayURI(appId: string): string {
+    return `uplay://launch/${appId}/0`;
+  }
+
+  private launchGOGGame(api: IExtensionApi, appInfo: any): Bluebird<void> {
+    return toBlue(async (): Promise<void> => {
```
erri120: No bluebird.

src/renderer/src/util/GameStoreHelper.ts
erri120, line 1: Lots of bluebird references that can be removed. Public facing API should return `PromiseLike<T>` for compatibility and implementations can use native `Promise<T>`.

src/main/src/games/scanners/SteamScanner.ts
```diff
@@ -0,0 +1,357 @@
...
+
+function isValidSteamPath(steamPath: string): boolean {
+  const libraryFoldersPath = path.join(
+    steamPath,
+    "config",
+    "libraryfolders.vdf",
+  );
+  try {
+    // eslint-disable-next-line no-restricted-imports
+    require("node:fs").statSync(libraryFoldersPath);
```
erri120: No reason to use `require` here, can just do `import { statSync } from"node:fs";` at the top if you need it...

erri120 (dismissed):
Important points:

1) API changes break extensions, instead use stubs that redirect and mark them as deprecated.
2) No new CLI. Our build process is still a mess and I really don't want to complicate it even more with a second entry point. This PR is already large enough, we don't need a CLI in here.

Code quality:

- no new Jest tests, new tests should use Vitest,
- no Bluebird resolve code, use `PromiseLike<T>` for API and native `Promise<T>` for implementations

src/shared/src/types/preload.ts
halgari: yes, I can remove it now, we used to send game store results across the IPC, but now it's more "find" based instead of querying for all games in one call

Aragas:
I absolutely agree with that we need to have stubs, but I would expand it to have stubs for any API that _was_ exposed by us and got removed. I believe just redirecting everything should not be an issue.
Ideally we need to also add at the same time a better API that a renderer extension could use to get the same data in a proper way. If it's already exposed by DuckDB alone, I guess just having the ability to query data is enough then

Just deprecate the stubs and show the proper way to do it in the future

halgari:
game adaptors remove much of the need for this code, closing for now, will come back to it later if needed

PR #22092, opened 2026-04-01
Add @vortex/game-extension-helpers package

Sewer56:
## Summary

Adds `@vortex/game-extension-helpers`, a new workspace package for shared code
that helps modernize game extensions which have to interop with the legacy
extension system we can't change. This achieves two things:

- Bridges existing `vortex-api` with the new `@vortex/paths` library
- Puts directory/file detection logic that's copied and pasted across extensions
  into one common location.

Starts with paths helpers; can grow with more shared helpers over time.

`vortex-api` wasn't the place for it, as that's technically the 'contract', so a new package was made.

## Paths

`vortex-api` installer callbacks hand extensions mixed lists of file and
directory paths with inconsistent separators. This is a legacy design we cannot
change. Rather than each extension normalising and separating these paths
itself, this package does it once through `@vortex/paths`.

```typescript
import { splitPathsByKind } from "@vortex/game-extension-helpers/paths";

const result = splitPathsByKind([
    "Pack\\Content\\",                       // directory
    "Pack\\Content\\Data.xnb",               // file
    "Pack/Mods/",                            // directory
    "/Pack/Mods/Helper/manifest.json",       // file
]);
// result.files: ["Pack/Content/Data.xnb", "Pack/Mods/Helper/manifest.json"]
// result.directories: ["Pack/Content", "Pack/Mods"]
```

### Public API

- **`PathGroups`** - `{ files: RelativePath[], directories: RelativePath[] }`
- **`isDirectoryPath(path)`** - detects directory markers (trailing `/` or `\\`)
- **`splitPathsByKind(paths)`** - separates files from directories
- **`toRelativePaths(files)`** - normalises file paths to `RelativePath[]`

## Testing

19 tests covering trailing slashes, backslashes, invalid paths, and ordering
preservation.

---

Fixes [APP-221](https://linear.app/nexus-mods/issue/APP-221/add-shared-game-extension-helper-package)


erri120 (changes requested):
I'm strongly against making more packages in general but especially packages like this that have almost no code in them. If the paths library is lacking in APIs, then you can add more to it, otherwise have the code in the extension itself. We don't know what to abstract yet until we have three or more usages of the same thing.

Sewer56:
@erri120 I've considered 3 places, prior to this.

- Paths Library
- `vortex-api`
- A New Package

My general thoughts were...

## Paths Library

Doesn't seem to be the best fit.

The paths library is general purpose; while this deals with data specific to legacy Vortex code.

Behaviour such as `split an array of string[]` into 2 arrays of `RelativePath[]` based on whether there is a trailing separator is very domain specific.

As such, it, at best belongs in a `util` namespace if placed in paths lib. At which point your review comment would be `don't create generic namespaces such as 'util'`.

## vortex-api

This is the contract between extensions and Vortex itself.
Shouldn't contain hard code; just what's needed for the two to communicate.

So it doesn't fit here.

## A New Package

With nowhere else left to go, an alternative is making a new package.
I figured it's best to make a package that's a shim between `vortex-api` and our more modern tech.

We can't break the old contract, but we can provide with something that helps work with it.

Sure, I could make a `vortex-api-wrapper`, type of thing, so make a nicer API ontop of the older API, but that's way out of scope.

The risk is making a package that's too barebones.
While, admittedly, it is barebones right now; I'm happy to see what else we'd need down the road.

------------------

# Additional considerations

A lot of game extensions live outside of this repo.
Vortex has 500+ game extensions, not even 20% of those live in this repository.
Some popular ones aren't here and are community maintained.

So any code written here should *eventually* be accessible externally too (paths lib included). For the sake of reliable xplat support.

A very quick rough 5 min search which I did prior, revealed (at least) 17 game extensions that currently have their own code to demux folders from files. Usually by copy-pasting the regex `/\/$|\\$/` , `endsWith(path.sep)` or some variant thereof.

My point is simply to stop extension authors from copy-pasting common code like this.

erri120:
We're far away from changing how extension authors write code. While I appreciate thinking ahead, we're not at that phase yet. For now, if you need this code in an extension, put it there. Once we have a better idea of how extensions 2.0 are going to work and what kind of API we want, we'll update the API and make it easier to write extensions.

Sewer56:
I respectfully disagree.

We need backcompat for the foreseeable future. The many extensions people depend on daily aren't going to port themselves, and a lot of them aren't even actively maintained. This package provides a pathway to incrementally modernize within the current system; for extensions that need logic beyond 'Folder A goes to B' that could be solved by a data/manifest driven approach.

I'm don't want to have the same problem as ChemBoy's extensions have, i.e. same logic copy-pasted across many extensions and repos.

The point of this package is to stop repeating ourselves. Putting existing code on modern primitives like `@vortex/paths` makes it easier to port to a future system. Code already running on paths is code you don't have to rewrite (for the most part). That means:
- Less duplicated code across extensions
- Less code to port when we do get to extensions 2.0
- A natural starting point for any game that needs xplat

This is just the first common pattern I've found. The broader intent is to progressively lift out common copy-pasted patterns as I work on extensions, so what remains is closer to pure game-related logic - not boilerplate every extension reinvents.

This also feeds into planning a future API: if extensions are stripped down to only what they genuinely need, with common behaviours living in one place, it becomes easier to see what an extension API actually has to provide.

5 minutes of search found 17+ extensions doing their own variation of this demuxing. The rule of three is long met.

"Wait for the future API that is far away and I've no idea what it looks like" is not a reason to keep duplicating code today.
The point is to leave the codebase better than you found it, not worse.

Sewer56:
If what you want are for the existing extensions to have their code de-duped raise a ticket should this be merged. Otherwise that'll be me going outside of sprint/ticket related goals- which wouldn't be a good look on me.

I had to edit some parts of older Stardew Valley code last week to keep you happy- even though it was outside of the scope of what I was required to do (which was rearrange for maintainability).

I considered it less friction, than bickering; but this is a bit larger scope than that.

PR #22222, opened 2026-04-06
Create `@vortex/fs` package

erri120:
Combines the new approach found in VS Code with our existing ideas and requirements into a new package.

See [Concepts](https://github.com/Nexus-Mods/Vortex/tree/task/vortex-fs/packages/fs#concepts) documentation in the package README for details.

### DIFF
--- a/packages/fs/src/browser/filesystem.ts
+++ b/packages/fs/src/browser/filesystem.ts
@@ -69,7 +69,7 @@ export interface FileSystem {
    *
    * @throws {@link FileSystemError}
    * */
-  deleteRecursive(path: ResolvedPath): Promise<void>;
+  deleteRecursive(path: QualifiedPath): Promise<void>;
 
   /**
    * Returns the status of the entry.
@@ -101,7 +101,7 @@ export interface FileSystem {
       include?: Pattern;
       exclude?: Pattern;
     },
-  ): Promise<AsyncIterator<ResolvedPath>>;
+  ): Promise<AsyncIterator<QualifiedPath>>;
 
   /**
    * Returns an async iterator to enumerate the directory.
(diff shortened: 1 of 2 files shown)


packages/fs/src/node/filesystem.ts
```diff
@@ -0,0 +1,568 @@
...
+    const overwrite = options?.overwrite ?? false;
+    if (!overwrite) {
+      let targetExists = false;
+
+      try {
+        await stat(target);
+        targetExists = true;
+      } catch (err) {
+        const { code } = parseNodeError(err);
+        targetExists = code === "not found";
```
IDCs: Shouldn't this be inverted? `code !== "not found"`

packages/fs/src/browser/filesystem.ts
```diff
@@ -0,0 +1,360 @@
...
+   * @throws {@link FileSystemError}
+   * */
+  delete(path: QualifiedPath): Promise<void>;
+
+  /**
+   * Deletes a file or a directory recursively.
+   *
+   * @throws {@link FileSystemError}
+   * */
+  deleteRecursive(path: ResolvedPath): Promise<void>;
```
IDCs: Shouldn't this be using `QualifiedPath` as a parameter?

packages/fs/src/browser/filesystem.ts
```diff
@@ -0,0 +1,360 @@
...
+  enumerateDirectory(
+    path: QualifiedPath,
+    options?: {
+      includeStatus?: false;
+      types?: "all" | "files" | "directories";
+      recursive?: boolean;
+      include?: Pattern;
+      exclude?: Pattern;
+    },
+  ): Promise<AsyncIterator<ResolvedPath>>;
```
IDCs: I'm still getting used to the new paths library so this might not be valid.

But if we're enumerating a directory from an extension, shouldn't we be using `QualifiedPath` here as well to abstract away resolved paths?

erri120: Yeah, I just copy pasted the API and forgot to update the type.

PR #22245, opened 2026-04-07
Switched to the new upload API for Collections

Aragas:
Closes BLU-247
Closes LAZ-869

* Replaced the 3-step GraphQL collection upload flow (getRevisionUploadUrl → PUT → createCollection/createOrUpdateRevision) with the new 5-step V3 REST API that supports large file uploads via **multipart.**

image


Aragas:
Looks like the commit formatting code isn't working

src/renderer/src/extensions/nexus_integration/util_v3/manifestMapping.ts
```diff
@@ -0,0 +1,68 @@
...
+    author: mod.author ?? null,
+  };
+}
+
+function emptyToNull(value: string | undefined | null): string | null {
+  if (value === undefined || value === null || value === "") {
+    return null;
+  }
+  return value;
+}
```
erri120: Undefined, null and an empty string are falsey, can be simplified to single boolean operation.

src/renderer/src/extensions/nexus_integration/util_v3/uploadV3.ts
erri120, line 1: Two things:

1) Uploading, similar to downloading, should be in the main project and use `got` as well as the existing retry facilities.
2) Why are we polling for results?

erri120 (changes requested):

src/renderer/src/extensions/nexus_integration/util_v3/uploadV3.ts
Aragas: 1. That's a solid point, will be implemented
2. We need the part/parts upload to be finished before calling `client.createCollection`/`client.createCollectionRevision`, that's why we wait and see if any error popped up

src/shared/src/types/ipc.ts
```diff
@@ -250,6 +303,9 @@ export interface MainChannels extends MainCallbackChannels {
 
   // Feature flags: main pushes updated flags after each successful poll
   "flags:synchronize": (flags: FeatureFlag[]) => void;
+
+  // Uploads: main pushes byte progress for an in-flight upload (throttled)
+  "upload:progress": (progress: WireUploadProgress) => void;
```
erri120: Should be a pull instead of a push model, same as the downloader.

erri120 (changes requested):

erri120 (approved):

Issue #21864, opened 2026-03-25
"Re-apply installer choices/preset on manual re-install of a mod" does not work as expected


bl1te:
**Your Nexus Mods username**
enbl1tenment

**Describe the bug**
1.16.7 changelog contains this line: _Re-apply installer choices/preset on manual re-install of a mod_
But it does not work when re-installing a mod. Buuut.. it _kinda_ does work when you update a mod.

**To Reproduce**
Steps to reproduce the behavior - 1:
1. Download any mod with FOMOD installer.
2. Pick some options different from default ones.
3. After installing, try to re-install the mod (without re-downloading, straight within Vortex).
4. FOMOD options are reset to default ones, you have to pick them again.

Steps to reproduce the behavior - 2:
1. Download any **older version** of a mod with FOMOD installer.
2. Pick some options different from default ones.
3. Now download the **latest** version of the mod.
4. Switch the version in Vortex.
5. Vortex won't show any prompt, it just installs the newer version silently.

**Expected behavior - 1**
When re-installing a mod, Vortex will memorize FOMOD choices and re-apply them, while still showing all FOMOD windows for user to verify it.

**Expected behavior - 2**
When updating a mod, Vortex will memorize or reset FOMOD choices because there could be some changes to the mod. And show all FOMOD windows as well for user to verify it.

**Screenshots**
Gif demo: https://imgur.com/a/5sB7YRy
As you can see, after changing the version no FOMOD appeared.
And after re-installing the latest version, I had to pick options again. While contents of the installer are different from older version.

**Platform (please complete the following information):**
 - OS: Windows 11
 - Vortex Version: 1.16.7

bl1te:
@IDCs Hello, could you please clarify whether you will fix another part of this issue where upon re-installing a mod it doesn't show the FOMOD window but re-installs the mod silently? It is explained in "steps to reproduce - 1" and "expected behavior - 1".
This seems inconvenient since if one re-installs a mod, there **is** a reason to do so, such as changing the options or checking what were the choices last time.

Currently, in order to change the FOMOD options you have to perform it in **two** steps:
1. **Un**install the mod first.
2. Install it again as if it was just downloaded.
3. Resolve file conflicts again. And since we are talking about FOMOD things, it means that the mod might contain many files => many conflicts to resolve.

IDCs:
[...]

PR #22250, opened 2026-04-07
on manual fomod re-install, show fomod dialog with preselected values

IDCs:
Rather than keep flip flopping between fresh fomod dialog and re-install automatically with preset. We now show the dialog with pre-selected choices. (not applied when installing a collection)

Users that want to maintain the preset without changing anything can just keep clicking next -> next -> finish

Users that want to add/modify something can change the pre-selection as they choose.

This is the only middleground currently.

fixes nexus-mods/vortex#21864

### DIFF
--- a/src/renderer/src/extensions/installer_fomod_native/installer.ts
+++ b/src/renderer/src/extensions/installer_fomod_native/installer.ts
@@ -53,11 +53,22 @@ export const install = async (
     // thread — converting/dispatching large installSteps arrays — for no
     // visible benefit since the dialog is never shown.
     const isUnattended = unattended === true && fomodChoices != null;
+
+    // When attended (manual reinstall) with saved choices, don't pass them to
+    // the C# engine — it would auto-advance through matching steps without
+    // showing the dialog. Instead, pass them as "attended presets" so the
+    // DialogManager can pre-select options in the UI while still showing the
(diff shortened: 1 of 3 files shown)


Aragas:
It might be a better idea to actually pass a new flag to the FOMOD installer instead - use the passed choices as presets (headless mode) or as pre-filled ones (headfull mode). This way we would avoid unnecessary logic in the FOMOD integration.
Only question is, how hard would the C# code change be vs TS here?

IDCs:
I have small concerns about people spamming the "next" button without letting the state sync and was considering to place a debouncer, so pre-filled options within the C# code directly would probably be better. Other than that the TS implementation works quite well.

If you prefer to do it on the C# side of things, I can handover that specific ticket to you (you're more up to date with the fomod native code)

Aragas:
Done, you'll just need to pass true for the headfull mode with existing presets
https://github.com/Nexus-Mods/fomod-installer/pull/44

IDCs:
@Aragas - build is failing on that one.

Worth mentioning that this functionality may be backported to 1.16.x. Given that fomod is a NPM package now, I'm not sure if the C# code can be easily backported as well.

So I'm thinking this TS code will be applied to the v1.16 version.

Aragas (approved):

PR #22258, opened 2026-04-07
feat: add game adaptor service system with renderer bridge and mod installer

halgari:
## Summary

Adds a complete game adaptor system that allows game support to be defined declaratively via adaptor packages, loaded into isolated Worker threads, and bridged into Vortex's existing extension system (registerGame, registerModType, registerInstaller).

### Adaptor API (`@vortex/adaptor-api`)

- Branded types with runtime validation (URI, PID, SemVer, store IDs)
- Service contracts: `IGameInfoService`, `IGamePathService`, `IGameToolsService`, `IGameModTypesService`, `IFileSystemService`
- Builder helpers (`gameInfo()`, `gameTools()`, `gameModTypes()`) for ergonomic shorthand input
- `@provides` decorator and `ServiceRegistry` module augmentation for type-safe service wiring

### Adaptor Host (`src/main/node-adaptor-host/`)

- Worker-isolated adaptor loading with VM-sandboxed module evaluation
- Bidirectional RPC transport over MessagePort with correlation-based call/response
- Registry and name service for URI → PID routing
- Crash handling and graceful shutdown with 2s force-terminate fallback

### Renderer Bridge (`src/renderer/extensions/adaptor_bridge/`)

- Translates adaptor contracts into Vortex's imperative registration API
- Lazy resolution chain: game info (eager) → paths → tools + mod types (after discovery)
- Topological sort for mod type priority ordering from declarative runsAfter/runsBefore
- Stop pattern → regex conversion for FOMOD installer compatibility
- Adaptor-aware mod installer (priority 50) with stop pattern and file pattern detection

### IPC Bridge

- Typed IPC channels (`adaptors:list`, `adaptors:call`) with `Serializable` return types
- Preload API exposure for renderer-side adaptor queries

### Reference Implementation

- Cyberpunk 2077 adaptor demonstrating all contracts (info, paths, tools, mod types)
- Ping-test adaptor for integration testing

### DIFF
--- a/src/main/package.json
+++ b/src/main/package.json
@@ -12,7 +12,8 @@
     "typecheck": "pnpm tsc",
     "lint": "pnpm eslint --concurrency auto .",
     "lint:quiet": "pnpm eslint --concurrency auto --quiet .",
-    "test": "pnpm vitest run --config ./vitest.config.ts && pnpm vitest run --config ./vitest.integration.config.ts",
+    "test": "pnpm vitest run --config ./vitest.config.ts",
+    "test:integration": "pnpm vitest run --config ./vitest.integration.config.ts",
     "package": "node ./prepare-dist-package.mjs && pnpm install --dir=./dist && pnpm electron-builder --config ./electron-builder.config.json --config.win.forceCodeSigning=true --config.win.sign=./sign.cjs --publish never",
     "package:nosign": "node ./prepare-dist-package.mjs && pnpm install --dir=./dist && pnpm electron-builder --config ./electron-builder.config.json --publish never",
     "prepare:win": "curl -L \"https://aka.ms/vs/17/release/vc_redist.x64.exe\" -o \"./build/VC_redist.x64.exe\" --create-dirs && curl -L \"https://aka.ms/dotnet/9.0/windowsdesktop-runtime-win-x64.exe\" -o \"./build/windowsdesktop-runtime-win-x64.exe\" --create-dirs"
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -9,7 +9,6 @@ export default defineConfig({
   test: {
     projects: [
       "./src/**/vitest.config.ts",
-      "./src/main/vitest.integration.config.ts",
       "./packages/**/vitest.config.ts",
       "./extensions/**/vitest.config.ts",
       "./scripts/vitest.config.ts",


src/main/src/adaptors.ts
```diff
@@ -1,17 +1,106 @@
...
+} from "./node-adaptor-host/loader";
 
 // Infrastructure packages — not adaptors, don't try to load them
 const INFRA_PACKAGES = new Set(["adaptor-api"]);
 
+function createPathResolver(): PathResolver {
+  if (process.platform === "linux") {
+    return new LinuxPathProviderImpl();
+  }
+  // TODO: Add WindowsPathProviderImpl when available
```
IDCs: A `win32` path provider must be added before we can even consider merging this in.

Pretty sure this will crash on Windows on startup.

src/renderer/src/extensions/adaptor_bridge/index.ts
```diff
@@ -0,0 +1,688 @@
...
+  const modTypes = gameModTypeCache.get(gameId);
+  if (!modTypes) return null;
+
+  for (const [typeId, modType] of Object.entries(modTypes)) {
+    // Priority 1: Check stop patterns (directory markers in the archive)
+    for (const stopDir of modType.stopPatterns) {
+      const normalizedStop = stopDir.replace(/\\/g, "/");
+      for (const file of files) {
+        const normalizedFile = file.replace(/\\/g, "/");
+        const idx = normalizedFile.indexOf(normalizedStop);
```
IDCs: `indexOf` is a bit too eager and will give false positives.

Silly example:
`data/skse/plugins` would match `userdata/skse/plugins/foo.dll`

src/renderer/src/extensions/adaptor_bridge/index.ts
```diff
@@ -0,0 +1,688 @@
...
+  let cachedTools: GameToolsInfo | null = null;
+  let cachedModTypes: GameModTypesInfo | null = null;
+
+  /** Resolves game folder paths. Called once after discovery. */
+  async function getFolders(
+    store: string,
+    gamePath: string,
+  ): Promise<GameFolderMap> {
+    if (!cachedFolders && pathsUri) {
+      cachedFolders = (await callAdaptor(name, pathsUri, "resolveGameFolders", [
```
IDCs, line 218: We need to invalidate `cachedFolders` in case the user changes the game location and wants to manually change the directory through the app. Or if he decides to manage the same game with different game stores (yes it's apparently a thing).

1. Activate cp77 - folders cached, e.g. `D:\cp77`
2. Manually change the path to `E:\cp77`
3. `cachedFolders` will return a stale path.

src/main/src/adaptors.ts
IDCs: Yep

```
App threw an error during load
Error: No PathResolver available for platform: win32
    at createPathResolver (C:\Projects\vortex\src\main\out\main.cjs:3874:8)
    at createFileSystemHandler (C:\Projects\vortex\src\main\out\main.cjs:3877:19)
    at Object.<anonymous> (C:\Projects\vortex\src\main\out\main.cjs:3911:28)
    at Module._compile (node:internal/modules/cjs/loader:1714:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1848:10)
    at Module.load (node:internal/modules/cjs/loader:1448:32)
    at Module._load (node:internal/modules/cjs/loader:1270:12)
    at c._load (node:electron/js2c/node_init:2:17993)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:244:24)
```

packages/adaptor-api/src/contracts/filesystem.ts
erri120, line 1: After checking the service system, I don't think the filesystem API should be exposed as a proxy service. The service system with RPC works fine for simple methods like `stat` or `createDirectory` but as you noticed fails at complex method like `enumerateDirectory` or the `createStream` methods.

The `FileSystem` API was designed to be consumed directly by extensions/adaptors. My idea was that we create an implementation of the API that does RPC calls:

Currently:

```
Adaptor -- RPC via IFileSystemService  --> Main
```

Alternative:

```
Adaptor --> FileSystemAdaptorImpl -- RPC --> Main
```

The implementation for the adaptor can be shipped with the `adaptor-api` package and is then communicating back to the main process. The implementation is then responsible for all the cruft on top of RPC, like batching requests and stream and iterator management.

packages/adaptor-api/src/contracts/game-paths.ts
erri120, line 1: The goal of the FS abstraction with `QualifiedPath` and the `PathProvider` + `PathResolver` system was to allow paths to contain opaque data and have provider-specific "bases"/"anchors" and methods for creating paths. Here's how I had imagined that would be used for games:

```typescript
// Union type representing all supported stores
export type StorePathProviders = SteamPathProvider | GOGPathProvider /** others */;

// This base is always available for all store-specific path providers
export type GamePathBase = "game";

// Steam-specific base
export type SteamPathBase = GamePathBase | "workshop" | "cloud saves" | "compatdata";
export interface SteamPathProvider extends PathProvider<SteamPathBase> {
  readonly scheme: "steam":
  constructor(appId: number);
}

// Steam locator would create a per-game path provider
const pathProvider = new SteamPathProvider(12312);

// game code is invoked with the provider
function getGamePath(provider: StorePathProviders) {
  // steam-specific logic
  if (provider.scheme === "steam") {
    provider.fromBase("compatdata").join("foo bar baz");
  }

  // otherwise generic game folder
  return provider.fromBase("game");
}
```

Paths, especially in game code should always be created from either a path provider or by joining from on an existing path. Only that way can we guarantee that the qualified path can actually be resolved.

erri120 (changes requested):
See comments. Also we should rename `URI` to something else because it's not a [URI](https://datatracker.ietf.org/doc/html/rfc3986).

src/main/src/adaptors.ts
halgari: This part of the PR is scope creep, I've removed all the FS code from this and we'll handle that in a future pr

packages/adaptor-api/src/contracts/filesystem.ts
halgari: I'm not sure I agree, but let's discuss next week, in the mean time I've removed the FS code from this PR as it's scope creep that isn't needed for the tickets its targeting

src/renderer/src/extensions/adaptor_bridge/index.ts
halgari: I removed this code, it's going to come in a separate cleaner PR

packages/adaptor-api/src/contracts/game-paths.ts
halgari: Let's discuss this this week and figure a path forward, I'm moving these changes into a new PR (PR not submitted yet)

src/renderer/src/extensions/adaptor_bridge/index.ts
halgari: I think we should have people restart Vortex in those cases, the number of cached paths in Vortex is likely going to be substantial throughout the system. Even inflight messages and the like.

halgari:
I think they are valid URIs in that they contain a scheme and a path, and we use them like a URI to locate services.

PR #22282, opened 2026-04-08
feat(linux): Phase 1 Linux port — boot and run on Linux without breaking Windows

atabisz:
## Summary

This PR adds Phase 1 Linux support: `pnpm run start` works on Linux without crashing, while the existing Windows build is fully unaffected. All changes are additive and platform-guarded — no existing Windows code paths are altered.

Tested on Ubuntu 24.04 in a devcontainer and via local dev.

### Boot & runtime
- **XDG paths**: `localAppData` resolves to `~/.local/share/Vortex` on Linux (XDG spec)
- **IPC path**: cross-platform `getIPCPath()` utility; all pipe path sites updated
- **App startup**: fix crash when launched from VSCode or another Electron parent process
- **winapi-bindings shim**: full JS stub for all Windows API exports; aliased in webpack/rolldown so Linux builds never attempt to load the native `.node`
- **LD_LIBRARY_PATH**: prepended at startup so libloot shared libraries resolve correctly

### Steam / Proton game discovery
- Multi-root Steam path scanning (`findAllLinuxSteamPaths`)
- oslist-aware Proton prefix detection for installed and never-launched games
- `getMyGamesPath()` with `PROTON_USERNAME` fallback for Bethesda title save/INI paths
- Async `iniFiles()` / `mygamesPath()` / `iniPath()` with Linux Proton branch

### Mod deployment
- Case-insensitive path resolution (`resolvePathCase`) wired into `LinkingDeployment`, FOMOD extraction, and `fs.ts` (`copyAsync`, `renameAsync`, `ensureDirAsync`)
- Backslash normalisation after archive extraction
- Deployment manifest synthesis and backup-manifest fallback on ENOENT
- Wine-era manifest detection in `loadActivation`
- Hardlink staging: `lstat` enrichment on Linux; sync-callback + promise queue for scan

### NXM protocol registration
- Linux NXM handler registration via `.desktop` file + `xdg-mime`
- AppImage desktop entry creation and KDE `kbuildsycoca6` refresh
- Cold-start NXM URL buffering in `mPendingDownload`
- Second-instance lock fix; dev-mode download path fix

### Elevation (pkexec / polkit)
- `runElevated` Linux branch using `pkexec` with injectable spawner seam for tests
- polkit action file (`io.nexusmods.vortex.policy`) and udev rules (`10-vortex.rules`)
- `isSteamOS()` detection with `sudo -n` fallback
- `_setNotifier` callback + `rejectWithSteamOSNotification` helper

### FOMOD
- asarUnpack entries for Linux FOMOD binaries
- Case-insensitive source path resolution in `extractArchive`
- Backslash normalisation in copy source/destination paths

### gamebryo (Bethesda plugin management)
- `gamebryo-plugin-management`: Linux autosort, gameSupport guards, `PluginPersistor` cleanup
- `libloot_wstring_stub.so` for wstring ABI compatibility on Linux
- `gamebryo-savegame-management`: platform guard, GCC 13 `CHAR_WIDTH` patch
- `gamebryo-archive-check`: Linux guard

### Build & packaging
- Windows-only `.exe` extraResources moved to `win` block in electron-builder config
- `electron-builder.config.cjs` for platform-conditional `files` array (excludes `winapi-bindings` on Linux)
- `exe-version` patched to return `0.0.0` on Linux (pure-JS PE parser also added)
- `postinstall-libloot.cjs` script: builds and installs libloot, fixes RPATH, verifies addon

### Devcontainer
- Electron 39 runtime shared libraries
- Ubuntu 24.04 `libgtk-3-0t64` package name fix
- `--no-sandbox` flag for root container use

### Tests
- Unit tests: `proton.ts`, `steamPaths.ts`, `xdg.ts`, NXM common helpers, `resolvePathCase`, `winapi-shim`, `elevated.ts`, `ipc.ts`
- Windows CI guards: `path.join` in gameSupport assertions, mock fixes, skip guards for Linux-only shim tests

## Approach

- **Platform guards everywhere**: `process.platform === 'linux'` / `'win32'` checks isolate all new code paths; existing Windows behaviour is unchanged
- **No new runtime deps on Windows**: Linux-only native work (libloot, polkit) is isolated to Linux code paths and build scripts
- **Windows CI stays green**: existing test suite passes unmodified; new Linux tests are skipped on Windows

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -28,6 +28,9 @@ catalogs:
     '@duckdb/node-api':
       specifier: 1.5.1-r.1
       version: 1.5.1-r.1
+    '@electron/rebuild':
+      specifier: 4.0.3
+      version: 4.0.3
     '@electron/remote':
       specifier: ^2.0.12
       version: 2.1.3
@@ -493,12 +496,6 @@ catalogs:
     jsonwebtoken:
       specifier: ^9.0.0
       version: 9.0.3
-    leveldown:
-      specifier: 5.6.0
-      version: 5.6.0
-    levelup:
-      specifier: 4.4.0
-      version: 4.4.0
     limiter:
       specifier: ^3.0.0
       version: 3.0.0
@@ -773,6 +770,15 @@ overrides:
   node-addon-api: 8.5.0
 
 patchedDependencies:
+  exe-version@2.3.0:
+    hash: 91e00f57460903f38a25b5a4ac079b466d8ef8cb8e3352c9e611de0f2e2266b6
+    path: patches/exe-version@2.3.0.patch
+  gamebryo-savegame@2.1.2:
+    hash: 13dfacf7de67dcd9115b21f2b44af06545b1e211c07f00314e8ab23aab358d30
+    path: patches/gamebryo-savegame@2.1.2.patch
+  loot@6.2.1:
+    hash: 3d4c1edc7a442f9db9c091845b9f808f1d84517b03a9612fef4e41bd141f47bc
+    path: patches/loot@6.2.1.patch
   playwright-core@1.58.2:
     hash: d20016d2ebd264257c6e1885fb814dc7d150de5b41b3a46557b431ab8de2cf34
     path: patches/playwright-core@1.58.2.patch
@@ -781,6 +787,9 @@ importers:
 
   .:
     devDependencies:
+      '@electron/rebuild':
+        specifier: 'catalog:'
+        version: 4.0.3
       '@eslint-react/eslint-plugin':
         specifier: 'catalog:'
         version: 2.3.7(eslint@9.39.1)(typescript@5.9.3)
@@ -1284,7 +1293,7 @@ importers:
         version: 2.4.1
       exe-version:
         specifier: 'catalog:'
-        version: https://codeload.github.com/Nexus-Mods/node-exe-version/tar.gz/eded60fc0a0f3c234e1d586d2eb9952401945406
+        version: https://codeload.github.com/Nexus-Mods/node-exe-version/tar.gz/eded60fc0a0f3c234e1d586d2eb9952401945406(patch_hash=91e00f57460903f38a25b5a4ac079b466d8ef8cb8e3352c9e611de0f2e2266b6)
       i18next:
         specifier: 'catalog:'
         version: 19.9.2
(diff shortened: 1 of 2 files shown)


halgari:
Thank you for the work you've put into this. It’s clear a huge amount of effort went into getting Vortex to boot and run on Linux, and we do appreciate the time and thought behind it.

That said, we can’t accept this PR in its current form.

The main issue is scope. This PR is very large and bundles together a lot of different concerns into one change set, which makes it extremely difficult for us to review, validate, and safely merge. On a project of this size, we have to take an incremental approach so we can properly test each change, understand regressions, and keep the codebase maintainable.

A few examples of what makes this hard to take as a single PR:

* it spans a very large number of commits and touches many unrelated areas
* it includes application/runtime changes, packaging changes, Linux-specific integration, native/shim handling, and policy/elevation related changes all at once
* it commits binary artifacts directly into git, which is something we do not want to accept in PRs because it creates both security and maintenance risks. Binaries are difficult to audit, harder to review meaningfully, and create long-term overhead when they need to be updated, rebuilt, or verified

There is also another important issue here: several of the areas touched by this PR overlap with work that is already well underway through other efforts. Things like filesystem handling and bluebird removal are already being addressed through separate development tracks. Any changes in those areas would need to take that work into account and align with it, rather than moving in a parallel or conflicting direction. That makes it even more important that work be proposed in smaller, focused PRs, so it can be evaluated in the context of the roadmap and ongoing refactors.

To move this forward, we’d need this broken down into smaller, reviewable pieces. For example, separate PRs for things like:

* path and filesystem abstractions
* IPC path handling
* Linux-only startup/bootstrap fixes
* packaging/build-system changes
* Steam/Proton discovery
* elevation / policy integration
* extension-specific Linux support

In general, each PR should be narrowly scoped, easy to reason about, and avoid including generated files or committed binaries unless there is a very strong and explicitly agreed reason. For areas already being actively reworked, any proposed contribution also needs to fit with those efforts and not cut across them.

So while we appreciate the contribution, we’re going to close/reject this PR as submitted. Please don’t take that as dismissal of the work. If you want to continue contributing Linux support, we’d be very happy to look at smaller, incremental PRs that build toward the same goal in a way we can realistically review and maintain, and that take current development work into account.

atabisz:
@halgari Thanks for the feedback and encouragement, I'll split the work into smaller incremental PRs and hope to integrate it better with the wider project.

PR #22310, opened 2026-04-08
fix(linux): dev container runtime libs and startup fixes

atabisz:
## Summary

Fixes two issues that prevent Vortex from launching in Linux development environments. These are the first in a series of incremental Linux support PRs following feedback on the previous submission.

> **Note:** This PR is part of a broader Linux port effort being developed at https://github.com/atabisz/Vortex. The goal is to get Vortex launching on Linux without breaking the existing Windows build — platform-guarded additions, no large refactors. Individual fixes will be submitted as upstream PRs as they stabilize. Feedback welcome.

**1. Dev container: missing Electron 39 runtime libraries (`docker/linux/Dockerfile.devcontainer`)**

The existing devcontainer Dockerfile was missing 17 shared libraries required by Electron 39 at runtime. Without these, `pnpm run start` crashes immediately inside the container with missing-library errors. Added the full set of Ubuntu 24.04 package names for the Electron runtime dependencies.

**2. Linux startup fixes (`src/main/package.json`, `src/main/src/Application.ts`)**

- `ELECTRON_RUN_AS_NODE=` added to the dev start script. When Vortex is launched from a VS Code terminal on Linux, the shell inherits `ELECTRON_RUN_AS_NODE=1` from VS Code's own Electron process. This causes the Electron binary to behave as plain Node.js — `require('electron')` returns the npm package binary path string instead of the Electron API, crashing immediately. Clearing it to empty string (falsy) restores normal behaviour. No-op on Windows/macOS.

- Chromium sandbox disabled for root on Linux (`Application.ts` constructor). The sandbox requires unprivileged user namespaces which are unavailable for root users. Scoped to `process.getuid() === 0` so non-root Linux desktops are unaffected.

- `setupContextMenu()` guarded on Linux (`Application.ts`). `electron-context-menu` calls `require('electron')` at module load time. Under the same `ELECTRON_RUN_AS_NODE` conditions this resolves to the npm binary path string and crashes before the window opens.

All changes are backward-compatible: the `package.json` script change is a no-op on Windows/macOS, and both `Application.ts` guards are platform-gated.

## Test plan

- [ ] Windows CI passes (no regressions)
- [ ] `pnpm run start` inside the Linux dev container (VS Code attached via Dev Containers extension, running as root) launches without crashing
- [ ] `pnpm run start` from a VS Code integrated terminal on a native Linux desktop (VS Code running as an Electron app on the host) launches without crashing

### DIFF
--- a/docker/linux/Dockerfile.devcontainer
+++ b/docker/linux/Dockerfile.devcontainer
@@ -17,6 +17,24 @@ RUN apt-get update && apt-get install -y \
     python3-setuptools \
     build-essential \
     libicu-dev \
+    # Electron 39 runtime shared libraries (Ubuntu 24.04 package names)
+    libglib2.0-0 \
+    libnss3 \
+    libatk1.0-0 \
+    libatk-bridge2.0-0 \
+    libcups2 \
+    libdrm2 \
+    libxkbcommon0 \
+    libxcomposite1 \
+    libxdamage1 \
+    libxfixes3 \
+    libxrandr2 \
+    libgbm1 \
+    libasound2t64 \
+    libpango-1.0-0 \
+    libcairo2 \
+    libexpat1 \
+    libgtk-3-0t64 \
     && rm -rf /var/lib/apt/lists/*
 
 # Install .NET SDK 9.0 (required for fomod-installer)
--- a/src/main/package.json
+++ b/src/main/package.json
@@ -6,7 +6,7 @@
   "type": "commonjs",
   "main": "out/main.cjs",
   "scripts": {
-    "start": "pnpm cross-env NODE_ENV=development pnpm electron .",
+    "start": "pnpm cross-env ELECTRON_RUN_AS_NODE= NODE_ENV=development pnpm electron . --no-sandbox",
     "build": "pnpm cross-env NODE_ENV=development node ./build.mjs",
     "dist": "pnpm cross-env NODE_ENV=production node ./build.mjs",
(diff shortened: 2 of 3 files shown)


atabisz:
Hi @halgari , what's the best way for me to submit these pr's for linux support? Just keep putting small pr's and wait/be patient, or is there a better way I can coordinate with the nexus-mods team? Not looking to put any more work/pressure on you guys, so whatever advice you could give is welcome.

atabisz:
Hi @halgari — just a gentle ping on this one. No rush at all, I know the team is busy. I'm happy to rebase, split differently, or adjust the approach if that helps. The follow-up PRs (#22342, #22343, #22344) are also ready for review if you'd prefer to look at the series as a whole. Whatever works best for you — thanks for your time.

halgari:
Thank you for taking the previous feedback to heart and splitting the work into smaller, focused PRs, and for the effort you've put in. The core issue on our side is that we haven't yet done the systematic design work needed to properly support Linux in Vortex. That design work covers things like how packaging and CI should handle a second platform, how game-support extensions should reason about Proton versus native installs, how our path and filesystem abstractions should look, how elevation should work, and what long-term maintenance and QA commitment we're signing up for. Until we've worked through that, we're going to close this PR along with the others in the series. In the future, checking in with us on scope before committing time to a large body of work is the better path, so that effort isn't invested ahead of where the project is ready to take it on.

PR #22343, opened 2026-04-10
feat(linux): winapi-bindings Linux shim

atabisz:
## Summary

On Linux, `winapi-bindings` is a Windows-only native module that fails to load. This PR provides a pure-JS shim with full export coverage so the rest of the codebase can import it unconditionally.

- Add `winapi-shim.ts` with safe Linux implementations:
  - `GetDiskFreeSpaceEx` — uses `fs.statfsSync` for real disk stats
  - `GetVolumePathName` — walks `stat.dev` with `ENOENT` fallback
  - `GetNativeArch` — returns `process.arch`
  - `ShellExecuteEx` — throws with a descriptive Linux message
  - All Windows-only elevated/UI/registry/ACL/task-scheduler APIs — throwing stubs or no-ops
- Wire the shim into the webpack (renderer) and rolldown (main) build configs via a platform-conditional `resolve.alias` — shim is inlined at build time, no runtime require
- Remove dead `require('winapi-bindings')` from the Fallout 4 extension (variable was never read; causes `MODULE_NOT_FOUND` on Linux since bundled extensions use `copyfiles`, not webpack)
- Add `winapi-bindings` to `neverBuiltDependencies` on Linux and exclude it from the dist `package.json` to prevent `EEXIST` conflicts during `electron-builder` packaging
- 19 Vitest tests covering all shim behaviours

## Part of the Linux port series

This is one of several incremental PRs porting Vortex to Linux. The full fork is at https://github.com/atabisz/Vortex if upstream context is useful.

### DIFF
--- a/extensions/games/game-fallout4/src/index.js
+++ b/extensions/games/game-fallout4/src/index.js
@@ -1,9 +1,8 @@
 const Promise = require('bluebird');
 const path = require('path');
 const { util } = require('vortex-api');
-const winapi = require('winapi-bindings');
 
-/* 
+/*
 Ignore the Meshes\AnimTextData\AnimationOffsets\PersistantSubgraphInfoAndOffsetData.txt file as a conflict. 
 It's present in a lot of weapon mods but doesn't matter if it's overwritten. 
 This issue is compounded by users extracting all their BA2s. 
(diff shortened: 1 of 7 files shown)


halgari:
Closing this along with the rest of the Linux series — full explanation on #22310.

PR #22344, opened 2026-04-10
feat(linux): IPC pipe path abstraction

atabisz:
## Summary

Windows named pipes (`\\?\pipe\...`) are not valid on Linux. This PR introduces a `getIPCPath(id)` utility that returns the correct platform path and patches all IPC call sites to use it.

- Add `src/renderer/src/util/ipc.ts` with `getIPCPath(id)`:
  - Linux: `/tmp/vortex-{id}.sock` (Unix domain socket)
  - Windows: `\\?\pipe\{id}` (named pipe, unchanged)
- Patch four IPC `.listen()` call sites: `fs.ts`, `ExtensionManager.ts`, `symlink_activator_elevate/index.ts`, `elevated.ts`
- Add Linux IPC path guard and `binding.gyp` Linux linker flag patch to `patches/loot@6.2.1.patch` — without this, `node-loot` uses the Windows pipe path unconditionally on Linux (causes `EACCES` on startup), and the native addon build fails with an unresolvable `-l../loot_api/libloot` flag
- 3 Vitest tests (Linux socket path, Windows pipe path, slash-in-id normalisation)

## Part of the Linux port series

This is one of several incremental PRs porting Vortex to Linux. The full fork is at https://github.com/atabisz/Vortex if upstream context is useful.

### DIFF
--- a/patches/loot@6.2.1.patch
+++ b/patches/loot@6.2.1.patch
@@ -1,0 +1,83 @@
+diff --git a/binding.gyp b/binding.gyp
+index f73f16d..d8b1865 100644
+--- a/binding.gyp
++++ b/binding.gyp
+@@ -21,9 +21,6 @@
+                 "./loot_api/include",
+                 "<!(node -p \"require('node-addon-api').include_dir\")"
+             ],
+-            "libraries": [
+-                "-l../loot_api/libloot"
+-            ],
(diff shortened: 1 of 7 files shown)


halgari:
Closing this along with the rest of the Linux series — full explanation on #22310.

PR #22400, opened 2026-04-13
multiple Starfield enhancements and fixes

IDCs:
- Added support for blueprint plugins
- Updated native plugin list
- Updated to libloot 0.29.3
- Fixed plugin management switch mechanism requiring app restart when checking page condition() (switching between rules-based/drag-and-drop)

Note that the Starfield game extension changes are on a different PR:
https://github.com/Nexus-Mods/game-starfield/pull/57

fixes https://linear.app/nexus-mods/issue/APP-260/update-starfield-native-plugins-list-for-new-dlcs fixes https://linear.app/nexus-mods/issue/APP-261/load-order-menu-doesnt-update-until-restart fixes https://linear.app/nexus-mods/issue/APP-263/add-support-for-blueprint-plugins-in-starfields-load-order

### DIFF
--- a/src/renderer/src/views/components/Spine/SpineContext.tsx
+++ b/src/renderer/src/views/components/Spine/SpineContext.tsx
@@ -59,6 +59,19 @@ export const SpineProvider: FC = ({ children }: { children: ReactNode }) => {
   const activeProfileId = useSelector(activeProfileIdSelector);
   const activeGameId = useSelector(activeGameIdSelector);
 
+  // APP-261: page.visible() predicates owned by extensions read state slices
+  // the Spine does not natively track. Subscribe to this one so the per-game
+  // page memos recompute when the user flips plugin management on/off via
+  // the game-starfield "Load Order Management Method" setting (which always
+  // dispatches GAMEBRYO_SET_PLUGIN_MANAGEMENT_ENABLED alongside its own
+  // management-type action), otherwise the left menu stays stale until a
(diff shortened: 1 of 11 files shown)


src/renderer/src/views/components/Spine/SpineContext.tsx
```diff
@@ -59,6 +59,19 @@ export const SpineProvider: FC = ({ children }: { children: ReactNode }) => {
...
 
+  // APP-261: page.visible() predicates owned by extensions read state slices
+  // the Spine does not natively track. Subscribe to this one so the per-game
+  // page memos recompute when the user flips plugin management on/off via
+  // the game-starfield "Load Order Management Method" setting (which always
+  // dispatches GAMEBRYO_SET_PLUGIN_MANAGEMENT_ENABLED alongside its own
+  // management-type action), otherwise the left menu stays stale until a
+  // restart. Typed as `any` because this path lives in the gamebryo-plugin-
+  // management extension and isn't in core IState.
+  const pluginManagementEnabled = useSelector(
```
Aragas, line 70: Can't say I like this implementation detail,but it would require some kind of an abstraction otherwise, as I understand

Aragas (approved):

src/renderer/src/views/components/Spine/SpineContext.tsx
IDCs: The alternative would've been to replace `useMemo` with `useSelector` so that the condition/visible function gets called. Not worth the hassle.

PR #22409, opened 2026-04-13
Fixed collections_download_completed event not being emitted

Aragas:
Closes APP-277

### DIFF
--- a/extensions/collections/src/index.ts
+++ b/extensions/collections/src/index.ts
@@ -318,12 +318,29 @@ async function installCollection(
     .then((result) => {
       if (result.action === "Install") {
         const gameId = revision.collection.game.domainName;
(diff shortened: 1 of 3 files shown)


IDCs:
@Aragas this used to work in alpha.3 though 🤔 any idea what changed that could've induced it to break on alpha.4 ?

I'm just worried that we might be masking a different issue.

insomnious:
Tested this branch and the original problem was...

If started collection downloads from app, we don't have
- `collections_download_completed` (sent from vortex)
- `Mods: Download started events` (sent from api) - batched and not sent per event

If started collection downloads from site, all events are accounted for

This is now fine using this PR branch

IDCs (dismissed):

insomnious (approved):

PR #22418, opened 2026-04-14
Expose IFileSystem to adaptors via host filesystem service

halgari:
## Summary

- Introduces a `vortex:host/filesystem` service that adaptors can consume via `@vortex/fs`'s `IFileSystem` interface. The service runs in the main process and is exposed to adaptor workers over the existing RPC transport.
- Each worker gets an isolated filesystem session with its own cursor map, so enumeration state doesn't leak between adaptors. Cursors are cleaned up when the owning worker exits.
- Adds `QualifiedPath` serialization/rehydration at every service boundary (transport, IPC bridge, filesystem client) so paths survive cross-process round-trips intact.
- Implements `WindowsPathProviderImpl` with a `windows:///C/...` encoding scheme that normalizes drive letters and avoids backslash ambiguity in the qualified path format.
- Adds `NodeFileSystemImpl` (path-resolving layer), `NodeFileSystemBackendImpl` (raw Node `fs` operations), and `WebFileSystemImpl` (Node-to-Web stream adapter).
- Adds a browser-side `createFileSystemClient` that wraps RPC calls with proper error rehydration (`FileSystemError`, `Status`) and cursor-based enumeration batching.
- Ships an `fs-test` adaptor as a live conformance probe that exercises write/read, missing-file errors, and directory listing over the full RPC stack.
- Extends the renderer adaptor bridge to forward filesystem-backed service calls through the preload IPC layer.


src/main/src/filesystem/client.ts
erri120, line 1: Does this need to be in the fs package or can this be moved somewhere else?

packages/fs/src/browser/filesystem.ts
```diff
@@ -1,6 +1,16 @@
...
 
+/**
+ * Interface-only alias for {@link FileSystem}. The `I`-prefixed name
+ * matches the convention used by `@vortex/adaptor-api` service
+ * contracts, so a contract file can declare the host filesystem
+ * service with `ServiceRegistry["vortex:host/filesystem"]: IFileSystem`
+ * without introducing a parallel duplicate interface.
+ *
+ * @public */
+export type IFileSystem = FileSystem;
```
erri120, line 12: We should standardize our naming format. In some places we're using the `I`-prefix for types and interfaces and in others we're not.

src/main/src/filesystem/backend.ts
erri120, line 1: The implementation started out in the `fs` package but I moved it into the main project because I wanted the `fs` package to be as close to a type-only package as possible and because the only place where the implementation would be used directly is in the main project. The only real JavaScript objects that are actually in the compiled output are the error classes, the `QualifiedPath` class and path helper methods.

We should formalize how we want to structure our packages, especially the ones that end up being published to NPM and consumed by extensions/adaptors. I don't want this to turn into a similar situation as the current `vortex-api` package mess.

halgari:
Do we want to move PathResolverRegistryImpl as well? I'm thinking yes

src/main/src/filesystem/backend.ts
halgari: I've moved all of those out of the fs package, it should now be just types (and some helper functions) now

src/main/src/filesystem/client.ts
halgari: nope, moved out

packages/fs/src/browser/filesystem.ts
halgari: agreed, I've noted that down for future work

IDCs (approved):

PR #22440, opened 2026-04-14
Added Collection Tests

Aragas:
Tests against `extensions/collections` had to mock the entire `vortex-api`
barrel because importing any collections module pulled in electron, native
modules, React UI, and the full Redux store at module load. PR https://github.com/Nexus-Mods/Vortex/pull/22440
worked around this with a 231-line mock that partially reimplemented the
helpers it relied on.

This change splits the source so unit tests can call real implementations
instead. Pure logic stays in files named for what's in them; impure code
moves to files named for what it does. The remaining mock is a thin
re-export of real renderer code with no behavioural stubs.

Renderer:
- Extract `findModByRef` from dependencies.ts into its own file
- Extract `coerceToSemver`/`safeCoerce` and `isFuzzyVersion` from
  testModReference.ts into their own files
- Update all callers; no shim re-exports remain

Collections:
- New impure files: makeInstall, postprocessCollection, walk,
  uploadCollection, makeProgressFunction, renderReference, readCollection,
  modToCollection, cloneCollection, createCollection,
  createCollectionFromProfile, showQuickCollectionRestrictionsDialog,
  cacheEvictions
- util.ts, transformCollection.ts, postProcessRule.ts, testSupported.ts
  now contain only pure logic
- persistent.ts reducer rewritten with immutability-helper
- vitest config switched to happy-dom; original-fs aliased to fs
- Mock surface: 231 → 41 lines; node-7z/tmp/turbowalk/vortexmt mocks
  deleted

Behavioural note: `postProcessRule` now uses the renderer's full
`isFuzzyVersion` instead of the local simplified version. Rules with
range-style versionMatches (>=, ^, ~, partial) plus a logicalFileName
will now have their fileExpression stripped, matching the documented
intent of the backwards-compat hook.

fixes https://linear.app/nexus-mods/issue/APP-413


extensions/collections/__mocks__/vortex-api.ts
```diff
@@ -0,0 +1,300 @@
...
+ *
+ * Only the utilities actually exercised by the code under test are implemented.
+ * Everything else is a no-op / passthrough so imports don't explode.
+ */
+
+// ---------------------------------------------------------------------------
+// util
+// ---------------------------------------------------------------------------
+
+function getSafe(state: any, path: (string | number)[], fallback: any): any {
```
IDCs, line 12: `storehelper` functions do not require mocks - just re-export the real implementation. `setSafe` in this mock is already diverging from the real implementation.

```
// Original
copy[firstElement] = {};

// Mock
copy[firstElement] = typeof path[1] === "number" ? [] : {};
```

TBH most of the functions here do not need re-implementation as a mock.

extensions/collections/__mocks__/bsdiff-node.ts
```diff
@@ -0,0 +1,2 @@
+export async function diff() {}
```
IDCs, line 1: The `vitest.config.ts` is set to only include `src/**/*.test.ts` - `bsdiff-node.test.ts` is currently _not_ inside `src` which means it never runs.

Even if it was inside `src` - this mock would basically turn it into a no-op.

If the smoke test is deliberately being retired, delete it rather than leaving a dead file next to a mock that would break it. (FYI - I'd prefer we don't get rid of it)

extensions/collections/__mocks__/vortex-api.ts
```diff
@@ -0,0 +1,300 @@
...
+  // Make it behave as a passthrough proxy for property access
+  return new Proxy(proxy, {
+    get(_target, prop) {
+      const obj = proxy();
+      return (obj as any)?.[prop];
+    },
+  });
+}
+
+export const util = {
```
IDCs, line 162: I doubt this will be the last time we will have to mock `vortex-api` - SDV already does it. I think we need to centralize this somehow otherwise we're going to have duplicate mocks all over the repo.

@erri120 - this might be a good example of a required package `packages/test-utils` or something.

extensions/collections/__mocks__/modmeta-db.ts
```diff
@@ -0,0 +1,7 @@
+export interface ILookupResult {
+  key: string;
```
IDCs, line 2: types don't need mocking, just import them from the real packages

extensions/collections/__mocks__/nexus-api.ts
```diff
@@ -0,0 +1,11 @@
+export interface ICreateCollectionResult {
```
IDCs, line 1: types don't need mocking, just import them from the real packages

IDCs (changes requested):
Good spot on that revisionId bug - but I really don't know if this PR will get merged in. We need to have a team discussion regarding testing for the sake of not duplicating test code all over the place.

erri120 (approved):

PR #22450, opened 2026-04-14
Change resolveGameFolders to use store/game QualifiedPath bases

halgari:
## Summary

Refactors the `IGamePathService.resolveGameFolders` contract so adaptors receive **store-scoped and game-scoped base `QualifiedPath`s** instead of a raw store name string and install path. This lets adaptors express folder locations *outside* the game's install directory (saves in Documents, preferences in AppData, etc.) using the store's path scheme.

### Before

```ts
resolveGameFolders(store: string, installPath: QualifiedPath): Promise<GameFolderMap>
```

Adaptors received the store name and the discovered install path, and could only describe folders relative to the installation directory.

### After

```ts
resolveGameFolders(storeBase: QualifiedPath, gameBase: QualifiedPath): Promise<GameFolderMap>
```

- **`storeBase`** (e.g. `steam://1091500`) -- the adaptor joins well-known OS sub-paths onto this to describe where folders actually live on disk:

  | Sub-path | Resolves to |
  |---|---|
  | `storeBase/install` | Game install directory |
  | `storeBase/home` | `%USERPROFILE%` |
  | `storeBase/documents` | `%USERPROFILE%/Documents` |
  | `storeBase/my games` | `%USERPROFILE%/Documents/My Games` |
  | `storeBase/appData` | `%USERPROFILE%/AppData` |

- **`gameBase`** (e.g. `game://steam/1091500`) -- passed for context; the framework maps each returned folder name onto this to build the game resolver.

### Example (Cyberpunk 2077)

```ts
resolveGameFolders(storeBase, _gameBase) {
  return {
    [GameFolder.install]:     qpath`${storeBase}/install`,
    [GameFolder.saves]:       qpath`${storeBase}/home/Saved Games/CD Projekt Red/Cyberpunk 2077`,
    [GameFolder.preferences]: qpath`${storeBase}/appData/Local/CD Projekt Red/Cyberpunk 2077`,
  };
}
```

### Other changes

- Removes `GameFolder.config` and `GameFolder.cache` (zero consumers in the codebase).
- Adds `storeGameId()` helper to look up store-specific game IDs from `GameInfo` metadata.
- Adds `buildBasePaths()` to construct the serialized `QualifiedPath` pair for IPC.
- Updates type-level tests to match the new signature and reduced `GameFolder` union.

### DIFF
--- a/packages/adaptor-api/src/contracts/game-paths.test.ts
+++ b/packages/adaptor-api/src/contracts/game-paths.test.ts
@@ -11,7 +11,7 @@ import type {
 describe("GameFolder", () => {
   it("is a union of well-known folder names", () => {
     expectTypeOf<GameFolder>().toEqualTypeOf<
-      "install" | "saves" | "preferences" | "config" | "cache"
+      "install" | "saves" | "preferences"
     >();
   });
 });
@@ -35,5 +35,11 @@ describe("IGamePathService", () => {
     expectTypeOf<
       IGamePathService["resolveGameFolders"]
     >().returns.resolves.toMatchTypeOf<GameFolderMap>();
+  });
+
+  it("accepts storeBase and gameBase QualifiedPath parameters", () => {
+    expectTypeOf<
+      IGamePathService["resolveGameFolders"]
+    >().parameters.toEqualTypeOf<[QualifiedPath, QualifiedPath]>();
   });
 });
--- a/packages/adaptor-api/src/contracts/game-paths.ts
+++ b/packages/adaptor-api/src/contracts/game-paths.ts
@@ -8,8 +8,6 @@ export const GameFolder = {
   install: "install",
   saves: "saves",
   preferences: "preferences",
-  config: "config",
-  cache: "cache",
 } as const;
 
 export type GameFolder = (typeof GameFolder)[keyof typeof GameFolder];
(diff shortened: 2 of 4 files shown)


halgari:
@erri120 we should probably discuss this, I'm not thrilled about how the paths stuff interacts here. The problem comes down to how we handle the mapping of game specific folders: `saves`, `installation`, `config`, and their mapping to store folders `{documents}/Skyrim`. The problem I see with this approach comes down to live updating and restarting. If the host caches the resolve mappings from the adaptor to go from `game://...` -> `steam://...` then whenever the adaptor restarts we may have stale data. If we make the adaptor a resolver that performs that transform then we have way more RPC and the resolve becomes a multi-process RPC chain just to resolve a path. 

However I slice this I'm not thrilled about it.

erri120:
> @erri120 we should probably discuss this, I'm not thrilled about how the paths stuff interacts here. The problem comes down to how we handle the mapping of game specific folders: `saves`, `installation`, `config`, and their mapping to store folders `{documents}/Skyrim`. The problem I see with this approach comes down to live updating and restarting. If the host caches the resolve mappings from the adaptor to go from `game://...` -> `steam://...` then whenever the adaptor restarts we may have stale data. If we make the adaptor a resolver that performs that transform then we have way more RPC and the resolve becomes a multi-process RPC chain just to resolve a path.
> 
> However I slice this I'm not thrilled about it.

What about something like this:

```typescript
type GamePaths<T extends string> = {
  [K in T]: QualifiedPath;
};

type MyGamePaths = "game" | "saves";

class MyGame {
  async paths(
    storeProvider: StorePathProvider,
  ): Promise<GamePaths<MyGamePaths>> {
    const game = await storeProvider.fromBase("game");
    let saves: QualifiedPath;
    if (storeProvider.isWindows) {
      const documents = await storeProvider.fromBase("documents");
      saves = documents.join("My Game", "saves");
    } else {
      saves = game.join("saves");
    }

    return {
      game,
      saves,
    };
  }
}
```

Adaptor implements a method `paths`/`getPaths` that accepts a `StorePathProvider` and returns an object with string keys and `QualifiedPath` values. All paths are created from values provided by the `StorePathProvider`. The object can be stored on the host and used when calling other methods on the adaptor. Since it's generic, the paths are game specific (we can force `game` or other essential keys):

```typescript
  async doSomething(fs: FileSystem, paths: GamePaths<MyGamePaths>): Promise<void> {
    await fs.stat(paths.saves);
  }
```

No RPC required for resolving paths.

halgari:
Closing the PR as this work will be folded into some more work I'm doing today

PR #22510, opened 2026-04-16
Add download API to preload script

erri120:
Closes APP-227.

### DIFF
--- a/src/shared/src/types/ipc.ts
+++ b/src/shared/src/types/ipc.ts
@@ -2,6 +2,7 @@
 // Everything in here is compile-time only, meaning the interfaces you find here
 // are never used to create an object. They are only used for type inferrence.
 
+import type { ByteRange } from "./download";
 import type { SerializedSpan } from "../telemetry/types";
 import type {
   BrowserViewConstructorOptions,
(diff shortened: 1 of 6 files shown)


src/shared/src/types/ipc.ts
```diff
@@ -336,6 +356,14 @@ export interface InvokeChannels {
   // Compile stylesheets
   "styles:compile": (filePaths: string[]) => Promise<string>;
 
+  // Download channels
+  "download:start": (
```
IDCs, line 360: Not a biggie, but the naming of this channel got me thinking for a second - primarily about the fact that we're just sending off the destination as a parameter (without actually specifying what we're downloading); had to look at the code to understand that this is basically a per design `enqueue` to this path rather than a `start` downloading this file to `dest`.

So for example if I were to use this channel via a community extension, I'd be wondering why it's asking me just for a destination.

IDCs (approved):
Overall pretty happy with this, but as I mentioned previously - the `download:start` channel naming is somewhat misleading as a contributor.

src/shared/src/types/ipc.ts
erri120: The IPC mechanism should never be exposed to extensions directly. They would just use the existing API as usual.

PR #22644, opened 2026-04-21
Fix formatting of game name for Fallout: New Vegas

Pickysaurus:
Removed the `/t` character. I believe this was added as a workaround for the old games UI?


IDCs:
- FNV is not the only game that uses tabs in its name. This is a band aid fix for a single game extension which is probably going to get replaced by an adaptor.
- Fix needs to be done on the UIV2 rendering component to cater for all game extensions that use this pattern, regardless of whether it's internal or a community extension.

PR #22688, opened 2026-04-22
Replace native C++ savegame parser with pure TypeScript

halgari:
## Summary

Replaces the `gamebryo-savegame` native C++ addon (NAPI) with pure TypeScript parsers for all 6 Gamebryo savegame formats. This eliminates the node-gyp, CMake, and MSVC build toolchain requirements for this extension.

- **Add** `src/savegame/` — pure TS parsers for Oblivion, Skyrim, Skyrim SE, Fallout 3, Fallout NV, and Fallout 4 save formats
- **Fix** zlib decompression: use `inflateSync` instead of `inflateRawSync` to match the C++ `inflateInit` behavior for zlib-wrapped streams (Skyrim SE saves)
- **Replace** `gamebryo-savegame` native dependency with `lz4js` (~20KB, pure JS LZ4 decompression)
- **Remove** native module remapping from `build.mjs` and Windows-only platform guard from build/dist scripts
- **Add** vitest tests with 18 savegame fixtures (3 per format) covering quick reads, full reads, plugin lists, screenshot hashes, and metadata

### Performance vs C++ native

| Operation | C++ native | TypeScript | Notes |
|-----------|-----------|------------|-------|
| Quick read (header only) | ~0.12ms | ~0.03ms (4x faster) | Only reads ~256 bytes |
| Full read (single save) | ~0.13ms | 3-11ms | Pure-JS LZ4 for compressed saves |
| Full list (200 saves) | ~26ms | ~6ms | Quick mode is the common path |

Quick reads (the common path for listing saves) are 4x faster. Full reads with decompression are slower due to pure-JS LZ4, but imperceptible to users (~11ms worst case).


extensions/gamebryo-savegame-management/src/savegame/index.ts
erri120, line 1: Should just use `parseSaveGame` directly at this point since the native module doesn't exist anymore.

halgari: done

IDCs (approved):

PR #22697, opened 2026-04-22
Replace native vortexmt with pure TypeScript MD5 hashing

halgari:
## Summary

Removes the `vortexmt` C++ native addon and replaces all usage with pure TypeScript using Node's built-in `crypto.createHash('md5')` + `fs.createReadStream()`.

`vortexmt` was a ~595-line C++ NAPI module that reimplemented MD5 from scratch and used an AsyncProgressWorker to hash files in 8KB chunks. This is unnecessary because Node's `crypto` module (backed by OpenSSL) provides the same functionality natively, and streaming file I/O via `createReadStream` is already non-blocking (libuv thread pool handles the I/O, each chunk's `hash.update()` is fast).

## History

`vortexmt` was written by TanninOne in September 2019 (Node 12 era). At the time, while Node's crypto module could do streaming MD5, there was no clean one-liner for "hash this file with progress reporting." Writing a native addon with NAPI's AsyncProgressWorker was a reasonable shortcut to get `fileMD5(path, callback, progressCb)` as a single call. A progress callback was added in June 2021, Buffer support in October 2025, and the module was otherwise only touched for maintenance (dependency bumps, prebuild setup, Linux build support). The pure-JS equivalent just never got revisited.

## What changed

- **`checksum.ts`**: Enhanced `fileMD5` to accept `string | Buffer` (was string-only) and an optional `progress` callback. Buffers are hashed synchronously (already in memory). Files are stat'd for size, then streamed.
- **`ExtensionManager.ts`**: Switched from `vortexmt` to `checksum.fileMD5`. Simplified `genMd5Hash` since `fileMD5` now returns a Promise directly, removing the `toPromise` callback wrapper.
- **`download_management/index.ts`**: Same switch. Replaced `toPromise((cb) => fileMD5(filePath, cb, () => {}))` with `fileMD5(filePath)`. Fixed untyped catch block.
- **`gameversion-hash/index.ts`**: Removed `vortexmt` import and the `fileMD5Async` callback wrapper. Uses `util.fileMD5` from vortex-api directly.
- **`collections/util.ts`**: Same pattern. Removed `vortexmt` import, simplified `fileMD5Async` to delegate to `util.fileMD5`.
- **`checksum.test.ts`** (new): 9 tests covering RFC 1321 test vectors, file hashing (empty, binary, 1MB), progress callbacks, and error handling.

## Performance

Benchmarked on the same machine, same Node.js process, random data at each size. Hashes verified identical at every size.

| Size | TypeScript | Native C++ | Ratio |
|------|-----------|------------|-------|
| 1 KB | 0.3ms | 0.1ms | 2.70x slower |
| 1 MB | 2.4ms | 1.8ms | 1.32x slower |
| 10 MB | 22.7ms | 17.8ms | 1.27x slower |
| 100 MB | 200.7ms | 187.5ms | 1.07x slower |
| 512 MB | 994.6ms | 956.7ms | 1.04x slower |

The overhead is fixed-cost stream/event-loop setup (~0.2ms). At file sizes that matter for mod archives (100MB+), the difference is 4-7% and both implementations are I/O-bound. The 2.7x on 1KB is 0.2ms absolute, imperceptible in practice.

## Motivation

- Eliminates a native C++ build dependency (node-gyp, MSVC toolchain)
- One fewer `.node` binary to prebuild and ship per platform
- Removes a hand-rolled MD5 implementation in favor of OpenSSL (via Node crypto)
- Simplifies call sites: promise-based API instead of callback wrapping

### DIFF
--- a/extensions/collections/package.json
+++ b/extensions/collections/package.json
@@ -48,7 +48,6 @@
     "ts-v-gen": "catalog:",
     "turbowalk": "catalog:",
     "typescript": "catalog:",
-    "vortex-api": "workspace:*",
-    "vortexmt": "catalog:"
+    "vortex-api": "workspace:*"
   }
 }
--- a/extensions/collections/src/util/util.ts
+++ b/extensions/collections/src/util/util.ts
@@ -11,7 +11,6 @@ import { doExportToAPI } from "../collectionExport";
 import { ICollectionModRuleEx } from "../types/ICollection";
 import { IEntryEx } from "../types/IEntryEx";
 import { IModEx } from "../types/IModEx";
-import { fileMD5 } from "vortexmt";
 import turbowalk, { IEntry, IWalkOptions } from "turbowalk";
 import { TOS_URL } from "../constants";
 
@@ -146,14 +145,7 @@ export function calculateCollectionSize(mods: {
 }
 
 export async function fileMD5Async(fileName: string): Promise<string> {
-  return new Promise((resolve, reject) => {
-    fileMD5(
-      fileName,
-      (err: Error, result: string) =>
-        err !== null ? reject(err) : resolve(result),
-      () => null,
-    );
-  });
+  return util.fileMD5(fileName);
 }
 
 export async function walkPath(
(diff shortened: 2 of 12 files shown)


IDCs (approved):

erri120:
It looks like no extension is importing `vortexmt` directly based on static analysis, so this should be safe to remove.

PR #22719, opened 2026-04-23
Add load order and prelaunch adaptor contracts

halgari:
## Summary

Milestone 4 of the Game Adaptors project: load order management, prelaunch tool orchestration, mod file access, and Vortex integration. Proven end-to-end with Cyberpunk 2077 as the reference adaptor.

### Load Order (`contracts/load-order.ts`)
- `IGameLoadOrderService<T>`: multiple load orders per game, entry management with enable/disable, sorting rules (before/after constraints), and disk serialization notifications
- Covers APP-281, APP-282, APP-283

### Prelaunch (`contracts/prelaunch.ts`)
- `IGamePrelaunchService<T>`: declarative prelaunch task definitions with conditional execution
- Covers APP-285, APP-286

### Mod Files (`contracts/mod-files.ts`)
- `IModFilesService`: host-provided service for gated mod file access via QualifiedPaths
- `getInstalledMods()`, `getModFiles(modId)`, `getDeployedFiles()` (flattened view)
- Adaptors never get raw disk access; all paths go through qualified path resolvers
- Covers APP-287, APP-288

## Adaptor bridge integration

- **Load order bridge**: each adaptor load order definition registers with Vortex's `file_based_loadorder` system. Serialize/deserialize callbacks delegate to the adaptor via IPC. Covers APP-322.
- **Prelaunch start hook** (priority 90): queries the adaptor's prelaunch service before game launch, evaluates conditional tasks, logs execution intent. Covers APP-319.

### Archive load order (APP-296)
- Scans `archive/pc/mod/` via the host filesystem service for `.archive` and `.xl` files
- Reads existing `modlist.txt` to preserve user ordering
- Writes `modlist.txt` on serialize (first = highest priority, matching game convention)
- Removes `modlist.txt` when no entries are enabled (falls back to alphabetical)

### REDmod load order (APP-295)
- Declared as a second load order (REDmod packages under `mods/`)
- REDmod ordering is alphabetical by folder name (no config file needed)

### Conditional REDmod prelaunch (APP-297)
- Checks for REDmod packages under `mods/` via the filesystem service
- Skips the deploy tool when no REDmods are present
- Runs `redMod.exe deploy` when REDmods exist

### Multiple load orders per game not yet supported in UI
The adaptor API supports declaring multiple load orders per game (e.g. Cyberpunk's archive and REDmod orderings), but Vortex's `file_based_loadorder` system deduplicates by `gameId`, so only the first load order registers. The REDmod load order is silently dropped. This needs UX design work and a load order system that supports multiple tabs/views per game before it can be surfaced properly.

## Linear issues covered

**Game Adaptors project (milestone 4):**
APP-281, APP-282, APP-283, APP-285, APP-286, APP-287, APP-288, APP-319, APP-322

**In-House Cyberpunk Extension project:**
APP-295, APP-296, APP-297


src/renderer/src/extensions/adaptor_bridge/index.ts
```diff
@@ -681,6 +871,75 @@ function init(context: IExtensionContext): boolean {
...
+  // Register a prelaunch start hook that runs adaptor prelaunch tasks
+  // before the game launches. Priority 90 runs after deployment checks
+  // (100) but before the actual launch. The hook queries the adaptor's
+  // prelaunch service, evaluates conditions, and runs qualifying tasks.
+  context.registerStartHook?.(
+    90,
+    "adaptor-prelaunch",
+    async (input) => {
+      const state = context.api.getState();
+      const profile = (state as any).settings?.profiles?.lastActiveProfile;
```
IDCs: No need for `any` here - we know the type, it's `IState`.

Note that `lastActiveProfile` is typed as `lastActiveProfile: { [gameId: string]: string };`

So I think `profile` should be changed to `profileId` so it's explicit, and the `activeGameId` selector should be used to specify which `lastActiveProfile` for which game.

e.g.
```
const gameId = selectors.activeGameId(state);
const profileId = state.settings?.profiles?.lastActiveProfile?.[gameId];
```

There's also a selector you could use:
```
const lastActiveProfileId = lastActiveProfileForGame(state, gameId);
```

src/renderer/src/extensions/adaptor_bridge/index.ts
```diff
@@ -681,6 +871,75 @@ function init(context: IExtensionContext): boolean {
     },
   );
 
+  // Register a prelaunch start hook that runs adaptor prelaunch tasks
+  // before the game launches. Priority 90 runs after deployment checks
```
IDCs: The comment is incorrect - hooks are running in ascending order. So this hook will run _before_ `check-deployment`

src/renderer/src/extensions/adaptor_bridge/index.ts
IDCs: Theoretically you don't need any of this if you're just checking that gameId is defined.

```
const gameId = selectors.activeGameId(state);
 if (!gameId) return input;
```

Should be sufficient?

IDCs (changes requested):
Left a few comments, everything else looks fine.

IDCs (dismissed):

PR #22829, opened 2026-04-27
Fingerprinting workflows

Aragas:
- Adds three GitHub workflows that record error fingerprint resolution in ClickHouse: PR-merge marks fingerprints as `fixed`, `v*` tag push marks them as `released`, and a manual workflow lets ops add or remove entries by hand.
- All three are thin orchestration over a single Node 24 action at `.github/actions/fingerprints/` with a `mode: pr | release | resolve` input.
- Closes the loop on telemetry — errors reported with the stable 8-char fingerprints generated in `src/shared/src/errors.ts` can now be tracked from "fix merged" through "fix shipped" without manual SQL.

## Workflows

| Trigger | Workflow | Behaviour |
|---|---|---|
| `pull_request: closed` (merged) | `fingerprint-fixed.yml` | Parses `Fixes fingerprint XXXXXXXX` lines from the merged PR's body, inserts each as `status: fixed`. |
| `push: tags v*` | `fingerprint-released.yml` | Walks PRs merged since the previous `v*` tag, marks every referenced fingerprint as `status: released` for that version. |
| `workflow_dispatch` | `fingerprint-resolve.yml` | Manual add/remove. Inputs: `fingerprints` (comma- or whitespace-separated 8-hex chars), `remove` (bool), `status` (`fixed`/`released`), `release_version`. |


.github/actions/fingerprints/src/clickhouse.ts
```diff
@@ -0,0 +1,75 @@
...
+  const values = rows.map((r) => ({
+    fingerprint: r.fingerprint,
+    pr_url: r.pr_url ?? "",
+    updated_at: updatedAt,
+    updated_by: r.updated_by ?? "",
+    release_version: r.release_version ?? "",
+    status: r.status,
+  }));
+
+  await client.insert({
```
IDCs, line 21: Never used the clickhouse client so might not be a problem - can it insert duplicate rows?

Should do `INSERT … WHERE NOT EXISTS` under the hood if possible.

Aragas: Duplicates are not possible, the table will override automatically instead, IIRC

.github/actions/fingerprints/action.yml
```diff
@@ -0,0 +1,42 @@
...
+  clickhouse-user:
+    description: ClickHouse username.
+    required: true
+  clickhouse-password:
+    description: ClickHouse password.
+    required: true
+
+runs:
+  using: node24
+  main: dist/index.js
```
IDCs, line 42: Have you tried running this locally? I'm pretty sure `dist/index.js` doesn't exist when running this workflow.

Aragas: dist/index.js will be compiled, but I don't remember whether we need the compiled code present or not. I think it will do pnpm install and pnpm build and the run dist/index.js

IDCs: Perhaps I'm misunderstanding how github actions would work in this scenario, but none of the workflows is calling install or build. It looks like `action.yml` will try to call `node run dist/index.js` which will complain that `dist/index.js` is missing as it was never compiled.

IDCs: I guess I'll approve and we can fix it if it's broken.

IDCs (dismissed):

.github/actions/fingerprints/action.yml
Aragas: I think you were right. I've included dist code into the sources

IDCs (approved):

PR #22865, opened 2026-04-28
QA-82 Playwright - Login to Vortex function

RedRanger14:
login test covering the full OAuth flow (open dialog → capture URL → launch Chrome → fill credentials → authorise → verify logged-in state) - UI login test.

I then made it into a function for any future tests to use that need a login.

*note: apologies for the branch name, I didn't realise this needed doing before collections could be touched.


packages/e2e/helpers/login.ts
```diff
@@ -0,0 +1,129 @@
...
+    expect(oauthUrl).toMatch(/^https?:\/\//i);
+
+    if (loginPage !== null) {
+      await loginPage.waitForLoadState("domcontentloaded");
+      await expect(loginPage).toHaveURL(/nexusmods|users\./i);
+    }
+
+    authBrowser = await chromium.launch({
+      channel: "chrome",
+      headless: false,
```
IDCs: Two issues:

`headless: false` opens a real browser window. That works locally (you have a screen), and it works on GitHub's runners because they have a virtual display. But it'll fail on any CI/Docker/self-hosted environment without one - the launch errors with "Missing X server or $DISPLAY". It also makes tests slower and uses more memory.

`channel: "chrome"` tells Playwright to use the system-installed Google Chrome instead of the Chromium that we already download in the workflow (playwright install chromium). It works today only because GitHub's runner images happen to include Chrome - if that ever changes, or if anyone runs this in a container, you get "Browser distribution 'chrome' is not found".

Could we change it to:

```
authBrowser = await chromium.launch({
  headless: !process.env.PWDEBUG,
});
```

This matches what the Vortex fixture already does [vortex-app.ts:55-58](https://github.com/Nexus-Mods/Vortex/blob/ebd26a9e752237e34c9f460259e97804f09c385e/packages/e2e/fixtures/vortex-app.ts#L57) headless by default, and a visible window only when you set PWDEBUG=1 to debug locally.

packages/e2e/helpers/login.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    const nexusLoginPage = new LoginPage(authPage);
+
+    await expect(nexusLoginPage.authoriseButton).toBeVisible();
+    await nexusLoginPage.authoriseButton.click();
+
+    await expect(nexusLoginPage.authorisationSuccessTitle).toBeVisible();
+
+    if (authBrowser !== null) {
```
IDCs: Not a biggie, but - `authBrowser` is only closed on the success path. If login form/2FA/network hiccup fail, the spawned browser is orphaned. Wrap this in a try/finally block

packages/e2e/helpers/login.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    const nexusLoginPage = new LoginPage(authPage);
+
+    await test.step("Enter username", async () => {
+      await nexusLoginPage.usernameInput.fill(username);
+      await expect(nexusLoginPage.usernameInput).toHaveValue(username);
+    });
+    await test.step("Enter password", async () => {
+      await nexusLoginPage.passwordInput.fill(password);
+      await expect(nexusLoginPage.passwordInput).toHaveValue(password);
```
IDCs: When this assertion fails, Playwright includes the expected value verbatim in the error message - which then gets written into the trace files that the workflow uploads as artifacts. Anyone with repo read access can download those.

`fill()` already throws if it can't set the value, so the assertion isn't adding coverage. Suggest dropping it:
```
await test.step("Enter password", async () => {
  await nexusLoginPage.passwordInput.fill(password);
});
```
(or `expect(...).not.toHaveValue("")` if you want a non-revealing assertion).

IDCs (changes requested):

packages/e2e/helpers/users.ts
```diff
@@ -0,0 +1,19 @@
...
+
+function requireEnvVar(name: string): string {
+  const value = process.env[name];
+  if (!value) throw new Error(`Missing required environment variable: ${name}`);
+  return value;
+}
+
+export const freeUser: NexusUser = {
+  get username() {
+    return requireEnvVar("E2E_NEXUS_FREE_USER_USERNAME");
```
IDCs, line 14: There are currently no secrets set up in the repo that match these environment variables.
`E2E_NEXUS_FREE_USER_USERNAME`
`E2E_NEXUS_FREE_USER_PASSWORD`

`e2e.yml` doesn't even set those up as environment variables.

We either need to add those for the CI to use, or make `e2e.yml` configurable so it accepts manual input from whoever is running the workflow.

packages/e2e/helpers/login.ts
RedRanger14: Yeah sorry, I think this is from when I was trying to get Vortex to show in the UI mode

packages/e2e/helpers/users.ts
RedRanger14: Can you help me with this one please @IDCs ?

IDCs: Sure, would you rather have it as manual input or use the `E2E` GH secrets option?

packages/e2e/helpers/login.ts
RedRanger14: I have added 

`authBrowser = await chromium.launch({
        channel: "chromium",
        headless: !process.env.PWDEBUG,
      });`
  
  I think, headless: !process.env.PWDEBUG keeps CI headless by default and channel: "chromium" avoids relying on system Google Chrome. This test does not work without it, not sure why as I would have thought it would be covered in the playwright install

packages/e2e/helpers/users.ts
RedRanger14: Basically, I made an env variable as I didn't know we had one. Happy to add them to the existing one or use the one I've made for test users? Maybe we could have a chat about how we want to store the sensitive data for this

IDCs (dismissed):

IDCs (approved):

PR #22872, opened 2026-04-28
Port bsatk (BSA archive reader) to pure TypeScript

halgari:
## Summary

- Replaces the native C++ `bsatk` addon with a pure TypeScript BSA parser
- Merges BSA and BA2 support into a unified `gamebryo-archive-support` extension that registers both `"bsa"` and `"ba2"` archive types
- Supports all three BSA format versions: v0x67 (Oblivion), v0x68 (FO3/FNV/Skyrim LE), v0x69 (Skyrim SE)
- Adds LZ4 frame decompression for Skyrim SE BSAs (which the native C++ module couldn't handle)

## Benchmark (TS vs Native)

| File | Native Extract | TS Extract | Ratio |
|------|---------------|------------|-------|
| RaceMenu.bsa (14MB, 577 files) | 2351ms | 518ms | **0.2x** |
| Harvest [Flora].bsa (9MB, 161 files) | 588ms | 112ms | **0.2x** |
| SkyUI_SE.bsa (2.8MB, 71 files) | 379ms | 62ms | **0.2x** |

The speedup is largely due to eliminating per-file N-API overhead. The native codepath creates a new `AsyncWorker`, queues it on the libuv thread pool, and round-trips back into JS for **every file** — 577 times for RaceMenu.bsa. The TS implementation opens the BSA file descriptor once and iterates all files in a single async loop with positional reads, avoiding the heap allocation + libuv scheduling + JS callback cost per file. The native benchmark also performs extra disk I/O (read-back + delete per file) that the TS benchmark skips, but the dominant factor is the N-API async boundary crossing at scale.

Resolves APP-374


IDCs (dismissed):

extensions/gamebryo-archive-support/src/bsa.ts
```diff
@@ -0,0 +1,725 @@
+import * as fs from "fs";
+import * as path from "path";
+import { inflateSync } from "zlib";
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const lz4js = require("lz4js") as {
+  decompress(src: Uint8Array, maxSize?: number): Uint8Array;
+};
+
```
erri120: What happened here?

halgari: Fixed, not sure how that got through

PR #22898, opened 2026-04-29
Qa 83/login logout of vortex

RedRanger14:
Adding a logout test for Vortex

Added premium user for Vortex

Because all tests need to be independent from one another. I have created a vortex-app-isolated fixture that launches a fresh Electron app and fresh user-data directory per test.

### DIFF
--- a/packages/e2e/fixtures/vortex-app-isolated.ts
+++ b/packages/e2e/fixtures/vortex-app-isolated.ts
@@ -1,0 +1,205 @@
+import {
+  test as base,
+  _electron as electron,
+  expect,
+  type ElectronApplication,
+  type Page,
+} from "@playwright/test";
+import path from "node:path";
+import fs from "node:fs";
+import os from "node:os";
+import { createRequire } from "node:module";
+
+/** Package root (packages/e2e/) — used for resolving node_modules. */
+const PACKAGE_ROOT = path.resolve(import.meta.dirname, "..");
+
+/** Repo root directory. */
+const REPO_ROOT = path.resolve(PACKAGE_ROOT, "..", "..");
+
+// createRequire from the package root so pnpm-linked electron resolves correctly
+const require = createRequire(path.join(PACKAGE_ROOT, "package.json"));
+
+function resolveMainDir(): string {
+  return path.resolve(REPO_ROOT, "src", "main");
+}
+
+function resolveElectronBinary(): string {
+  return require("electron") as unknown as string;
+}
+
+function safeRemoveDir(dir: string): void {
+  try {
+    fs.rmSync(dir, {
+      recursive: true,
+      force: true,
+      maxRetries: 10,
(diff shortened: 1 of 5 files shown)


packages/e2e/fixtures/vortex-app-isolated.ts
```diff
@@ -0,0 +1,205 @@
...
+    // Best-effort cleanup: Windows can keep file handles briefly.
+  }
+}
+
+/**
+ * Build a clean env for the Electron process.
+ * Removes ELECTRON_RUN_AS_NODE (set by pnpm) and isolates user data.
+ */
+function buildElectronEnv(userDataDir: string): Record<string, string> {
+  const env: Record<string, string> = {};
```
IDCs, line 48: As far as I can see, this is copied verbatim (+/- a few comments) from `vortex-app.ts`. There are currently ~150 lines of duplication between the two fixtures, lets extract those functions into their own file and import them into the fixtures to avoid divergence between the two.

As far as I can see the duplicated functions are: `resolveMainDir`, `resolveElectronBinary`, `buildElectronEnv`, `waitForMainWindow`

packages/e2e/fixtures/vortex-app-isolated.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    try {
+      await mainWindow.locator("body").first().waitFor({
+        state: "visible",
+        timeout: 60_000,
+      });
+      await use(mainWindow);
+      return;
+    } catch (error) {
+      if (!mainWindow.isClosed()) {
```
IDCs, line 189: So although your code avoids this at runtime, calling `use` twice within the same fixture is violating the playwright fixture contract. This is potential for undefined behaviour. Additionally we're potentially dropping relevant errors from the first use call.

I suggest something like:
```
vortexWindow: async ({ vortexApp }, use) => {
  const waitForReady = async (win: Page) => {
    await win.waitForLoadState("domcontentloaded");
    await win.waitForFunction(
      () => (document.body?.innerText?.length ?? 0) > 0,
      { timeout: 60_000 },
    );
  };

  let mainWindow = await waitForMainWindow(vortexApp);
  try {
    await waitForReady(mainWindow);
  } catch (error) {
    if (!mainWindow.isClosed()) throw error;
    mainWindow = await waitForMainWindow(vortexApp);
    await waitForReady(mainWindow);
  }

  await use(mainWindow);
},
```

Additionally I can see that `waitForReady` functionality has already diverged between this fixture and the other one. Why not create a new function `waitForRenderReady` to avoid this going forward.

IDCs (changes requested):
Left a couple of comments. Note that I wasn't sure if this PR is ready for review yet as it's in draft, so apologies if I jumped the gun.

erri120:
Closing this PR as it's very outdated and much has changed. I recommend re-doing the test entirely.

PR #22909, opened 2026-04-29
Replace native drivelist addon with enumeration via winapis + koffi

halgari:
## Summary

- Replaces the native C++ `drivelist` addon with pure TS implementations
- **Windows**: Uses [koffi](https://koffi.dev/) FFI to call `GetLogicalDriveStringsW` + `GetDriveTypeW` from kernel32.dll — the standard Win32 APIs for drive enumeration. Instant (no process spawn), filters to `DRIVE_FIXED` (type 3).
- **Linux**: Parses `/proc/mounts` filtering for real block devices on common filesystems (ext4, btrfs, xfs, etc.)
- Removes `drivelist` from both `package.json` files, `pnpm-workspace.yaml` (catalog + allowBuilds), `flatpak/generated-sources.json`, and the dependency report
- The `getDriveList()` API and error handling behavior are preserved — callers are unchanged

Resolves APP-403

## Test plan

- [x] All 1188 tests pass
- [x] `pnpm install` succeeds without drivelist
- [x] Verified koffi drive enumeration returns correct drive letters on Windows 11


Aragas:
It's usually not a good idea to spawn a new process that does some action, are we sure it's an acceptable solution?

halgari:
The code this replaces calls out to wmi, but now that we already have koffi in place we may be able to replace it with calls into the Nt API, I'll look down that path

halgari:
Done, swapped to koffi, and the implementation remains rather clean.

src/renderer/src/extensions/gamemode_management/util/getDriveList.ts
```diff
@@ -1,50 +1,111 @@
...
+/**
+ * Enumerate fixed drives via Win32 GetLogicalDriveStringsW + GetDriveTypeW.
+ *
+ * GetLogicalDriveStringsW returns a double-null-terminated buffer of drive
+ * root paths ("C:\\\0D:\\\0\0"). GetDriveTypeW returns the type for each;
+ * DRIVE_FIXED (3) is what we want.
+ */
+function getFixedDrivesWindows(): Promise<string[]> {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const koffi = require("koffi");
```
erri120:
```suggestion
  const koffi = await import("koffi");
```

PR #22910, opened 2026-04-30
Replace native turbowalk with pure TypeScript workspace package

halgari:
## Summary

- Replaces the external native `turbowalk` package (which wraps `winapi-bindings`' `WalkDir()` on Windows) with a local workspace package at `packages/turbowalk`
- **Windows**: Uses [koffi](https://koffi.dev/) FFI to call NT kernel APIs directly — no C++ compilation needed
- **Linux**: Falls back to async `fs.readdir` + concurrent `lstat` (64 workers)
- Properly checks `FILE_ATTRIBUTE_HIDDEN` on Windows via NT directory info (not dot-prefix convention)
- Supports batched progress callbacks with threshold (matching old native behavior)
- Returns `Bluebird<void>` to match the type signature callers expect
- **Zero import changes** — the workspace package uses the same name (`turbowalk`) so all 30+ import sites continue to work unchanged
- Updated 14 `package.json` files from `catalog:` to `workspace:*` and removed the catalog entry
- Removed turbowalk from `flatpak/generated-sources.json` and the dependency report
- Added koffi to `pnpm-workspace.yaml` allowBuilds (koffi ships prebuilds for all platforms — no C++ toolchain needed)

### Why `NtQueryDirectoryFile` instead of `FindFirstFileW`/`FindNextFileW`

The old C++ addon (`winapi-bindings`) used `NtQueryDirectoryFile` from `ntdll.dll` — the low-level NT kernel directory enumeration API. This is the same API that `FindFirstFileW` calls internally, but with a critical difference: **it returns multiple directory entries per syscall** in a caller-provided buffer.

Each entry in the buffer is a variable-length `FILE_FULL_DIR_INFORMATION` struct containing:
- `FileAttributes` (hidden, directory, reparse point flags)
- `EndOfFile` (file size)
- `LastWriteTime` (modification time as FILETIME)
- `FileName` (variable-length UTF-16)

This means a single `NtQueryDirectoryFile` call replaces what would be N separate `FindNextFileW` calls, and no separate `lstat()`/`GetFileAttributesW()` calls are needed — all metadata comes back inline.

We initially implemented the koffi walk using `FindFirstFileW`/`FindNextFileW`, which was 2.2-2.9x slower than the old native addon because each entry required its own FFI round-trip. Switching to `NtQueryDirectoryFile` eliminated that per-entry overhead.

### The 1KB buffer size

The old C++ addon used a 1KB buffer (`static const unsigned int BUFFER_SIZE = 1024` in `walk.cpp`). We initially used 64KB thinking "bigger buffer = fewer syscalls = faster." Benchmarking showed the opposite:

| Buffer size | Median (89k entries) |
|-------------|---------------------|
| **1 KB** | **382ms** |
| 4 KB | 370ms |
| 16 KB | 440ms |
| 64 KB | 586ms |
| 256 KB | 568ms |
| 1024 KB | 2160ms |

**Why smaller is faster:** koffi marshals the entire buffer across the FFI boundary on every `NtQueryDirectoryFile` call. With a 64KB buffer, it copies 64KB of data even if the kernel only filled a few hundred bytes of entries. The 1KB buffer minimizes this marshaling overhead while still fitting multiple entries per syscall (a typical entry with a short filename is ~100 bytes, so 1KB fits ~8-10 entries per call).

The sweet spot is 1-4KB. We use 1KB to match the proven C++ implementation.

### Other Win32 APIs considered

- **`FindFirstFileW`/`FindNextFileW`**: Higher-level Win32 API. One entry per call, each crossing the FFI boundary. 2.2-2.9x slower than native.
- **`GetFileInformationByHandleEx` with `FileIdBothDirectoryInfo`**: Another bulk API, but requires opening a handle first and doesn't offer meaningful advantages over `NtQueryDirectoryFile`.
- **`NtQueryDirectoryFile` with `FileFullDirectoryInformation`**: What we use. Same API the old C++ addon used. Returns attributes, size, and timestamps inline — no separate stat calls needed.

## Benchmark (NVMe — Samsung 970 EVO 1TB, Windows 11, Node 24)

All times are median of 5-10 runs on a warm filesystem cache.

| Directory | Total entries | Old Native (C++) | koffi NtQuery | Ratio |
|-----------|-------------|-----------------|---------------|-------|
| src/ | 1,911 | 8.8ms | 7.0ms | **0.8x (faster)** |
| extensions/ | 3,934 | 22.1ms | 20.4ms | **0.9x (faster)** |
| node_modules/ | 89,390 | 341ms | 383ms | **1.1x** |

The koffi FFI path **matches or beats** the old C++ addon on typical directory sizes (~1k-10k entries). The small regression on the 89k-entry stress test is from residual FFI call overhead (one `CreateFileW` + N `NtQueryDirectoryFile` + one `CloseHandle` per directory).

> **Note:** Benchmarks run on an NVMe SSD. The performance profile should be reliable as the bottleneck is FFI call overhead, not disk I/O.

Resolves APP-405

## Test plan

- [x] All 1188 tests pass
- [x] `pnpm install` succeeds with workspace package resolution
- [x] Same `IEntry` / `IWalkOptions` / `Bluebird<void>` return types — API-compatible
- [x] Benchmarked on NVMe: matches old native performance (results above)
- [x] `FILE_ATTRIBUTE_HIDDEN` properly checked via NT directory info
- [x] CI green


src/renderer/src/extensions/move_activator/index.ts
```diff
@@ -278,9 +278,11 @@ class DeploymentMethod extends LinkingDeployment {
...
         details: true,
       },
-    ).then(() =>
-      PromiseBB.map(links, (entry) => this.restoreLink(entry.filePath)),
-    );
+    )
+      .then(() =>
+        PromiseBB.map(links, (entry) => this.restoreLink(entry.filePath)),
+      )
+      .then(() => undefined);
```
Aragas: I believe we usually do `.then(() => {});`

Aragas:
I'll be a bit pedantic, but I wouldn't say it's a pure TS solution, we still have a native dependency, it's just out of our maintenance scope now.
We do need to maintain still the WinAPI calls, they are just demoted to a dynamic solution instead of a strict C++ solution.
Are we sure it's the direction we want to take?

packages/turbowalk/package.json
erri120, line 1: Missing build script

PR #22921, opened 2026-04-30
[QA-182][QA-99] Add game management function, manage sdv e2e

Marcin-NM:


RedRanger14 (approved):

Aragas (approved):
Should be good. We'll need to make game mocking more concrete - a generic mocker most likely isn't the best solution

PR #22923, opened 2026-04-30
Remove dead code

erri120:
Closes APP-391.

### DIFF
--- a/src/renderer/src/IPCDownloadAdapter.ts
+++ b/src/renderer/src/IPCDownloadAdapter.ts
@@ -626,7 +626,7 @@ export class IPCDownloadAdapter {
       log("debug", "pausing download", { downloadId });
       const checkpoint = await window.api.downloader.pause(downloadId);
       this.#api.store.dispatch(setDownloadCheckpoint(downloadId, checkpoint));
-      this.#api.store.dispatch(pauseDownload(downloadId, true, []));
+      this.#api.store.dispatch(pauseDownload(downloadId, true));
       callback?.(null);
     } catch (err) {
       callback?.(unknownToError(err));
@@ -707,7 +707,6 @@ export class IPCDownloadAdapter {
             downloadId,
             state.bytesReceived,
             state.size ?? 0,
-            [],
             [],
           ),
         );
--- a/src/renderer/src/extensions/download_management/actions/state.test.ts
+++ b/src/renderer/src/extensions/download_management/actions/state.test.ts
@@ -15,11 +15,11 @@ describe("addLocalDownload", () => {
 
 describe("downloadProgress", () => {
   it("creates the action", () => {
-    const action = actions.downloadProgress("id", 42, 43, [], []);
+    const action = actions.downloadProgress("id", 42, 43, []);
     expect(action).toEqual({
       error: false,
       type: "DOWNLOAD_PROGRESS",
-      payload: { id: "id", received: 42, total: 43, chunks: [], urls: [] },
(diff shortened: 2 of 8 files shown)


IDCs (changes requested):
Removing the handler makes sense - it's been a `callback(0)` no-op since `feat/app-257` deleted the old `manager.getFreeSlots()`. The remaining emitter is at [InstallManager.ts:227](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/extensions/mod_management/InstallManager.ts#L227), feeding `DynamicDownloadConcurrencyLimiter`. With free slots permanently 0 the cb inside `do()` never gets popped from the queue.

The cb is what reaches `downloadDependencyAsync` → `downloadURL` ([InstallManager.ts:5749](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/extensions/mod_management/InstallManager.ts#L5749)) → `emit("start-download", ...)`. `initDownload` has exactly one dispatch site in the codebase - [IPCDownloadAdapter.ts:567](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/IPCDownloadAdapter.ts#L567), inside `#handleStartDownload`, only reached via the `start-download` listener at [IPCDownloadAdapter:133](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/IPCDownloadAdapter.ts#L133). So for any dep routed through `queueDownload` / `resumeDownload` ([InstallManager:6491](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/extensions/mod_management/InstallManager.ts#L6491), [InstallManager:6617](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/extensions/mod_management/InstallManager.ts#L6617)) no Redux download record is created and the InstallDriver state-watcher in collections doesn't have anything to flip to `"downloading"`. Deps stay at the initial pending tracking state.

Confirmed against a real run - log goes silent right at "Set initial allowed phase" ([InstallManager:7050](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/extensions/mod_management/InstallManager.ts#L7050)) with 76 unfulfilled deps queued and no `"starting download"` lines after:

```
13:01:07.495 [DEBG] determined unfulfilled dependencies {"count":76,"errors":0}
13:01:07.496 [INFO] Set initial allowed phase {"sourceModId":"latest","allowedPhase":0}
<silence>
```

Cancellation paths don't unstick it: `queuedDownloads.push` lives inside the cb so `clearQueued` finds nothing to act on, and `reset-dependency-installs` swaps in a fresh limiter while the old queue entries stay pinned by the awaiting Promise chain. (`reset-dependency-installs` is a hack that is probably not needed for `2.0` anyway, and should be removed)

The reason this isn't caught in CI: [InstallManager.test.ts](https://github.com/nexus-mods/vortex/blob/674e72f72/src/renderer/src/extensions/mod_management/InstallManager.test.ts) is bookkeeping-only (404 lines covering phase-state shapes, no actual install flow), and [extensions/collections](https://github.com/nexus-mods/vortex/tree/674e72f72/extensions/collections) only has `bsdiff.test.ts` (for now).

Suggested fix to land alongside this cleanup:
drop `DynamicDownloadConcurrencyLimiter` + `getDownloadFreeSlots` and let `queueDownload` / `resumeDownload` call their cbs directly.

IDCs (approved):
Awesome, works great now!

PR #22981, opened 2026-05-04
Upgrade pnpm from 10.33.0 to 11.0.4

erri120:
Closes APP-414.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -39,6 +39,9 @@ jobs:
         with:
           cache: true
 
+      - name: Configure node-gyp for Electron
+        run: node scripts/setup-electron-env.mjs
+
       - name: Configure node-gyp for VS 2022
         if: runner.os == 'Windows'
         shell: pwsh
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,9 @@ jobs:
         with:
           cache: true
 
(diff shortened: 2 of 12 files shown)


erri120:
No idea why Windows CI suddenly doesn't want to compile `crash-dumps` anymore...

Aragas:
Based on Claude, the removal of .npmrc forces now pnpm to fetch node headers instead of electron headers
>The smallest safe change is to keep a minimal .npmrc with just disturl, target, runtime, arch, target_arch and let setup-electron-env.mjs only handle local-dev convenience.

erri120:
> Based on Claude, the removal of .npmrc forces now pnpm to fetch node headers instead of electron headers
> 
> > The smallest safe change is to keep a minimal .npmrc with just disturl, target, runtime, arch, target_arch and let setup-electron-env.mjs only handle local-dev convenience.

I've added a script to load environment variables. We can't continue using `.npmrc` as both PNPM 11 and NPM 11 are deprecated foreign values from it and it will break with the next version.

Aragas:
If we drop `crash-dumps`, does Windows build it? Because the workarounds look pretty bad, perhaps we _could_ just remove it?

erri120:
I'll wait with upgrading PNPM from 10 to 11 until @halgari finished his native to typescript conversion project. We have no urgency to upgrade to PNPM 11 for now.

PR #23015, opened 2026-05-06
[QA-92][QA-93] Header badges tests

Marcin-NM:
Pure login + header assertion — covers free user must NOT see the Premium indicator and premium user must see it. Adds a `data-testid` to the PremiumIndicator Typography so the e2e test can reliably distinguish the header indicator from other "Premium" text on the page (DashboardBanner, LoginIcon).

Also adds `premiumUser` to helpers/users.ts mirroring `freeUser`.

### DIFF
--- a/packages/e2e/helpers/login.ts
+++ b/packages/e2e/helpers/login.ts
@@ -65,7 +65,7 @@ export async function loginToNexus(
       await expect(nexusLoginPage.authLoginHeading).toBeVisible();
     });
 
-    await test.step("Login with FreeUser credentials", async () => {
+    await test.step("Login with Nexus Mods credentials", async () => {
       if (authPage === null) {
         throw new Error("Auth page was not available for login.");
       }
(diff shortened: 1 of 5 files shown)


IDCs (changes requested):
This is a nitpick really - but within `loginToNexus` we're registering a step `Login with FreeUser credentials`. Given that we're forwarding premium users via this function as well now, it's the ideal moment to change that to `Login with Nexus Mods credentials` or something similar.

(Will be forgotten otherwise)

Marcin-NM:
> This is a nitpick really - but within `loginToNexus` we're registering a step `Login with FreeUser credentials`. Given that we're forwarding premium users via this function as well now, it's the ideal moment to change that to `Login with Nexus Mods credentials` or something similar.
> 
> (Will be forgotten otherwise)

Great catch! Done in: https://github.com/Nexus-Mods/Vortex/pull/23015/commits/175ffeedcbdd84b44a7756e4fa207e818ed6a4e0

IDCs (approved):

PR #23032, opened 2026-05-06
Fix exe-version build setup and add CI guard

halgari:
## Summary

- Add tsdown/tsconfig build setup to `packages/exe-version` so it produces proper CJS/ESM output in `dist/` instead of pointing `main` at raw `.ts` source
- Add a vitest CI test (`scripts/check-package-build-setup.test.ts`) that scans every package under `packages/` and asserts any package with a runtime entry point has a build script and doesn't point to `.ts` source

Fixes APP-422


scripts/check-package-build-setup.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      });
+    }
+  }
+
+  return results;
+}
+
+function hasRuntimeEntryPoint(pkg: Record<string, unknown>): boolean {
+  return pkg.main !== undefined || pkg.module !== undefined;
+}
```
erri120, line 46: The `exports` field takes priority in resolution for Node.js, tsc, webpack, rolldown, and most other tools.

erri120 (approved):

PR #23033, opened 2026-05-06
ran the formatter on the entire branch

IDCs:
no runtime changes - this is literally so we don't trip over ourselves when trying to cherry pick into master


Sewer56 (approved):
If you say it's good, then good with me.

PR #23034, opened 2026-05-06
Replace crash-dump native addon with Electron crashReporter

halgari:
## Summary

- Removes the `crash-dump` native C++ addon (~200 lines, Windows-specific) that installed a Structured Exception Handler to write minidumps
- Replaces it with Electron's built-in `crashReporter` API (Crashpad-based, cross-platform)
- Uses `app.setPath("crashDumps", ...)` to direct crash dumps to the existing `temp/dumps` directory
- Removes the `CRASH_REPORTING` A/B test env var from main process and MainWindow (no longer two competing implementations)
- Cleans up `crash-dump` from `package.json`, `pnpm-workspace.yaml`, `pnpm-lock.yaml`, `flatpak/generated-sources.json`, and dependency report

Resolves APP-402. Unblocks APP-414 (pnpm upgrade).

## Notes

- Electron's Crashpad writes dumps in its own format/directory structure within the `crashDumps` path, which may differ from the old flat `crash-{process}-{timestamp}.dmp` naming. The feedback extension's `findCrashDumps()` searches for `.dmp` files in the top-level dumps directory — this may need to be updated to also search Crashpad subdirectories (e.g. `pending/`, `completed/`).
- The `.dmp.log` companion files that the old crash-dump wrote (used by `recognisedError()` in the feedback extension) will no longer be generated. The existing code handles missing `.log` files gracefully via catch clauses.

### DIFF
--- a/etc/Dependency Report.md
+++ b/etc/Dependency Report.md
@@ -36,7 +36,7 @@ This is a list of all modules leaked by Vortex to extensions. Any module listed 
 | commander                               | 4.1.1                           |
 | content-disposition                     | 0.5.4                           |
 | content-type                            | 1.0.5                           |
-| crash-dump                              | 2.0.2                           |
+
 | d3                                      | 5.16.0                          |
 | date-fns                                | 2.30.0                          |
 | dayjs                                   | 1.11.20                         |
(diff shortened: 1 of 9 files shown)


Sewer56 (approved):
Looks good to me.
Wow- I didn't think I'd see SEH.

Should be fine, as we've been knocking out native deps.

PR #23049, opened 2026-05-07
coalesce persistence writes to reduce LevelDB stalls

### LINKED ISSUE #23030: Vortex will only start once with v2.0.0
> **Important:** Tickets that do not provide sufficient information may be closed without further notice. Context is everything when diagnosing bugs — do not assume that a screenshot alone is enough for us to identify the problem. Logs and reproduction steps are extremely important and, unless the issue is immediately obvious, we cannot assist without them. Please attach all `.log` files from `%appdata%/vortex`.

**Your Nexus Mods username**
IskanderNexus

**Describe the bug**
After closing the application it will not start again

**To Reproduce**
I updated Vortex as i got the Notification. After that i looked around a bit and changed profile to another game.
The Switch Back to Skyrim was working. I closed Vortex and wanted to restart. Didn't work. I had to kill processes via Taskmanager first. After that i could start again, but the profile was back to Witcher3 ... ? 
I guess something is off when closing the application.

**Expected behavior**
Well ... run again ?
**Logs**
Please attach all `.log` files from `

`. Without logs, we are unlikely to be able to help.

**Screenshots**
does not make sense, as it is not starting, after closing in once

 - OS:  Windows 11
 - Vortex Version: [e.g. 1.16.7]
 - Game: [e.g. Skyrim]

**Additional context**

> **Note:** If your issue is specific to a community extension or a game supported by a community extension, please report it on the extension's page on Nexus Mods instead. This tracker is for Vortex core and officially supported functionality only.

### LINKED ISSUE #23052: Vortex deletes huge chunks of nexus collections. Bulk Missing Meta Data. 100 prompts asking if mods should be reinstalled. Crashes and vortex asking for permission to edit the json file.
This has happened on 4 separate occasions with several collections for Skyrim SE (Gate to Sovngarde, Path to Skuldafn, Constellations 2.0, GTS again).

Collection finishes installing, deploys and sorts plugins. I try to close vortex so I can restart some time later and install some more mods. Vortex won't open. After a few tries I have 6 instances of vortex in task manager. Close them from there, and open vortex again.

I see roughly half of the mods, some 1000 of them, are disabled, missing meta data, and deleted. Collections that I just finished installing is incomplete. I resume, only to be prompted with 1000 prompts asking me if I want to reinstall the mods in question. Several clicks later, I see vortex screaming at me that it can't modify disk files (on drive D, not C or program files) and that it wants me to either quit or retry. Click retry 25 times to be asked again and again. Then be told to give permission by windows. Only for vortex to crash and when re-openning to see that no mods has been installed and half the collection is still missing.

IDCs:
The persist queue used to call setItem once per Redux diff entry, and each setItem issued ~3 DuckDB statements (SELECT to check existence, then UPDATE or INSERT). The SELECT was a read; the UPDATE or INSERT produced one LevelDB Write via the Sink's batch.commit. A 1000-op Redux diff (heavy plugin sort, collection install, bulk rule edit) therefore went through ~2000 DuckDB statements and ~1000 sequential LevelDB Writes through LevelDB's single write thread - long enough to provoke force-quits.

Three changes:

- Drop the SELECT-then-branch in setItem. Raw-mode level_pivot tables have no UNIQUE constraint and the insert Sink calls batch.put unconditionally, so plain INSERT already overwrites. One statement per setItem instead of three. Pinned upstream by the level_pivot upsert tests on the [Nexus-Mods/duckdb-level-pivot fix/app-429 branch](https://github.com/Nexus-Mods/duckdb-level-pivot/pull/2).
- Add optional bulkSetItem / bulkRemoveItem to IPersistor and a matching SubPersistor delegation that prefixes the hive. LevelPersist implements them as a single multi-row INSERT or DELETE ... WHERE key IN (...).
- Rewrite ReduxPersistorIPC.processOperations to group consecutive same-type ops into runs and dispatch each run through the bulk path, chunked at 256 rows. A 1000-op diff is now ~4 multi-row statements (~4 LevelDB Writes) total instead of ~2000 statements / ~1000 Writes. The 256 chunk cap preserves recovery granularity: any single Write failure can lose at most 256 ops of progress.

Adds breadcrumb logging in LevelPersist: any successful write that exceeds 250 ms is logged at warn level, and setting VORTEX_TRACE_DB_WRITES=1 in the environment turns on enter/exit pairs at debug level for every persistence write so an indefinite hang is identifiable from vortex.log. Documented in AGENTS-DEBUGGING.md.

Adds vitest coverage for the bulk methods, run-grouping, chunking, fallback path when bulk methods are absent, transaction-state correctness (including stale-flag handling on COMMIT/ROLLBACK failures), and a force-quit-style atomicity test that injects a mid-transaction write failure and asserts the queue rolls back rather than continuing.

part of https://linear.app/nexus-mods/issue/APP-429

### DIFF
--- a/AGENTS-DEBUGGING.md
+++ b/AGENTS-DEBUGGING.md
@@ -7,3 +7,16 @@
 
 > sewer note: This will be replaced with better .vscode config down the road.
 > am just keeping this as a leftover note for now.
+
+## Diagnostic environment variables
+
+- `VORTEX_TRACE_DB_WRITES=1` - turns on per-write breadcrumb logging in
+  `LevelPersist`. Every persistence write (`setItem`, `removeItem`,
+  `bulkSetItem`, `bulkRemoveItem`) emits a `level_pivot Write enter`
+  line at debug level before the call and a `level_pivot Write exit`
+  line after, with `method`, `alias`, `count`, and `elapsedMs`. Useful
+  for diagnosing indefinite shutdown hangs in the persist queue (the
+  last enter line with no matching exit pinpoints the wedged call).
+  Off by default. With the env var unset, the same code path still
+  emits a `level_pivot slow Write` warning when a single write exceeded
+  250 ms (`SLOW_WRITE_THRESHOLD_MS` in `LevelPersist.ts`).
--- a/src/main/src/Application.ts
+++ b/src/main/src/Application.ts
@@ -804,8 +804,9 @@ class Application {
       );
     }
 
-    // Wrap all operations in a single transaction to avoid concurrent
-    // BEGIN TRANSACTION calls from individual setItem/removeItem calls.
+    // Wrap all operations in a single transaction so the import is atomic;
+    // a partial backup restore on failure would leave state in a confusing
+    // mixed-version condition.
     await persistor.beginTransaction();
     try {
       for (const [hive, hiveData] of Object.entries(backupData)) {
--- a/src/main/src/store/LevelPersist.test.ts
+++ b/src/main/src/store/LevelPersist.test.ts
@@ -1,0 +1,411 @@
+import type { DuckDBConnection } from "@duckdb/node-api";
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+vi.mock("../logging", () => ({ log: vi.fn() }));
+
+import { log } from "../logging";
+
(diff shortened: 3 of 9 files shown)


Aragas (approved):

IskanderNexus1:
thx , really appreciated !!

halgari (approved):
Looks good

PR #23054, opened 2026-05-07
Consolidate build scripts and outputs

erri120:
Closes APP-427.


IDCs (changes requested):
I can see that the extra resources/assets (.NET + VC runtimes) are now located in `./temp/`. Both the `package.yml` file and the NSIS installer `installer.nsh` should be updated to reflect that (they're still looking for those files in `./build/`)

IDCs:
What about `package.yml` ? 😋

```
- name: Download Runtime Dependencies
        shell: pwsh
        run: |
          New-Item -ItemType Directory -Force -Path "src\\main\\build"
          Invoke-WebRequest -Uri "https://aka.ms/vs/17/release/vc_redist.x64.exe" -OutFile "src\\main\\build\\VC_redist.x64.exe"
          Invoke-WebRequest -Uri "https://aka.ms/dotnet/9.0/windowsdesktop-runtime-win-x64.exe" -OutFile "src\\main\\build\\windowsdesktop-runtime-win-x64.exe"
```

IDCs (approved):

PR #23060, opened 2026-05-07
use App token for cherry-pick workflow so created PRs trigger CI

IDCs:
This needs to go in before any other `v2.0` -> `master` cherry-pick PRs are created

fixes https://linear.app/nexus-mods/issue/APP-411

### DIFF
--- a/.github/workflows/cherry-pick.yml
+++ b/.github/workflows/cherry-pick.yml
@@ -15,18 +15,32 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Generate GitHub App token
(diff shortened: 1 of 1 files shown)


Sewer56:
Note that before merging, someone will need to make a GitHub App, and wire up the 2 secrets.

`ID` and `PRIVATE_KEY`.

Needs to be someone higher up with perms to edit repo settings.

IDCs:
Thanks @Sewer56 - both should already be present. `main.yml` is using the same pattern.

Sewer56 (approved):

PR #23067, opened 2026-05-07
Discovery Data Path Bug

Aragas:
Suppress discovery error if path is not present, silently throw instead

Fixes fingerprints 4b1decf8
Closes APP-439


erri120 (approved):

Aragas:
Incorrect fix, another PR coming

Issue #14, opened 2017-02-28
small memory leak


TanninOne:
Right now we're leaking approx. 500kb every time an xml fomod is installed.
The leaked memory is c# objects containing functions that are passed from c# to js (startDialog call).

As far as I can tell all references to these functions are released on the js side so this may be a bug in the edge library and may disappear in a future release.
For now, the amount of leaked memory is not that significant so this is low priority.

TanninOne:
This is no longer relevant because we no longer use edge to marshal objects between js and C#

PR #23127, opened 2026-05-12
[QA-176] SDV manual mod download e2e

Marcin-NM:
E2E for QA-176 (SDV [#14.7] manual mod download). Also satisfies the generic QA-109 flow.

- New test `mods-manual.spec.ts` (free + premium tiers): website Manual download → optional slow-download interstitial → file download → Vortex Install From File → mod in library.
- New `__VORTEX_TEST_INSTALL_FILE_PATH__` test hook in `ExtensionManager.selectFile`, mirroring the existing `__VORTEX_TEST_GAME_PATH__` pattern. Bypasses the native file picker in tests; production behaviour unchanged.

Verified locally: both tiers pass.


src/renderer/src/ExtensionManager.ts
```diff
@@ -1846,6 +2011,15 @@ class ExtensionManager {
...
 
   private selectFile(options: IOpenOptions): PromiseBB<string> {
+    // Test hook: e2e tests can't drive native file pickers, so they preload
+    // the path via globalThis.__VORTEX_TEST_INSTALL_FILE_PATH__ and bypass
+    // the dialog. Mirrors __VORTEX_TEST_GAME_PATH__ in gamemode_management.
+    const testPath = (global as any).__VORTEX_TEST_INSTALL_FILE_PATH__;
+    if (testPath !== undefined) {
+      delete (global as any).__VORTEX_TEST_INSTALL_FILE_PATH__;
+      return PromiseBB.resolve(testPath);
+    }
```
erri120: A cleaner solution would be to override the open dialog API instead of adding more test-only globals and polluting the real code with testing code.

Marcin-NM: Done: reverted the source hook, stub dialog.showOpenDialog from the test side instead (same pattern as shell.openExternal). Both tiers verified locally.

IDCs (approved):

PR #23136, opened 2026-05-12
fixed inability to apply group rules when filter is defined

### LINKED ISSUE #23124: when specifying search criteria in the mods tab manage rules icon, "after all" and "before all" do not work
**Your Nexus Mods username**
fractalbase

**Describe the bug**
without specifying the search field, "before all" and "after all" links work.
with specifying the search field, "before all" and "after all" links do not work.
the program requires individual rule management with a search field specification.

Windows 11 Pro
Vortex 2.0.0
Skyrim Special Edition (and AE)

IDCs:
The conflict editor had two filter predicates that disagreed: the display
(applyFilter) matched a row if either source or referenced mod name contained
the filter, but applyGroupRule only checked the referenced mod name. When the
user filtered by the source mod's name to narrow to one group, the group was
still rendered with all its conflict rows, but "Before All" / "After All"
silently did nothing because the apply path computed an empty ref set.

Reuse applyFilter inside applyGroupRule so the rule applies to exactly the
rows the user sees, regardless of which side of the conflict the filter
matched.

(actual code is in commit d07ecc41a54f85a43d3fe8979a7f1fbdafb8ea2b)

Fixes APP-451
Fixes https://github.com/Nexus-Mods/Vortex/issues/23124

### DIFF
--- a/extensions/mod-dependency-manager/src/views/ConflictEditor.tsx
+++ b/extensions/mod-dependency-manager/src/views/ConflictEditor.tsx
@@ -629,24 +629,14 @@ class ConflictEditor extends ComponentEx<IProps, IComponentState> {
     }
 
     let newRules = { ...rules };
-    const hasAppliedFilters = hideResolved || !!filterValue;
-    const refIds = hasAppliedFilters
-      ? Object.keys(rules[modId]).filter((refId) => {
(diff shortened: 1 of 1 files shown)


Aragas (approved):
Thanks for the 2 commit approach, made the review way easier

PR #23137, opened 2026-05-12
Use pnpm deploy instead of brittle DIY script

erri120:
Massively simplifies how we publish and package Vortex. Also fixes an issue with copying assets.

Closes APP-423.


package.json
```diff
@@ -15,13 +15,11 @@
...
     "prepare": "husky",
     "preinstall": "npx only-allow pnpm && node ./scripts/create-env-file.mjs",
-    "start": "pnpm -F @vortex/main run start",
+    "start": "pnpm nx run @vortex/main:start",
     "build": "pnpm nx run-many -t build lint typecheck && pnpm run assets",
-    "dist": "pnpm run build && pnpm --filter \"./src/*\" -r run dist",
     "assets": "npx tsx scripts/download-duckdb-extensions.ts && node ./scripts/dependency-report.mjs",
-    "package": "pnpm run dist && pnpm -F @vortex/main run package",
-    "package:nosign": "pnpm run dist && pnpm -F @vortex/main run package:nosign",
-    "package:local": "pnpm -F @vortex/main run prepare:win && pnpm run package:nosign",
```
insomnious, line 24: Do we not need package:local now? It made it easier running from root to download the assets needed to package.

erri120: I'll remove `prepare:win` so there won't be a need for it anymore.

insomnious: So we won't need to manually download those redist files at all ?

erri120: Nope.

IDCs (changes requested):
We should trim the `Download Runtime Dependencies` step in `package.yml` - it's still calling the now-defunct `prepare:win`.

Additionally `tsconfig.node.json` still includes `InstallAssets.mjs` which is now removed.

NGL, I'm a bit concerned about removing `cross-env` from the start scripts. If I read the code correctly, the bundle is dev-baked and should be fine during compile-time checks. But at runtime `NODE_ENV` will be set to `undefined` unless the developer sets it in his shell/env.

This means that dynamic extensions that attempt to change their logic based on `NODE_ENV` could potentially generate unwanted side-effects. I can't recall which community extensions listen to this environment variable (probably just a few of them use it)

But taking our bundled extensions into consideration - `extensions/test-setup/src/index.ts:85` will now register the uninstall test on dev environments and run unnecessary registry operations.

IDCs (dismissed):

IDCs (approved):

PR #23158, opened 2026-05-13
fix Tailwind source detection for renderer

IDCs:
Tailwind v4 auto-detects sources from the CLI's cwd. After 095ed0f95 moved the output from src/main/build/assets/css/ to src/stylesheets/dist/, the scan no longer reached src/renderer/, so utility classes used in the Spine and other UIv2 components were stripped from tailwind-v4.css and the layout rendered unstyled.

Add an explicit `@source` pointing at the renderer so detection no longer depends on where the output happens to land.

### DIFF
--- a/src/stylesheets/tailwind-v4.css
+++ b/src/stylesheets/tailwind-v4.css
@@ -12,6 +12,9 @@
 
 @import "tailwindcss";
 
+/* Explicit source paths so detection doesn't depend on the CLI's cwd. */
+@source "../renderer/src/**/*.{ts,tsx,html}";
+
 /* Import design system */
 @import "./ui/theme";
 @import "./ui/elements";
--- a/src/stylesheets/package.json
+++ b/src/stylesheets/package.json
@@ -32,7 +32,8 @@
         "cache": true,
         "inputs": [
           "default",
-          "{projectRoot}/**/*.css"
+          "{projectRoot}/**/*.css",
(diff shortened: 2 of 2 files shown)


src/stylesheets/tailwind-v4.css
```diff
@@ -12,6 +12,9 @@
 
 @import "tailwindcss";
 
+/* Explicit source paths so detection doesn't depend on the CLI's cwd. */
+@source "../renderer/src/**/*.{ts,tsx,html}";
```
erri120, line 16: I went back on forth whether tailwind CLI actually did this. I tried to proof it didn't scan our files implicitly because the CLI finishes within 30ms which seemed very sus.

You also need to update `src/stylesheets/package.json`:

```diff
diff --git a/src/stylesheets/package.json b/src/stylesheets/package.json
index e96df8a2a..e34084111 100644
--- a/src/stylesheets/package.json
+++ b/src/stylesheets/package.json
@@ -32,7 +32,8 @@
         "cache": true,
         "inputs": [
           "default",
-          "{projectRoot}/**/*.css"
+          "{projectRoot}/**/*.css",
+          "{workspaceRoot}/src/renderer/src/**/*.{ts,tsx,html}"
         ],
         "outputs": [
           "{projectRoot}/dist/tailwind-v4.css"
```

erri120 (approved):

PR #23164, opened 2026-05-13
feat(ci): add nightly scheduled builds to package workflow

Sewer56:
Closes [LAZ-221](https://linear.app/nexus-mods/issue/LAZ-221/run-packaging-workflow-nightly)

## Summary

Adds scheduled nightly builds to the existing `Package` workflow.

Produces a signed, auto-versioned artifact every day at 19:00 UTC.
Also adds a `nightly` boolean input to `workflow_dispatch` for on-demand nightly-style builds.

Exact details etc. are up for discussion. I just went with 19:00 for the time being because it's after most people clock off.

## Notes

- Versioning: `<package.json version>-nightly.<yyyyMMdd>` (e.g. `1.0.0-nightly.20260710`), fetched at runtime from reading `package.json`.
    - For now this always is `1.0.0`, which is redundant. This is something we'll work out in the future. 
    - We want next version to be there, i.e. our `package.json` always contains next version.
    - But exact deployment steps to facilitate that were not discussed; could just be a commit we make after we ship.
- Artifacts: always uploaded with 2-day retention. Release and staging-release steps are skipped.
- Codesigning: always enabled for nightlies regardless of the `use-codesigning` input.
- Non-nightly `workflow_dispatch` behaviour is unchanged.
- We now test `Package` workflow as part of CI when it changes.

Edit Note: Test filters lifted out, will be future PR.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,4 +1,4 @@
-name: Nightly Build
+name: Nightly Package
 
 on:
   schedule:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -5,6 +5,9 @@ on:
     - cron: "0 19 * * *" # Nightly at 19:00 UTC
   workflow_dispatch:
 
+permissions:
+  actions: write
+
 jobs:
   trigger:
     runs-on: ubuntu-latest
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,23 @@
+name: Nightly Build
+
+on:
+  schedule:
+    - cron: "0 19 * * *" # Nightly at 19:00 UTC
+  workflow_dispatch:
+
+jobs:
+  trigger:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            await github.rest.actions.createWorkflowDispatch({
+              owner: context.repo.owner,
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,23 @@
+name: Nightly Build
```
insomnious:
```suggestion
name: Nightly Package
```

insomnious:
Where do we see the list of nightly actions run and their artifacts? They don't seem to be under nightly.yml, but do appear under the package.yml which they invoke. 

In future, we want version string being 'next version'. This mechanism doesn't currently exist as version is manually added as an input when we run package.yml

erri120:
Why do we need a complex filter on the package workflow? It just needs to run nightly and manually.

Sewer56:
To find exact PRs when packaging breaks. (As much as possible)
Prevent merging PRs that break packaging; like yesterday morning.

Rather than finding a change broke it at the end of the day, and someone having to then do a binary search of commits, running a 15 min packaging workflow on each commit during the search to find the culprit.

insomnious:
Can't we just get the nightly up and running (just running package once a day, make sure artifacts have an appropriate nightly/date/something string), then we can iterate in future if we hit a speed bump.

Sewer56:
Well, this does that.

insomnious (dismissed):

insomnious (approved):

PR #23165, opened 2026-05-13
Package inline issue fixes

Aragas:
# Fix: packaged installer crashes at startup with "Cannot find module" errors

## Problem

After the recent NX + `pnpm deploy` migration, the packaged installer launches but the main process (and then the renderer) crash at startup with a cascading sequence of `Cannot find module` errors:

```
Error: Cannot find module '@babel/runtime/helpers/typeof'
  required by app.asar/node_modules/i18next/dist/cjs/i18next.js

Error: Cannot find module 'source-map'
  required by app.asar/node_modules/source-map-support/source-map-support.js

SyntaxError: The requested module 'is-stream' does not provide an export named 'isReadableStream'

Error: Cannot find module 'redux-freeze'
  required by app.asar/renderer.js
```

## Root cause

`pnpm deploy` (with `forceLegacyDeploy: true`) produces an **isolated** `node_modules` layout where direct deps are top-level symlinks and transitives only live under `.pnpm/<shard>/node_modules/`. electron-builder 24.x doesn't traverse the `.pnpm` virtual store, so 98+ transitive packages reachable from main's runtime imports never make it into `app.asar`.

A separate but related issue: the renderer is webpack-bundled with `nodeExternals({ allowlist: [/@vortex\/shared/] })`, which marks **every** dep as an external runtime `require()`. Those externals need to live in `@vortex/main`'s production dependencies (since `renderer.js` lives inside `app.asar`, governed by main's `package.json`).

## Approaches considered

| # | Approach | Verdict |
|---|---|---|
| 1 | Declare missing transitives explicitly in main's `dependencies` | Tried with `@babel/runtime` — audit revealed **98 missing transitives** (`source-map`, `js-yaml`, `keyv`, `tldts`, `chokidar`, `graceful-fs`, …). Whack-a-mole with new failures every dep update. Rejected. |
| 2 | Post-deploy in-house hoister (walk `.pnpm/`, create top-level junctions) | Got us further but hit a hard wall on **version conflicts**: three versions of `is-stream` in the tree, only one can sit at the top, picked the wrong one. Flat hoisting can't faithfully represent multi-version graphs. Rejected. |
| 3 | **Post-deploy `pnpm install --node-linker=hoisted`** — let pnpm do the hoist+nest itself | Adopted. pnpm correctly puts compatible versions at the top and nests conflicts under their consumer (e.g. top-level `is-stream@4` for `got`, `execa/node_modules/is-stream@1` for `execa`). |

## Solution

After `pnpm deploy` produces the (isolated) dist, a follow-up `pnpm install --prod --node-linker=hoisted` re-lays out `dist/node_modules` in npm-style hoisted-with-nesting. electron-builder's `includeSubNodeModules` is flipped to `true` so the nested copies (for conflict versions) also make it into the asar.

For the install to succeed in `dist/` (which is outside the workspace), `prepare-dist-package.mjs`:
- Resolves `catalog:` deps via the **raw catalog spec** from the root `pnpm-workspace.yaml` (so git-URL catalog entries like `bbcode-to-react: git+https://...` are preserved verbatim — `pnpm deploy`'s resolution turns them into internal version numbers that don't exist on npm).
- Resolves `workspace:*` deps to `file:<absolute-path>` pointing at the workspace **source** (parsed from the workspace's `packages:` glob), not at the about-to-be-wiped pnpm-deploy copy under `.pnpm/`.
- Writes a shadow `dist/pnpm-workspace.yaml` carrying just `overrides` and `allowBuilds` from root (so transitive version overrides and native-module build-script approvals survive). The shadow exists so pnpm stops walking up at `dist/` and doesn't pick up the root's `catalogMode: strict` / `forceLegacyDeploy` settings.
- Wipes `dist/node_modules` so the follow-up install rebuilds it cleanly.

### [`src/main/prepare-dist-package.mjs`](src/main/prepare-dist-package.mjs)
Rewritten to:
- Read the **source** `src/main/package.json` (preserves `catalog:`/`workspace:*` protocols intact).
- Parse the root `pnpm-workspace.yaml` for the `catalog:` block and `packages:` glob.
- Resolve `catalog:` → raw catalog spec, `workspace:*` → `file:<absolute-path>`.
- Write the resolved `package.json` to both `dist/package.json` and `dist/build/package.json`.
- Write a shadow `dist/pnpm-workspace.yaml` with `overrides` + `allowBuilds`.
- Wipe `dist/node_modules`.

### [`src/main/package.json`](src/main/package.json)
- `publish` script appends `&& pnpm install --dir=./dist --prod --node-linker=hoisted` after the existing `rimraf` + `pnpm deploy` + `prepare-dist-package.mjs` chain.
- Moved `@welldone-software/why-did-you-render` and `redux-freeze` from `devDependencies` → `dependencies` (renderer externals require them at runtime).
- Added `csstype` to `dependencies` (renderer external, previously not declared anywhere in main).

### [`src/main/electron-builder.config.json`](src/main/electron-builder.config.json)
- `includeSubNodeModules: false` → `true`. Required so nested `node_modules/<pkg>/node_modules/<dep>` directories (which pnpm/npm produce on version conflicts) are bundled into the asar.

### [`src/shared/tsconfig.json`](src/shared/tsconfig.json)
- `"include": ["**/*"]` → `"include": ["src/**/*"]`. Drive-by fix for a pre-existing nx parallelism race that surfaced when this PR invalidated the build cache: `tsdown` rewrites content-hashed `dist/*` files while `tsc` is reading them, producing intermittent `TS6053: File not found` errors. Matches the pattern used by `src/main`, `src/preload`, and `packages/adaptor-api`.

## Verification

Locally, after the changes:

- `pnpm install --prod --node-linker=hoisted` in `dist/` completes in ~47s with all native modules (winapi-bindings, xxhash-addon, leveldown, protobufjs, drivelist) successfully building via `allowBuilds`.
- `is-stream@4` sits at top level (satisfies `isReadableStream` consumers via `got`), `is-stream@1` nested under `execa` (satisfies execa's `^1.1.0`).
- All previously-missing transitives — `@babel/runtime`, `source-map`, `chokidar`, `keyv`, `js-yaml`, `graceful-fs`, `colors`, `tldts`, etc. — resolve at top level.
- Workspace deps `@nexusmods/adaptor-api` and `exe-version` install from their workspace source via `file:`.
- `pnpm nx run-many -t build lint typecheck` clean.

### DIFF
--- a/docs/install-instructions/windows.md
+++ b/docs/install-instructions/windows.md
@@ -67,6 +67,35 @@ Some native build pieces need CMake during local builds.
 winget install --id Microsoft.DotNet.SDK.9 -e
 ```
 
+## Enable Windows Settings
+
+The following Windows settings must be enabled. Both require an
+**Administrator** PowerShell session and a sign-out (or reboot) to take effect.
+
+### Enable Long Paths
+
+Vortex and its native modules produce paths longer than the legacy 260-character
+`MAX_PATH` limit. Enable long path support in Windows and Git:
+
+```powershell
+New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" `
+    -Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
+git config --system core.longpaths true
+```
+
+### Enable Windows Developer Mode
+
+Developer Mode allows symlink creation without elevation, which `pnpm` and some
+native build steps rely on.
+
+```powershell
+New-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock" `
+    -Name "AllowDevelopmentWithoutDevLicense" -Value 1 -PropertyType DWORD -Force
+```
+
+Alternatively, enable it via **Settings → System → For developers → Developer
+Mode**.
+
 ## Verify Install
 
 ```powershell
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3981,6 +3981,9 @@ importers:
       '@types/uuid':
         specifier: 'catalog:'
         version: 3.4.13
+      '@welldone-software/why-did-you-render':
+        specifier: 'catalog:'
+        version: 10.0.1(react@16.14.0)
       bbcode-to-react:
         specifier: 'catalog:'
         version: https://codeload.github.com/TanninOne/bbcode-to-react/tar.gz/c67356006470e5066ea447e04a3968dca367339d(react@16.14.0)
@@ -4005,6 +4008,9 @@ importers:
       content-type:
         specifier: 'catalog:'
         version: 1.0.5
+      csstype:
+        specifier: 'catalog:'
+        version: 3.2.3
       d3:
         specifier: 'catalog:'
         version: 5.16.0
@@ -4194,6 +4200,9 @@ importers:
       redux-batched-actions:
         specifier: 'catalog:'
         version: 0.5.0(redux@4.2.1)
+      redux-freeze:
+        specifier: 'catalog:'
+        version: 0.1.7
       redux-thunk:
         specifier: 'catalog:'
         version: 2.4.2(redux@4.2.1)
@@ -4375,9 +4384,6 @@ importers:
       '@vortex/shared':
         specifier: workspace:*
         version: link:../shared
-      '@welldone-software/why-did-you-render':
-        specifier: 'catalog:'
-        version: 10.0.1(react@16.14.0)
       cross-env:
         specifier: 'catalog:'
         version: 10.1.0
@@ -4390,9 +4396,6 @@ importers:
       electron-devtools-installer:
         specifier: 'catalog:'
         version: 3.2.1
-      redux-freeze:
-        specifier: 'catalog:'
-        version: 0.1.7
 
   src/preload:
     dependencies:
--- a/src/main/electron-builder.config.json
+++ b/src/main/electron-builder.config.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://github.com/electron-userland/electron-builder/raw/refs/tags/v24.13.3/packages/app-builder-lib/scheme.json",
   "appId": "com.nexusmods.vortex",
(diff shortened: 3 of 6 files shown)


Sewer56:
Fix

docs/install-instructions/windows.md
```diff
@@ -67,6 +67,35 @@ Some native build pieces need CMake during local builds.
...
+The following Windows settings must be enabled. Both require an
+**Administrator** PowerShell session and a sign-out (or reboot) to take effect.
+
+### Enable Long Paths
+
+Vortex and its native modules produce paths longer than the legacy 260-character
+`MAX_PATH` limit. Enable long path support in Windows and Git:
+
+```powershell
+New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" `
```
Sewer56, line 81: I'd need to dig up my earlier comments, but I never got this to work on my end. At least a bit around a month ago. Some of the tooling under the hood still enforced MAX_PATH as 260 chars; irrespective of whether registry was set.

Also; do you know when since we need dev mode? I didn't need it last time I built on Windows.

Aragas: Packaging requires it. That path enforcement did work for me, I think it was the same command

Sewer56: We've yeeted a lot of the native deps, so maybe things changed since. But I remember having to clone to `C:\v` because using the reg tweak didn't do it for me, haha.

It's what the guide says too.

Sewer56 (approved):
That's a lot of vibing, not perfect, but aight; does the job for now.

Unless @erri120 has any ideas for a cleaner approach here. Not specifically my field of expertise, but having had a read, can do for time being.

erri120:
Will put up a PR with a different solution.

Aragas:
> That's a lot of vibing, but aight; does the job for now.
> 
> Unless @erri120 has any ideas for a cleaner approach here. Not specifically my field of expertise, but having had a read, does the job for now.

The idea was to give a reference for a working implementation, but worst case merge it now and then refine it

PR #23172, opened 2026-05-13
Fix broken asar after packaging

erri120:
I'll leave the swearing out of the PR and just say that `electron-builder` doesn't do what it's supposed to do.

The `node_modules` directory created by `pnpm deploy` didn't get copied by `electron-builder` entirely. This PR adds a `afterPack` script to simply overwrite whatever garbage `electron-builder` does with the correct `node_modules` directory.


erri120:
ASAR doesn't support symlinks so this approach is also mot. There's no way around using a hoisted structure when packaging.

erri120:
Using hoisted deployment, should work now.

Aragas (dismissed):

Aragas (approved):

Sewer56 (approved):

Sewer56:
@Aragas Step looks correct, can you check actual built Windows artifact before merging? Should work.

Aragas:
It does work, just confirmed, but we'll need to switch it to the v2.1 branch. I hope that it still will work with pnpm 10, hahaha!

PR #23186, opened 2026-05-14
fix collections bsdiff crash in renderer by using a DOM Worker

IDCs:
Electron renderers do not register V8 with worker support, so `new Worker(...)` from `node:worker_threads` throws "The V8 platform used by this instance of Node does not support creating Workers" the first time a collection with binary patches is installed.

Split the worker into two entry points sharing pure WASM helpers (bsdiffWasm.ts): bsdiffWorker.dom.ts for the renderer (DOM Worker, relies on nodeIntegrationInWorker for fs access; wasm path passed via URL query) and bsdiffWorker.ts kept for the vitest node environment. bsdiff.ts dispatches between them based on whether a DOM Worker global exists.

Adds a regression test that stubs globalThis.Worker and traps any use of node:worker_threads to guard against re-introducing the crash.

fixes APP-461

### DIFF
--- a/extensions/collections/src/util/bsdiff.ts
+++ b/extensions/collections/src/util/bsdiff.ts
@@ -1,60 +1,94 @@
 /**
  * BSDIFF40-compatible binary diff and patch using WASM.
  *
- * The WASM computation runs in a worker thread to avoid blocking the
- * Electron main/renderer thread. A single long-lived worker is spawned
- * on first use and reused for all subsequent calls.
+ * Production (Electron renderer) runs the WASM in a DOM Worker —
+ * Electron's renderer does not support `node:worker_threads`, so the
+ * Web Worker is the only viable off-thread option. Tests (vitest in
+ * node environment) fall back to `worker_threads`. The WASM logic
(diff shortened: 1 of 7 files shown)


extensions/collections/src/util/bsdiff.ts
```diff
@@ -1,60 +1,91 @@
...
+    onError: (cb) =>
+      w.addEventListener("error", (e: ErrorEvent) =>
+        cb(e.error instanceof Error ? e.error : new Error(e.message)),
+      ),
+  };
+}
+
+function spawnNodeWorker(): WorkerHandle {
+  const { path: scriptPath, isTsSource } = resolveWorkerScript("bsdiffWorker");
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
```
IDCs, line 61: Lazy require keeps this branch sync (`await import` would force the whole postToWorker chain to become async and avoids touching worker_threads from the renderer's module-load path, where it isn't needed.

Aragas (dismissed):

Aragas (approved):

PR #23237, opened 2026-05-19
Preparing to publish `@nexusmods/vortex-api` from monorepo

erri120:
Part of LAZ-231.

- renames `vortex-api` to `@nexusmods/vortex-api`
- updates all package and imports
- updates require monkey patching for extensions to look out for both module names
- updates build configs to mark API as external
- use `pnpm version` in CI to set API and main versions
- prepares package workflow for publishing, still requires setup
- removes `typecheck` nx target from projects without a `typecheck` script

@Aragas @insomnious we need to set up NPM provenance publishing, similar to what we did for the FOMOD repo. Not sure what tokens are involved there.


Aragas:
Provenance won't work with the current state.
We need to publish `@nexusmods/vortex-api` at least once either manually or via CI and after that we can supply the GitHub Repo settings so NPM would register GitHub Provenance data.
If you don't want to do another PR for that you can get a token for publishing and publish some v0.0.0 empty package once.

insomnious:
Let me know what you need from me/what secrets to add

Aragas:
More of a curiosity question, can we actually have a local `vortex-api` package that imports and re-exports `@nexusmods/vortex-api` instead of doing monkey patching? And we just make sure that the local package is exposed just like the npm one.

Aragas:
If the monkey patching just this line?
`if (id === "vortex-api" || id === "@nexusmods/vortex-api")`

IDCs (approved):
I get we're now waiting for the NPM package to be present, but PR Looks fine to me so going to approve.

PR #23255, opened 2026-05-20
nx inputs and outputs improvements

erri120:
Cleans up some inputs and outputs for nx targets. Also enables `build` caching for `@vortex/main`.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -8,18 +8,25 @@
   },
   "namedInputs": {
     "default": [
-      "{projectRoot}/package.json",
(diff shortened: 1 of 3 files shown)


IDCs (approved):

Sewer56 (approved):

Sewer56:
I actually did some of these changes locally last night. Unfortunately due to caveats of how the whole caching system works, it wasn't great in caching on CI out of the box; because that was in part keyed by machine.

Also found out they had some sort of cloud service to work around it, but alas.

Was late last night though, would need to pull up docs again.

PR #23270, opened 2026-05-20
CI: add Electron rebuild postinstall controls

Sewer56:
Part of [LAZ-411](https://linear.app/nexus-mods/issue/LAZ-411/optimize-vortex-ci-build-pipelines) . From my background experiments on Monday.

## Summary

Add explicit controls for `@vortex/main` Electron native rebuild during postinstall.

This is a building block for future PRs.

## Change

`src/main/package.json` now runs:

```
"postinstall": "node ./postinstall.mjs"
```

Default behavior remains equivalent to the old postinstall:

```
pnpm exec electron-rebuild
```

Two opt-in CI controls are added:

- `VORTEX_SKIP_ELECTRON_REBUILD=1`
  - skip rebuild completely
  - used by format/package optimization PRs
- `VORTEX_DEFER_ELECTRON_REBUILD=1`
  - skip during install so workflow can run explicit rebuild later
  - used by future PRs to cache native inputs

Wrapper uses `shell: true` on Windows so `pnpm.cmd` resolves correctly.

`postinstall.mjs` is included in `src/main/package.json` `files` for deployed dist packages.


IDCs (dismissed):

erri120:
What exactly is the difference between `VORTEX_SKIP_ELECTRON_REBUILD` and `VORTEX_DEFER_ELECTRON_REBUILD`? Is it just intend or behavior?

Sewer56:
> Is it just intend or behavior?

Yeah; it's intent (logging).
Can merge together, but I had them separated out to make CI logs cleaner.

src/main/postinstall.mjs
```diff
@@ -0,0 +1,23 @@
...
+
+if (process.env.VORTEX_SKIP_ELECTRON_REBUILD === "1") {
+  console.log("Skipping electron-rebuild because VORTEX_SKIP_ELECTRON_REBUILD=1.");
+  process.exit(0);
+}
+
+if (process.env.VORTEX_DEFER_ELECTRON_REBUILD === "1") {
+  console.log("Deferring electron-rebuild because VORTEX_DEFER_ELECTRON_REBUILD=1.");
+  process.exit(0);
+}
```
erri120: Let's consolidate this into one switch.

src/main/postinstall.mjs
```diff
@@ -0,0 +1,23 @@
...
+
+const result = spawnSync("pnpm", ["exec", "electron-rebuild"], {
+  shell: process.platform === "win32",
+  stdio: "inherit",
+});
```
erri120, line 17: Since we have a script, we should use the `@electron/rebuild` library directly to control which modules get re-build. Honestly we might not even need to rebuild any module. I added the rebuild step after removing the node-gyp config values as that was just a complete mess.

`@nexusmods/fomod-installer-native` might be the only module that explicitly needs a rebuild, but even that might work fine with node headers then we don't need to bother with this at all.

erri120 (changes requested):

src/main/postinstall.mjs
Sewer56: Sure, works for me. Just let me update dependent PRs after this.

src/main/postinstall.mjs
Sewer56: We can do this; but I'm not sure if it's really worth it in this context.

All of this work will be cached in dependent PRs; so while we could save a bit by filtering out projects here- in most runs, because of caching, what will happen here would be a no-op. That's why I preserved existing behaviour we had as closely as possible.

*Could you put this piece of feedback into a ticket as a future piece of work?*

I'll be told off if I stray too much from issues in *Vortex Execution Plan* project; since this is supposed to be the focus right now; so it's not something I can afford to do at the moment. Especially if there isn't much notable gain.

-----------

(I technically wasn't supposed to be optimising CI runtime either; but I was just under the impression we were tasked to find whatever can be improved, make an issue on Linear, and go through with it while stuff was sorted out.

I only learned there were existing tickets once @insomnious told me on Tuesday morning :sweat_smile: . I didn't notice issues not grouped to a milestone were on separate tab; instead of an ungrouped set under milestones within a Linear project)

src/main/postinstall.mjs
Sewer56: Sorted. I went through all the other PRs and rebased them one by one to work with this.

PR #23272, opened 2026-05-20
CI: cache Electron native rebuild outputs in main workflow

Sewer56:
Part of [LAZ-411](https://linear.app/nexus-mods/issue/LAZ-411/optimize-vortex-ci-build-pipelines)

Depends on:
- https://github.com/Nexus-Mods/Vortex/pull/23270

Merge that first.

## Summary

Cache Electron native rebuild outputs in `main.yml`.

## Change

`main.yml` now:

1. installs dependencies with `VORTEX_DEFER_ELECTRON_REBUILD=1`
2. restores/caches native build headers:
   - `~/.electron-gyp`
   - `~/.cache/node-gyp`
3. restores/caches Electron native outputs under `node_modules/.pnpm/...`
4. runs one explicit rebuild:

```sh
pnpm --filter @vortex/main exec electron-rebuild
```

Cached modules:

- `@nexusmods/fomod-installer-native`
- `drivelist`
- `leveldown`
- `winapi-bindings`
- `xxhash-addon`
- `@parcel/watcher`

## Measurement

Baseline full-postinstall main runs:

- https://github.com/Sewer56/Vortex/actions/runs/26099098244
- https://github.com/Sewer56/Vortex/actions/runs/26099100086
- https://github.com/Sewer56/Vortex/actions/runs/26099101744

Warmup/verification run:

- https://github.com/Sewer56/Vortex/actions/runs/26099103356

Optimized warm-cache runs:

- https://github.com/Sewer56/Vortex/actions/runs/26099876762
- https://github.com/Sewer56/Vortex/actions/runs/26099878560
- https://github.com/Sewer56/Vortex/actions/runs/26099880447

Results:

| Platform | Metric                            | Baseline median | Optimized median | Delta |
| -------- | --------------------------------- | --------------- | ---------------- | ----- |
| Ubuntu   | install + cache restore + rebuild | 86s             | 32s              | -54s  |
| Windows  | install + cache restore + rebuild | 332s            | 260s             | -72s  |

All optimized warm-cache runs logged `Electron native cache hit: true` on Ubuntu and Windows.


IDCs (dismissed):

.github/workflows/main.yml
```diff
@@ -47,6 +47,40 @@ jobs:
...
         run: pnpm install
+        env:
+          VORTEX_DEFER_ELECTRON_REBUILD: "1"
+
+      - name: Cache native build headers
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.electron-gyp
+            ~/.cache/node-gyp
```
erri120, line 58: Does this work on Windows as well?

.github/workflows/main.yml
```diff
@@ -47,6 +47,40 @@ jobs:
...
+          key: native-headers-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('package.json', 'pnpm-lock.yaml', 'src/main/package.json') }}
```
erri120:
```suggestion
          key: native-headers-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('package.json', 'pnpm-lock.yaml') }}
```

No need to hash `src/main/package.json` if you include `pnpm-lock.yaml` as any changes to dependencies will be reflected in the lock file.

.github/workflows/main.yml
```diff
@@ -47,6 +47,40 @@ jobs:
...
+
+      - name: Cache Electron native rebuild outputs
+        id: electron-native-cache
+        uses: actions/cache@v4
+        with:
+          path: |
```
erri120, line 65: How are we going to maintain this list of paths?

.github/workflows/main.yml
IDCs: I was concerned about this one as well, but looks like `electron-rebuild` resolves headers through `@electron/get` which uses `~/.electron-gyp` on all platforms.

Still - would be good to add `~/AppData/Local/node-gyp/Cache` I guess.

Sewer56: It does; can be proven by looking in the CI logs actually.

I can add the Windows paths explicitly if it's more clear, but CI was successful on both platforms in this case- so I left it simple.

.github/workflows/main.yml
Sewer56: Yeah, seems you're right about this one 👍 thanks!

Originally I had `package.json` in here because this was all one big branch with all of the optimisations combined. Then I split this up into smaller PRs that are easier to describe, measure and review individually.

I had `package.json` in here because caching our Nx builds was one of the original things I tested at the time; thought was- if our build instructions change (scripts); the cache should be invalidated too. But I had to defer that for later; and so kept only native parts 

In which case, yup, now it's gone :clap: .
3eb7860a96cc423d10ee8a34efc3b68dc7f606fa

.github/workflows/main.yml
Sewer56: To be more specific, while `node-gyp` itself *does* use `%LOCALAPPDATA%\node-gyp\Cache`; we use `@electron/rebuild` (`pnpm exec electron-rebuild`) to build the native deps.

Then, as @IDCs noted; `electron-rebuild` defaults to `os.homedir()/.electron-gyp` regardless of platform.

So yeah; what's here is correct.

.github/workflows/main.yml
Sewer56: Maintenance burden is low because:

- **Cache miss != breakage**: an unlisted path just means that dep rebuilds every run (same as no cache, same as before). No risk.
- **Stale entries are harmless**: a listed path that doesn't exist on disk is simply ignored. No error; only warning.
- **The list shrinks, not grows**: we're actively removing native deps from Vortex, not adding them. We don't expect new entries anytime soon.
- **Could automate**: a short script scanning `node_modules` could regenerate this list when deps change. Given the above however, not worth the effort right now.

Not worried about it at the moment; since we're only planning to be removing deps. If a dep is removed then warning will show, so we'll know immediately by opening a CI run.

IDCs (dismissed):

IDCs:
Same as #23274

IDCs (approved):

PR #23276, opened 2026-05-21
Add GitHub workflow to publish Vortex releases via Upload API

Sewer56:
Closes [LAZ-229](https://linear.app/nexus-mods/issue/LAZ-229/add-github-workflow-to-publish-vortex-releases-via-the-upload-api).

First merge
- #23275

## Summary

Adds a GitHub Actions workflow that publishes Vortex releases to Nexus Mods via the Upload API.

## Change

New workflow `.github/workflows/publish-release.yml` triggers on release events.

## Prepare script

`scripts/publish-release-to-nexus/` is a TypeScript module with testable pure functions (inject a fake `ghRun` and temp dir):

1. Fetches release metadata via `gh release view --json`
2. Finds most recent stable release (not draft, not prerelease)
3. Selects the `.exe` installer asset from the release
4. Strips the leading `v` from the tag to produce a plain version
5. Downloads the installer (skipped in dry-run)
6. Writes `GITHUB_OUTPUT` keys for the upload step (multiline body uses heredoc syntax)

Documentation in `docs/publishing-releases.md` covers the publishing process and API usage.

There's a fair bit of logic here; so rather than making a long, untestable workflow yaml; it's been lifted out to a script; which is called from the yaml.

## Modified Files

- `.github/workflows/publish-release.yml` - new release-publishing workflow
- `scripts/publish-release-to-nexus/index.ts` - workflow entry point
- `scripts/publish-release-to-nexus/prepare.ts` - pure publish-preparation functions
- `scripts/publish-release-to-nexus/prepare.test.ts` - vitest suite for prepare module
- `docs/publishing-releases.md` - new documentation for the publishing process


scripts/package-lock.json
erri120, line 1: Where did this file come from?

erri120 (approved):

scripts/package-lock.json
Sewer56: Pretty sure it was from running `npm install`.
Safe to delete -> done.
Probably a leftover from when I tore this PR and the other one that cleans up unused scripts apart.

Sewer56:
image

?? What happened here. I pushed, then merged from GitHub UI, but GitHub merged without the delete commit, and didn't even note the commit happened here.

insomnious:
Could we potentially have a Vortex-Test mod (that is hidden) under the `site` game domain so we can fully test this e2e on production? Do hidden mod pages work with the upload api? Do we need to manually upload an old version first in order to get a group id for us to test the workflow?

PR #23313, opened 2026-05-26
E2E test improvements and fixes

erri120:
- Use global timeout to prevent sudden exists in CI
- Fixes language options selector for settings test
- Fixes screenshots by using electron `capturePage` which is much faster

Instead of needing 45 minutes to run tests, tests now finish in under a minute without the login tests and under 3 minutes with the login tests (which still fail locally).

### DIFF
--- a/packages/e2e/fixtures/vortex-app.ts
+++ b/packages/e2e/fixtures/vortex-app.ts
@@ -184,7 +184,7 @@ export const test = base.extend<VortexTestFixtures, VortexWorkerFixtures>({
 
       const app = await electron.launch({
         executablePath: electronBinary,
-        args: [mainDir, "--disable-gpu", "--disable-software-rasterizer"],
+        args: [mainDir],
         env: buildElectronEnv(sharedUserDataDir),
         cwd: mainDir,
         // CI runners are slower — allow up to 2 minutes for cold start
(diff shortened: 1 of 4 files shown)


packages/e2e/fixtures/vortex-app.ts
```diff
@@ -237,15 +242,40 @@ export const test = base.extend<VortexTestFixtures, VortexWorkerFixtures>({
...
+        .then((base64) => {
+          if (!base64) {
+            return Promise.resolve();
+          }
+          return testInfo.attach("screenshot", {
+            body: Buffer.from(base64, "base64"),
+            contentType: "image/png",
+          });
+        })
+        .catch((e) => console.error(`Failed to capture screenshot: ${e}`));
```
Aragas, line 278: Might be worth to create a screenshot() function that could be used by the tests?

erri120: From README:

> No inline `page.screenshot()` in tests.

So no screenshots in tests, regardless of implementation.

Aragas: Tbh the reason the text is there was because `page.screenshot()` wasn't working, so since we are a desktop app, it might be useful, but it would be good to hear @Nexus-Mods/vortex-qa opinion on this.
AFAIK the site is not doing screenshots

erri120: If anyone needs it they can make a function out of it.

Marcin-NM: We don't use `page.screenshot()` on the site, I doubt we'll need it here.

Marcin-NM (approved):

PR #23337, opened 2026-05-27
chore: remove stray scripts/package-lock.json

Sewer56:
See:
- https://github.com/Nexus-Mods/Vortex/pull/23276#issuecomment-4553904145


IDCs (approved):
odd

PR #23349, opened 2026-05-28
[QA-94] add view-profile-on-web e2e test

Marcin-NM:
## Summary
- Adds [QA-94](https://linear.app/nexus-mods/issue/QA-94) e2e test: free user clicks "View profile on web" in the profile dropdown and is taken to their Nexus Mods profile URL via `shell.openExternal`.
- Mirrors the QA-96 (upgrade-to-premium) pattern: stub `shell.openExternal` in the Electron main process to capture URLs, then assert the captured URL matches `nexusmods.com/users/<id>`.

## Implementation notes
- `loginToNexus`'s logged-in verification step clicks the avatar and leaves the dropdown open. The test dismisses it with `Escape` first, then drives its own open/click cycle so the assertions don't depend on the helper's side effects.
- No new selectors or helpers — purely a new spec.

## Test plan
- [x] Local: `pnpm --filter @vortex/e2e exec playwright test view-profile-on-web.spec.ts` passes (1.1m).
- [x] Local: `pnpm run format:check` clean under WSL Linux oxfmt.
- [x] Local: typecheck/lint clean as part of `pnpm run build`.

### DIFF
--- a/packages/e2e/src/tests/view-profile-on-web.spec.ts
+++ b/packages/e2e/src/tests/view-profile-on-web.spec.ts
@@ -1,0 +1,48 @@
+/**
+ * QA-94: free user opens the profile dropdown and clicks "View profile on web",
+ * which should open their Nexus Mods profile URL in the system browser.
+ */
+import { test, expect } from "../fixtures/vortex-app";
+import { freeUser } from "../helpers/users";
+
+test.describe("Account - View profile on web", () => {
+  test.use({ nexusUser: freeUser });
+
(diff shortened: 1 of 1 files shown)


erri120 (changes requested):
Use new timeout constants instead of magic numbers. E2E best practices have also changed on timeouts. Best to wait for #23376 as login has changed as well.

erri120:
Needs a rebase.

erri120 (approved):

PR #23365, opened 2026-06-01
UI component updates

r1chm8:
Refactors shared UI primitives, extracts TypographyLink into its own module, adds Image and PremiumBadge, and replaces hard-coded calc(var(--spacing) * N) radii with the named --radius-* tokens across element/form stylesheets.

### Components
- Button.tsx — when icon is a custom element, clone it with the nxm-button-icon class instead of wrapping in a <span>, so existing classes (sizing, colors) are preserved.
- Icon.tsx — adds shrink-0 by default to stop icons squashing in flex rows; tightens title falsy check.
- Image.tsx — new <Image> component, handles all default styles and aspects out the box.
- PremiumBadge.tsx — new component; replaces the inline diamond-badge markup previously hard-coded in ModRequirement.
- Typography.tsx — slimmed down: TypographyLink and its types removed (moved to TypographyLink.tsx); getTypographyStyles exported so the link module can reuse it. 
- TypographyLink.tsx — new file, hosts the extracted component.
  
### Stylesheets
- button.css — switch to --radius-sm; add flex-shrink: 0; nest size modifiers (xs/sm) inside @layer components; recolor secondary/tertiary variants to the translucent palette.
- dropdown.css, modal.css, pagination.css, checkbox.css, input.css — replace calc(var(--spacing) * N) corner radii with --radius-xs/sm/lg tokens.
- link.css — drop redundant outline declarations on :focus-visible; add a default --radius-xs.
- tab.css — add overflow-x: clip + overflow-clip-margin on .nxm-tab-bar to stop the focus-ring overhang from triggering a page-level horizontal scrollbar.
- theme/generic.css — add --aspect-collection (4/5) and --aspect-mod (16/9) tokens for new image component.
  
### Consumers
- ModRequirement.tsx — adopts the new <PremiumBadge /> in place of the inline diamond markup.
- HealthCheckDetailPage.tsx — imports TypographyLink from its new module and drops the invalid as="button" prop.

### DIFF
--- a/src/renderer/src/ui/components/icon/Icon.tsx
+++ b/src/renderer/src/ui/components/icon/Icon.tsx
@@ -1,6 +1,6 @@
 import React, { type SVGAttributes } from "react";
 
-import { joinClasses } from "../../utils/joinClasses";
+import { joinClasses } from "@/ui/utils/joinClasses";
 
 export type IconSize = "xs" | "sm" | "md" | "lg" | "xl" | "2xl" | "none";
 
@@ -26,12 +26,12 @@ export const Icon = ({
   title?: string;
 }) => (
   <svg
-    className={joinClasses([sizeMap[size], className])}
+    className={joinClasses(["shrink-0", sizeMap[size], className])}
     role={title ? "img" : "presentation"}
     viewBox="0 0 24 24"
     {...props}
   >
-    {title && <title>{title}</title>}
+    {!!title && <title>{title}</title>}
 
     <path d={path} fill="currentColor" />
   </svg>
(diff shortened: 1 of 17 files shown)


src/renderer/src/ui/components/icon/Icon.tsx
```diff
@@ -26,12 +26,12 @@ export const Icon = ({
...
 }) => (
   <svg
-    className={joinClasses([sizeMap[size], className])}
+    className={joinClasses(["shrink-0", sizeMap[size], className])}
     role={title ? "img" : "presentation"}
     viewBox="0 0 24 24"
     {...props}
   >
-    {title && <title>{title}</title>}
+    {!!title && <title>{title}</title>}
```
Aragas, line 34: IIRC we don't use `!!`, just do `{title` instead

r1chm8: We have a lint rule that says otherwise?

image

erri120: One of the rare cases where you do want `!!` to explicitly coerce the value into a boolean.

Details:

https://react.dev/learn/conditional-rendering#logical-and-operator-
https://eslint-react.xyz/docs/rules/no-leaked-conditional-rendering

Aragas: Thanks! Makes sense

Aragas (approved):

PR #23367, opened 2026-06-01
Added an optional discriminator to Fingerprints

Aragas:
Grouping by the first 5 lines of the stacktrace still combines error that should not be combined.
Discriminating by using the optional `error.code` should help  it for now


Sewer56 (dismissed):
Sure thing.

Could reduce the yapping though.
That's a hell of a lot of Anthropic-generated yapping for a concept that's rather quite simple.

Aragas:
Sorry, I though it was removed. Thanks for noticing

Sewer56 (approved):

PR #23393, opened 2026-06-03
Using error serialization with standard IPC

Aragas:
Trying to be as minimalistic as possible.
The issue is that we lose a lot of information with Eletron's error serialization. It basically drops all properties and only keeps the `message` value.
We'll need to discuss later whether we want and whether we can restore the errors prototype when crossing IPC, but this will suffice for now.
The main goal for this PR is to fix the error reporting:
* Some errors we skip based on `instanceof` (which doesn't work now - we use `name` instead via #23391
* Some errors we skip based on the `code` (IO stuff)

Both are not working at the moment, since neither `name` or `code` are passed correctly.


Sewer56 (approved):
It'll do for now :slightly_smiling_face:

PR #23427, opened 2026-06-08
feat: add Bullet and AdultAwareImage UI components

r1chm8:
Adds a reusable Bullet dot and an adult-aware Image wrapper that blurs Nexus content per the user's adultBlurImages preference (blurring by default when logged out). Includes design-system demos and README entries; CollectionTile now uses Bullet.

### DIFF
--- a/src/renderer/src/extensions/design_system_dev/views/DesignSystemPage.tsx
+++ b/src/renderer/src/extensions/design_system_dev/views/DesignSystemPage.tsx
@@ -7,6 +7,7 @@
 import React, { useState } from "react";
 
 import type { IExtensionApi } from "@/types/IExtensionContext";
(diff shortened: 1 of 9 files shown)


IDCs (approved):
Cool stuff, I'm assuming that we're leaving `collectionTile` as is intentionally?
(It's using the new bullet but not the `AdultAwareImage` stuff)

Approving either way.

r1chm8:
> Cool stuff, I'm assuming that we're leaving `collectionTile` as is intentionally? (It's using the new bullet but not the `AdultAwareImage` stuff)
> 
> Approving either way.

Technically a collection tile image should not contain adult content, so even if the collection is adult, the image won't be, on the web we don't blur them, so I've followed the same logic here. Should be a simple change if we want to change that logic though.

PR #23433, opened 2026-06-08
Encrypt reports

erri120:


Sewer56 (commented):
Curious- what's the motivation here? Not exposing test accounts, internal IPs in case of errors/traces or something?

erri120:
> Curious- what's the motivation here? Not exposing test accounts, internal IPs in case of errors or something?

See https://github.com/microsoft/playwright/issues/19992. Playwright traces and reports can include sensitive data and this is the only reliable way of not leaking them.

IDCs (approved):

PR #23436, opened 2026-06-09
fixed licenses in about page not being displayed

IDCs:
Wire dependency-report into nx as a copy-assets dependency so assets/modules.json is generated before the asset copy. It was missing from packaged builds, leaving the third-party license list empty.

Note: v2.1 has now diverged from master, which uses pnpm sbom / bom.json instead. This fix is specific to v2.1's modules.json mechanism.

closes app-499

### DIFF
--- a/src/main/package.json
+++ b/src/main/package.json
@@ -64,6 +64,22 @@
           "script": "package:nosign"
         }
       },
+      "dependency-report": {
+        "executor": "nx:run-commands",
+        "options": {
+          "command": "node ./scripts/dependency-report.mjs",
+          "cwd": "{workspaceRoot}"
+        },
+        "cache": true,
+        "inputs": [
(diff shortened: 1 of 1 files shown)


src/main/package.json
```diff
@@ -64,6 +64,22 @@
...
+        "executor": "nx:run-commands",
+        "options": {
+          "command": "node ./scripts/dependency-report.mjs",
+          "cwd": "{workspaceRoot}"
+        },
+        "cache": true,
+        "inputs": [
+          "{workspaceRoot}/scripts/dependency-report.mjs",
+          "{workspaceRoot}/pnpm-lock.yaml",
+          "{projectRoot}/package.json"
```
erri120:
```suggestion
        "inputs": [
          "default",
          "{workspaceRoot}/scripts/dependency-report.mjs"
```

erri120 (approved):

PR #23441, opened 2026-06-09
[QA-127][QA-128] add settings-defaults e2e tests

Marcin-NM:
## What

Adds a single e2e spec covering two Settings-page UI defaults:

- **QA-127** — Settings opens on the **Interface** tab by default
- **QA-128** — Default language is set to **English**

## Details

New spec: `packages/e2e/src/tests/settings-defaults.spec.ts` (2 tests).

Both are pure Settings-page assertions — no login, no game management, no
fake-game fixtures (default `nexusUser: null`). Follows the post-#23376
fixture style.

- QA-127 asserts the active `role="tab"` reads "Interface" (registered at
  priority 10, so it's the default/first tab).
- QA-128 asserts the Language `<select>` value is `"en"` (the redux default
  in `settings.interface.language`).

## Testing

- `pnpm run build` ✅
- `pnpm --filter @vortex/e2e exec playwright test settings-defaults.spec.ts` — 2 passed ✅
- `pnpm run format:check` (Linux oxfmt) ✅

### DIFF
--- a/packages/e2e/src/tests/settings.spec.ts
+++ b/packages/e2e/src/tests/settings.spec.ts
@@ -11,20 +11,18 @@ test.describe("Settings - Interface Tab", () => {
     await navigateToSettings(vortexWindow);
   });
 
-  test("opens on the Interface tab by default", async ({ vortexWindow }) => {
-    await test.step("Verify settings page has content", async () => {
-      const pageContent = await vortexWindow.locator("body").innerText();
-      expect(pageContent).toBeTruthy();
+  test("[QA-127] opens on the Interface tab by default", async ({ vortexWindow }) => {
+    await test.step("Verify the Interface tab is the active tab", async () => {
+      const activeTab = vortexWindow.getByRole("tab", { selected: true });
+      await expect(activeTab).toHaveText("Interface");
     });
   });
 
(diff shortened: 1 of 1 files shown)


erri120 (changes requested):
This is a duplicate test. See `src/tests/settings.spec.ts`.

Marcin-NM:
Good catch - I've dropped the separate `settings-defaults.spec.ts` and instead strengthened the two existing tests, which previously didn't actually assert the behaviour:

- **"opens on the Interface tab by default"** only checked that `body.innerText()` was truthy — it never verified the Interface tab was active. It now asserts the selected `role="tab"` reads `"Interface"`.
- **"default language is set to English"** was wrapped in `if (languageLabel.isVisible())`, so it silently passed when the label wasn't found. It now asserts `languageSelect` has value `"en"` unconditionally.

No new file — just the two assertions made real. Verified locally (`settings.spec.ts`: 11 passed) and force-pushed.

erri120:
Should probably annotate with QA-127 and QA-128 to keep track of that.

Marcin-NM:
Done — annotated both tests with `[QA-127]` / `[QA-128]` prefixes in their titles (matching the `[QA-94]` convention already in the suite).

erri120 (approved):

PR #23458, opened 2026-06-11
Extension auto-update fix

### LINKED ISSUE #23454: A Starfield Extension failed to load
> **Important:** Tickets that do not provide sufficient information may be closed without further notice. Context is everything when diagnosing bugs — do not assume that a screenshot alone is enough for us to identify the problem. Logs and reproduction steps are extremely important and, unless the issue is immediately obvious, we cannot assist without them. Please attach all `.log` files from `%appdata%/vortex`.

**Your Nexus Mods username**
Vlad 254

**Describe the bug**
I updated Vortex this morning and now I get a message saying a Starfield Extension failed to load but I checked and my Starfield extension is installed. I do not recall which other Starfield extensions I may have installed.

**To Reproduce**
Every time I launch Vortex I get the message

**Expected behavior**
I expect to either not receive the message or find out the culprit

**Logs**
Please attach all `.log` files from `%appdata%/vortex`. Without logs, we are unlikely to be able to help.

**Screenshots**

- OS: Windows 11

- Vortex Version: [e.g. 1.16.7]
2.1.0

Game:
Starfield

**Additional context**
none

Aragas:
We have a race condition that prevents at least some users from extension being auto-updated. The fix is self-healing, so affected users will have the issue fixed.

Closes [LAZ-571](https://linear.app/nexus-mods/issue/LAZ-571/extension-auto-update-fails-every-launch-starfield-v11010)
Fixes #23454

### DIFF
--- a/src/renderer/src/extensions/extension_manager/util.ts
+++ b/src/renderer/src/extensions/extension_manager/util.ts
@@ -28,7 +28,6 @@ import { writeFileAtomic } from "../../util/fsAtomic";
 import getVortexPath from "../../util/getVortexPath";
 import { log } from "../../util/log";
 import { jsonRequest, rawRequest } from "../../util/network";
-import { getSafe } from "../../util/storeHelper";
 import { INVALID_FILENAME_RE, truthy } from "../../util/util";
 import { addLocalDownload, setDownloadModInfo } from "../download_management/actions/state";
 import { downloadPathForGame } from "../download_management/selectors";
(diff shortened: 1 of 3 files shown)


src/renderer/src/extensions/extension_manager/util.ts
```diff
@@ -260,6 +260,45 @@ function doFetchAvailableExtensions(
...
+// to reach a terminal state instead of reading the store once.
+export function waitForDownloadRecord(
+  api: IExtensionApi,
+  dlId: string,
+  timeoutMS: number = 30000,
+  intervalMS: number = 250,
+): PromiseBB<IDownload> {
+  const tryFetch = (remainingMS: number): PromiseBB<IDownload> => {
+    const state: IState = api.store.getState();
+    const download: IDownload = getSafe(
```
IDCs: I know that we're using `getSafe` all over the place; but lets just use the ternary operator in new code.

IDCs (changes requested):

erri120:
Would also recommend a rebase so tests run on CI. (nvm, wrong branch)

IDCs (approved):
@Aragas - my bad for not mentioning this on the first review pass, but I think `waitForDownloadRecord` should be a standard utility function that can be used throughout the codebase, not just the extension manager.

Going to approve but would be good to extract this function in the future.

Aragas:
Created [LAZ-600 ](https://linear.app/nexus-mods/issue/LAZ-600/extract-waitfordownloadrecord-into-a-shared-utility)so we won't forget about this

PR #23462, opened 2026-06-11
Fold collections extension into the renderer (LAZ-483 Stage A+B)

IDCs:
First step of the LAZ-483 collections-reliability refactor. Moves the standalone
`collections` extension into the main repo as an in-repo static extension so that
later stages can make the Redux collections session the single source of truth for
install completion. Behaviour-neutral.

Targets the `laz-483-integration` branch (the long-lived base for the LAZ-483 work);
subsequent stage PRs stack into it, and it merges to master/release once 2.2 beta has
branched off.

### Commits
- **Fold collections into the renderer** — move `extensions/collections/src` to
  `src/renderer/src/extensions/collections`, rewire the `@nexusmods/vortex-api`
  barrel imports to relative core paths, register in `staticExtensions` (ordered
  after its init dependencies), route the stylesheet/icons/`hdiff.wasm` and the
  bsdiff worker into the core build pipeline, and remove the old bundled-extension
  nx project.
- **Fold `collections_integration` into `collections`** — colocate the Redux
  install-session types/selectors with the tracking reducer/actions.
- **Error-handling cleanup** — narrow `catch (err: any)` to `unknown` with the
  `@vortex/shared` helpers, and forward `unknownToError(err)` to
  `showErrorNotification` instead of a raw `unknown`.

### Checks
- Full build, lint, typecheck, format green across all 148 projects.
- 188 collections unit tests pass.
- Binary patching verified working in-app (patched files CRC-checked on disk).


erri120 (approved):
The bsdiff stuff is really scuffed, are those going into the main process later on?

IDCs:
No - I don't like the current workaround for the workers either, so those are going to change.

PR #23481, opened 2026-06-15
separate mod identity matching from install-spec matching (483 WIP)

IDCs:
A mod reference meant two things at once. In mod_management it identifies WHICH mod (id, md5, logical name, repo, version). Collections also hung installer choices, file list, and patches off the same IModReference, so it encoded HOW the mod was installed too, and testModReference matched on both. That conflated "is this mod installed" with "was it installed the way this rule asks for", leaving no single source of truth for what a rule means and making the matching logic hard to reason about and test.

This makes identity the sole concern of testModReference and moves install-spec matching into dedicated, testable helpers shared by both subsystems.

- testModReference matches by identity only; removed installerChoices/fileList/patches from IModReference.
- Add IModInstallSpec, modMatchesInstallSpec, and ruleInstallSpec (the single reader of a rule's install-spec fields).
- findModByRef takes an optional installSpec to pick the right variant after identity matches (multiple variants of a mod can be installed at once).
- Promote patches to a first-class IModRule field, with a legacy rule.extra.patches fallback in ruleInstallSpec so older rules still match.
- Solidify shared types: installerChoices=IChoiceType, patches=IModPatches, fileList=IFileListItem[] everywhere.
- Convert touched collections/load-order files from **/api barrels to direct imports to avoid transient deps in tests.
- Add regression tests for the decoupling, ruleInstallSpec fallback, and findModByRef variant selection.

### DIFF
--- a/etc/vortex.api.md
+++ b/etc/vortex.api.md
@@ -575,7 +575,7 @@ const collapseGroup: reduxAct.ComplexActionCreator3<string, string, boolean, {
 }, {}>;
 
 // @public
-type CollectionModStatus = "pending" | "downloading" | "downloaded" | "installing" | "installed" | "failed" | "skipped" | "optional";
+type CollectionModStatus = "pending" | "downloading" | "downloaded" | "installing" | "installed" | "failed" | "ignored" | "optional";
 
 // Warning: (ae-forgotten-export) The symbol "MixpanelEvent" needs to be exported by the entry point api.d.ts
 //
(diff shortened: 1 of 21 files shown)


erri120 (approved):
Should clean up `{ [something: K]: V }` to `Record<K, V>` at some point.

IDCs:
Will add a note on my end for the next consolidation pass.

PR #23484, opened 2026-06-15
Fix types in main project

erri120:
Removes `any` and turns the eslint warnings back into errors.

### DIFF
--- a/src/main/src/store/ReduxPersistorIPC.ts
+++ b/src/main/src/store/ReduxPersistorIPC.ts
@@ -285,10 +285,9 @@ class ReduxPersistorIPC {
 
   private async applySetRun(persistor: IPersistor, run: DiffOperation[]): Promise<void> {
     if (persistor.bulkSetItem !== undefined) {
-      const bulk = persistor.bulkSetItem.bind(persistor);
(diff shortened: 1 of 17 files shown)


src/main/src/store/ReduxPersistorIPC.ts
```diff
@@ -285,10 +285,9 @@ class ReduxPersistorIPC {
 
   private async applySetRun(persistor: IPersistor, run: DiffOperation[]): Promise<void> {
     if (persistor.bulkSetItem !== undefined) {
-      const bulk = persistor.bulkSetItem.bind(persistor);
       for (let start = 0; start < run.length; start += ReduxPersistorIPC.BULK_CHUNK_SIZE) {
         const chunk = run.slice(start, start + ReduxPersistorIPC.BULK_CHUNK_SIZE);
-        await bulk(
+        await persistor.bulkSetItem(
```
Aragas, line 290: I assume this change is fine?

Aragas (approved):

src/main/src/store/ReduxPersistorIPC.ts
erri120: The `bind` doesn't make a difference here since we're invoking the method directly anyways. `bind` is only necessary in callback situations.

PR #23509, opened 2026-06-16
Add Unleash feature flag support

erri120:
Closes LAZ-592.

The massive change count comes from the fact that this includes the `openapi.json` spec of the Unleash API.


src/main/src/main.ts
```diff
@@ -279,6 +281,11 @@ async function main(): Promise<void> {
     });
   });
   initTelemetryIpcHandler();
+
+  const unleashClient = new UnleashClient(app.getVersion());
+  synchronizeFeatureFlags(unleashClient);
```
IDCs, line 286: `synchronizeFeatureFlags` is already calling `unleashClient.start`, is the `start` call below needed?

IDCs (dismissed):
Other than the duplicate `client.start` call it looks good to me. Going to approve anyway.

src/main/src/main.ts
erri120: Nope, fixed.

IDCs (approved):

PR #23517, opened 2026-06-16
Update Flatpak pnpm source preparation

danyloburavlov:
## Summary

This updates the Flatpak packaging flow to use `pnpm` instead of the old Yarn-based dependency setup, then adds locked DuckDB extension artifacts so the Flatpak build can keep working offline.

The repo now uses `pnpm-lock.yaml`, but the Flatpak source generation path still depended on Yarn-style generated sources and Yarn install commands. This PR switches the Flatpak JavaScript dependency sources to pnpm, installs with `pnpm install --frozen-lockfile --offline`, and regenerates `flatpak/generated-sources.json` from the current pnpm lockfile.

While testing that path, DuckDB extension downloads became the next offline-build gap. The downloader was still resolving extension URLs at build time, which does not fit Flatpak because external files need to be declared before the sandboxed build starts. This PR adds a DuckDB extension lockfile with pinned URLs and SHA256 hashes, teaches the downloader to consume it, and generates Flatpak sources for those locked extension artifacts.

### Flatpak pnpm migration

- Updated the Flatpak manifest to use pnpm cache/store settings instead of Yarn cache settings
- Replaced Flatpak build commands from `yarn ...` to `pnpm ...`
- Switched `flatpak-node-generator` from Yarn mode to pnpm mode
- Updated the pinned `flatpak-builder-tools` commit to one that includes pnpm lockfile support
- Added `flatpak/scripts/flatpak_prepare_pnpm.py` to prepare the copied Flatpak build tree before offline installation
- Normalized the pnpm lockfile used by the Flatpak generator/build so the Flatpak SDK-provided Node runtime is used instead of the repo-managed Node runtime package
- Normalized generated Flatpak source paths to POSIX separators so source generation remains Linux-compatible when invoked from Windows
- Removed the old Yarn hash helper and `flatpak/yarnrc`
- Regenerated `flatpak/generated-sources.json` and `flatpak/generated-sources.hash` from the rebased `pnpm-lock.yaml`
- Included the offline dependencies required by the vendored Nexus API v3 OpenAPI codegen in the regenerated pnpm sources
- Updated Flatpak documentation to describe the pnpm-based workflow

### DuckDB extension locking

- Added `src/main/duckdb-extensions.lock.json` to pin DuckDB extension artifacts by DuckDB version, platform, URL, and SHA256
- Updated the DuckDB extension downloader to validate lock freshness, verify SHA256 before unpacking, support `--update-duckdb-lock`, and install from a prefetched source directory when `VORTEX_DUCKDB_EXTENSION_SOURCE_DIR` is set
- Changed the downloader to replace existing unpacked extensions whenever its Nx target executes, because an artifact can change without the DuckDB version changing and file existence alone does not prove it matches the current lockfile
- Added package scripts for downloading extensions and refreshing the DuckDB lockfile
- Updated the Nx target so the DuckDB lockfile is part of the cached extension download inputs
- Updated `scripts/generate-query-types.ts` to load the same locked local `level_pivot` extension instead of resolving `current_release`
- Added tests for lockfile validation, locked artifact lookup, SHA256 checking, platform mapping, and DuckDB extension path layout
- Made DuckDB path-layout tests portable across Windows and Linux
- Extended `flatpak/scripts/flatpak_sources.py` to emit `generated-duckdb-sources.json`
- Added `flatpak/generated-duckdb-sources.json` and `flatpak/generated-duckdb-sources.hash`
- Updated the Flatpak manifest/workflow so DuckDB extension artifacts are prefetched by Flatpak and consumed from the offline build sandbox

## Tech Notes

The generated JavaScript source diff is large, but that is expected. The output moved from Yarn-style generated sources to pnpm-style generated sources, including pnpm store population data based on `pnpm-lock.yaml`.

The Flatpak pnpm preparation step removes the managed Node runtime from the copied package metadata and normalizes the lockfile for the build sandbox. Flatpak already provides Node through `org.freedesktop.Sdk.Extension.node24`, so the offline pnpm install uses that SDK runtime instead of trying to resolve or download Node as a package dependency inside the sandbox.

The Nexus API v3 package generates types from the committed `packages/nexus-api-v3/schema/openapi.yaml` file rather than fetching the schema during installation. Its `openapi-typescript`, `openapi-fetch`, and Redocly dependencies are included in `flatpak/generated-sources.json`, so this codegen step remains offline-compatible.

Generated source paths are normalized after `flatpak-node-generator` runs. This prevents a source manifest generated on Windows from containing backslash-separated destinations or shell commands that would not resolve correctly in the Linux Flatpak build sandbox.

DuckDB extension locking pins the compressed `.duckdb_extension.gz` artifacts because those are the files Flatpak fetches before the build starts. The downloader verifies the compressed file hash first, then unpacks it into the normal build output directory.

Normal app builds download the locked DuckDB URLs directly. Flatpak builds set `VORTEX_DUCKDB_EXTENSION_SOURCE_DIR`, so the downloader reads the already-prefetched artifacts from the Flatpak source directory instead of using the network.

The downloader always replaces an existing unpacked extension when its Nx target executes. A mutable upstream URL can serve different bytes for the same DuckDB version, so an existing output file cannot prove that it matches the current lock. Nx caching avoids repeated work while the lockfile and other target inputs remain unchanged.

`scripts/generate-query-types.ts` also uses the locked extension flow. It validates the lockfile against the installed `@duckdb/node-api`, checks that the host platform artifact has been downloaded, sets DuckDB's `extension_directory`, and runs `LOAD level_pivot` from the local locked output. This keeps generated query types aligned with the extension artifact Vortex packages.

Refreshing the DuckDB lockfile is intentionally explicit:

```bash
pnpm run --filter @vortex/main duckdb:update-duckdb-lock
```

This keeps normal builds from silently accepting changed upstream DuckDB extension artifacts.

## Open Question

- [ ] DuckDB extension artifact URLs currently come from `current_release`. The lockfile pins SHA256, so changed upstream bytes fail builds instead of being accepted silently. However, previous builds are only reproducible while the old bytes remain available at that URL. How should the build process work if the actual bytes behind `current_release` change, but the previous bytes are still preferred?

**Proposed answer**: keep the SHA256 lock as the build gate, and do not auto-refresh it during normal builds. If `current_release` changes and the hash check fails, the build should fail with a clear message and require a maintainer decision. If the new bytes are intended, refresh the lockfile. If the previous bytes are still required, they need to come from an immutable source that still has them, such as a release asset, CI artifact, or project-owned mirror. If no such source exists, the previous build is not fully reproducible from `current_release` alone. Longer term, the extension publish flow should expose immutable versioned artifact URLs so the lockfile can pin both the URL and hash without depending on `current_release`.

## Verification Boundary

SHA256 verification remains centralized in the downloader. The downloader always recreates the unpacked extension when its Nx target executes, while Nx caching skips that work when the lockfile and other inputs are unchanged. The query type generator requires the resulting local extension to exist and relies on this verified downloader output.

If maintainers want query generation to be independently verifiable later, the lockfile could also record the unpacked extension hash, or the generator could invoke the downloader flow directly before loading DuckDB.

## Validation Performed

Regenerated the pnpm Flatpak node sources after rebasing:

```bash
python3 flatpak/scripts/flatpak_sources.py --only pnpm --force
```

Regenerated the Flatpak DuckDB sources from the DuckDB extension lockfile:

```bash
python3 flatpak/scripts/flatpak_sources.py --only duckdb --force
```

Confirmed that generated pnpm source paths and shell commands contain no Windows path separators.

Confirmed that generated pnpm sources include the Nexus API v3 OpenAPI codegen dependencies.

Ran the `@vortex/main` test suite, including the DuckDB extension tests:

```text
Test Files  27 passed (27)
Tests       309 passed | 2 skipped (311)
```

Ran targeted ESLint and formatting checks for the DuckDB downloader and tests.

Regenerated query types using the locked local DuckDB extension:

```bash
pnpm run generate:query-types
```

Checked that the generated Flatpak DuckDB source file is valid JSON:

```bash
python3 -c "import json; json.load(open('flatpak/generated-duckdb-sources.json'))"
```

## What Is Still Needed To Be Done

This gets the Flatpak dependency/source-generation path onto pnpm and makes DuckDB extension artifacts available to the Flatpak build offline, but it does not prove the final Flatpak package is ready to publish.

1. Run a full Flatpak build
2. Smoke test the built Flatpak
3. Confirm install and bundle workflows still work end to end
4. Wire generated pnpm, DuckDB source, and query type freshness checks into CI or Flatpak validation if needed


MattSturgeon:
When experimenting with a [nix package](https://github.com/MattSturgeon/Vortex/tree/nix-package), I noticed that [`download-duckdb-extensions`](https://github.com/Nexus-Mods/Vortex/blob/bd48c9cabbdf5362e5951d4ce86f9516af67599a/src/main/download-duckdb-extensions.ts) is downloading inputs from an unlocked manifest ([`duckdb-extensions.json`](https://github.com/Nexus-Mods/Vortex/blob/bd48c9cabbdf5362e5951d4ce86f9516af67599a/src/main/duckdb-extensions.json)).

Did you need to address this for the flatpak package too?[^1] IIRC, flatpak has similar requirements for deterministic dependencies.

[^1]: I don't see anything obvious in the diff

Sewer56:
Pretty interesting stuff; not sure at which time I'll be able to review this personally- as doing Linux support outside of dev-environment is more further down the road, but I'm all down for it.

danyloburavlov:
Thank you!

I do agree with Matt's comment and I'll try my best to address it.

> Linux support outside of dev-environment is much more further down the road

I hope it won't take _too_ long :sweat_smile:

danyloburavlov:
> When experimenting with a [nix package](https://github.com/MattSturgeon/Vortex/tree/nix-package), I noticed that [`download-duckdb-extensions`](https://github.com/Nexus-Mods/Vortex/blob/bd48c9cabbdf5362e5951d4ce86f9516af67599a/src/main/download-duckdb-extensions.ts) is downloading inputs from an unlocked manifest ([`duckdb-extensions.json`](https://github.com/Nexus-Mods/Vortex/blob/bd48c9cabbdf5362e5951d4ce86f9516af67599a/src/main/duckdb-extensions.json)).
> 
> Did you need to address this for the flatpak package too?[1](#user-content-fn-1-ebb16d9817343810d598e5a98cfc18e0) IIRC, flatpak has similar requirements for deterministic dependencies.
> 
> ## Footnotes
> 1. I don't see anything obvious in the diff [↩](#user-content-fnref-1-ebb16d9817343810d598e5a98cfc18e0)

I've updated this PR with DuckDB sources SHA256 config & verification. There's a couple of documentation updates needed, but I will hold them off until the final approach is decided (so we avoid unnecessary changes and code review).

There's one open question though (see PR description).

MattSturgeon:
I believe https://github.com/Nexus-Mods/Vortex/pull/23551 will also be an issue, as it is generating the API schema during postinstall. /cc @Aragas 

Removing `src/generated/` from [`packages/nexus-api-v3/.gitignore`](https://github.com/Nexus-Mods/Vortex/blob/1c34e34f970ef31c83d677883f11f030dc598b1f/packages/nexus-api-v3/.gitignore) and having a scheduled "update schema" CI, instead of regenerating every build, is one way to address that.

Also, I wonder if it'd be easier to land this if the DuckDB stuff was split off into a dedicated PR?

danyloburavlov:
@MattSturgeon I can split this into 2 PRs (Flatpak + DuckDB), or even 3 (Flatpak + DuckDB + workflow adjustments after proper review/verification of the DuckDB part). Happy to split it however you prefer.

Also, my bad if I’m missing something, but I don’t see how `generate-query-types.ts` is tied to `postinstall`. As far as I can tell, the new V3 API `codegen` target fetches the Nexus API OpenAPI schema, while `generate-query-types.ts` generates local DuckDB query types from `src/queries`. Is that the part you meant?

MattSturgeon:
> Happy to split it however you prefer.

I'm not part of the Nexus Mods team, so my opinion doesn't really matter, here. Just throwing suggestions out there 🙂 

> I don’t see how `generate-query-types.ts` is tied to `postinstall`.

I was going off of https://github.com/Nexus-Mods/Vortex/blob/1c34e34f970ef31c83d677883f11f030dc598b1f/packages/nexus-api-v3/README.md?plain=1#L11-L14 (I didn't look too closely beyond that).

EDIT: to be clear, this is unrelated to DuckDB. It's just another impure/non-deterministic build step, that I believe would block reproducible packaging (like flatpak).

danyloburavlov:
> I'm not part of the Nexus Mods team, so my opinion doesn't really matter, here. Just throwing suggestions out there 🙂

Me neither, but any reasonable opinion is welcome 🙂

~~As for the doc, yes, it makes sense, but it seems to be targeting just V3 package + Nexus API OpenAPI schema, not DuckDB. IDK, that's what I see, let's hear from maintainers too, as soon as they have a chance to reply.~~

EDIT: Gotcha. Well, sounds like a separate PR for sure.

MattSturgeon:
> I believe #23551 will also be an issue, as it is generating the API schema during postinstall.

Looks like #23698 has addressed this, as the resulting schema is now committed. Not sure whether deterministic builds will still need to explicitly "skip" updating the schema or not.

danyloburavlov:
> > I believe #23551 will also be an issue, as it is generating the API schema during postinstall.
> 
> 
> 
> Looks like #23698 has addressed this, as the resulting schema is now committed. Not sure whether deterministic builds will still need to explicitly "skip" updating the schema or not.

I will look into it a bit later and update the PR if needed. Thanks for the heads-up!

danyloburavlov:
Rebased the PR onto the latest master and re-generated sources.

As for the [#23698](https://github.com/Nexus-Mods/Vortex/pull/23698) - no actions needed so far in this PR.

src/main/download-duckdb-extensions.ts
MattSturgeon, line 1: I don't speak for the NexusMods team, but it _may_ be easier for them to review if the duckdb lock changes are pulled out into a dedicated PR.

Of course, they could review commit-by-commit, but a separate PR may be more focused and less friction.

danyloburavlov: Yes, I agree, but I haven't heard from the NexusMods team. I have no idea if that DuckDB extension is more important than Flatpak and could be shipped separately.

danyloburavlov: @erri120 @IDCs @Aragas do you need DuckDB changes in another PR?

insomnious:
Thanks for the PR. I'm closing this one. We're in the middle of some large foundational changes to Vortex and don't have the review capacity to take on something this size alongside that work. That won't be permanent.

We've updated our contributing guide with what we can realistically take right now ([CONTRIBUTE.md](https://github.com/Nexus-Mods/Vortex/blob/master/CONTRIBUTE.md)). The short version is small, self-contained PRs. Please have a read before opening another. If you want to talk an idea through first, come find us on Discord.

danyloburavlov:
Thanks for the reply.

I understand that the PR was anything but small, and it wasn't against the contributing rules. Still, a couple of times here it was mentioned to split this PR in 2 (or more) smaller ones; I even tagged the contributors, looking for an answer.

And then the PR just gets closed. No replies, no requests to split it in smaller ones. This is a bit frustrating.

PR #23520, opened 2026-06-17
fixed TypographyLink children/customContent as a precedence pair

IDCs:
`children` and `customContent` were typed with `XOr`, making them mutually exclusive. But the component defines a precedence relationship — `customContent` overrides `children` via the `??` fallback; so passing both is valid and is exactly what the "renders customContent instead of children" test exercises. `XOr` made that test fail to typecheck.

Replace the `XOr` wrapper with a plain optional pair (`{ children?: string; customContent?: ReactNode }`), matching the existing precedent in DropdownItem, Pill, Button, etc.


Aragas (approved):

IDCs:
@r1chm8 is already working on fixing the test

PR #23521, opened 2026-06-17
Fixed token refresh causing a loop/flickering

Aragas:
Update User info (e.g. is premium) on token refresh
Added better JWT token parsing
Added @types/jsonwebtoken to renderer

Related Discord Messages:
discord://discord.com/channels/215154001799413770/408252140533055499/1516440692547649547
discord://discord.com/channels/215154001799413770/408252140533055499/1515813866876047401

Closes [LAZ-605](https://linear.app/nexus-mods/issue/LAZ-605/avatar-and-premium-banner-flicker-on-oauth-token-refresh-leading-to)

### DIFF
--- a/src/renderer/src/extensions/nexus_integration/util.ts
+++ b/src/renderer/src/extensions/nexus_integration/util.ts
@@ -72,8 +72,8 @@ import { setUserInfo } from "./actions/persistent";
 import { setLoginId, setOauthPending } from "./actions/session";
 import { OAUTH_CLIENT_ID, OAUTH_REDIRECT_URL, OAUTH_URL, getOAuthRedirectUrl } from "./constants";
 import NXMUrl from "./NXMUrl";
-import { isLoggedIn } from "./selectors";
-import type { IJWTAccessToken } from "./types/IJWTAccessToken";
+import { isLoggedIn, userInfo as userInfoSelector } from "./selectors";
+import { accessTokenSchema } from "./types/IJWTAccessToken";
 import type { IValidateKeyDataV2 } from "./types/IValidateKeyData";
 import { IAccountStatus } from "./types/IValidateKeyData";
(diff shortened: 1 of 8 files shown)


src/renderer/src/extensions/nexus_integration/util.ts
```diff
@@ -1801,15 +1812,67 @@ export function transformUserInfoFromApi(input: IUserInfo & { preferences: IPref
...
+
+  // avoid a needless dispatch (and re-render) when nothing membership-related changed
+  if (
+    existing.isPremium === isPremium &&
+    existing.isSupporter === isSupporter &&
+    existing.isLifetime === isLifetime &&
+    existing.status === status
+  ) {
+    return;
+  }
```
Aragas: Is hardcoding all stuff we check against fine here? Does look like some potential bug in the future

erri120: This entire thing is super brittle since we don't have proper typing.

IDCs: Agreed, a membership type would be ideal. Then you can just compare/spread over its keys.

It's already ignoring two membership states - `isClosed`, `isBanned` (these aren't used anywhere within the codebase so I guess that's why they were left out) - but that's not on this PR as I think we have a similar check in other locations too.

Not sure if this should be blocking though - but a cleanup pass is definitely needed.

Aragas: I removed account status if we're not interested in that atm @IDCs.
@erri120 code should now be reactive to membership changes

Aragas: I won't touch `IValidateKeyDataV2` for the moment, but it's basically the type that exposes User Info, we'll need to change that

IDCs (approved):

PR #23534, opened 2026-06-17
fix getCollectionModByReference modId namespace bug (483 WIP)

IDCs:
The selector had a dead "find by modId" fast-path comparing the caller-supplied Nexus mod id against the installed Vortex mod id (different namespaces, never matched). Replace the hand-rolled per-field matcher with the shared testModReference over an IModLookupInfo, so there is one notion of reference identity and no mod-id confusion.

- getCollectionModByReference(state, lookup: IModLookupInfo); both callers (DownloadView, InstallManager) pass lookupFromDownload(download).
- fix the bundled-mod branch: symmetric basename compare (the old asymmetric one stripped the extension from only one side and never matched a real archive).
- add a makeLookup test-utils builder + selector tests (incl. a regression that a value equal to the installed Vortex mod id does not match).
- regenerate etc/vortex.api.md for the new signature.

This is the first bug found as part of the tests-first approach.

part of LAZ-483

### DIFF
--- a/src/renderer/src/extensions/download_management/views/DownloadView.tsx
+++ b/src/renderer/src/extensions/download_management/views/DownloadView.tsx
@@ -48,6 +48,7 @@ import { truthy } from "../../../util/util";
 import MainPage from "../../../views/MainPage";
 import type { IGameStored } from "../../gamemode_management/types/IGameStored";
 import type { IInstallOptions } from "../../mod_management/types/IInstallOptions";
+import { lookupFromDownload } from "../../mod_management/util/dependencies";
 import { convertGameIdReverse } from "../../nexus_integration/util/convertGameId";
 import { setShowDLDropzone, setShowDLGraph } from "../actions/settings";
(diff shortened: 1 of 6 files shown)


src/renderer/src/extensions/download_management/views/DownloadView.tsx
```diff
@@ -391,17 +392,14 @@ class DownloadView extends ComponentEx<IDownloadViewProps, IComponentState> {
...
           return false;
         }
         return (
-          getCollectionModByReference(this.context.api.store.getState(), {
-            tag,
-            fileId: identifiers.fileId,
-            modId: identifiers.modId,
-          }) != null
+          getCollectionModByReference(
+            this.context.api.store.getState(),
```
Aragas, line 400: Since you did change it, we should prefer `api.getState()`, since it's typed, but not necessarily worth a change in this PR

Aragas (approved):

PR #23536, opened 2026-06-17
Convert X Rebirth to the Game Description Language

halgari:
Switches the X Rebirth extension over to the Game Description Language (GDL) and brings GDL into the repo as a git submodule (`game-description-language/`, tracking `main`).

X Rebirth is now defined by a `game.yaml` plus `src/hooks.ts` (content.xml installer and the three health checks), built via `gdl build` into `dist/index.js`. This is the first GDL conversion; more games will follow.

Also removes the old `game-extension-test` harness, which only X Rebirth used. Its role is now covered by GDL's corpus runner (`gdl test:corpus`), which loads the built bundle and exercises the real installer and health-check paths against cached Nexus manifests.

Verification: GDL suite green, X Rebirth unit tests pass, and the corpus run is 299 matched / 3 unmatched / 0 failed (the unmatched are non-mod uploads).


extensions/games/game-xrebirth/package.json
erri120, line 1: Need to update the `build` nx target config. I recommend adding a new default target to the root `nx.json` file:

```json
"gdl": [
  "{workspaceRoot}/scripts/build-gdl-extension.mjs",
  "{projectRoot}/game.yaml"
]
```

And adding `gdl` to the build inputs:

```json
"build": {
  "inputs": ["default", "typescript", "vortex-api", "gdl"]
}
```

erri120 (changes requested):
We moved away from using submodules for our custom libraries, GDL should be published to NPM as `@nexusmods/game-description-language` or `@nexusmods/gdl`.

halgari:
Closing for now as we're going a different direction with this

PR #23537, opened 2026-06-18
unify event-based collection skip + add InstallDriver orchestration tests (483 WIP)

IDCs:
Collection members (mods and downloads) get skipped/ignored three ways: automatically during install (InstallManager), by a free user declining a queued download (nexus_integration), and manually from the collection page UI. The manual UI path already dispatches the decision directly (setRulesIgnored -> addModRule + resyncCollectionSessionRules) and is unchanged here. The other two used to emit a global event (collection-mod-skipped / free-user-skipped-download) for the InstallDriver to handle, an artifact of collections being a bundled extension. Now that collections is core, those two skip sites dispatch the decision directly too.

- Add util/collectionSkip.ts markCollectionMemberSkipped(api, {reference} | {identifiers}): one entry point that finds the skipped member in the active session and marks it ignored (session status + durable rule flag, the same dual-write the old markRuleIgnored did). InstallManager and nexus_integration call it directly; the driver's two skip listeners and markRuleIgnored are removed.
- Fix the free-user matching that the old inline handler got wrong: the fileNames dereference now happens after its guard (no longer throws when a skip carries no file names), and a fuzzy-version member matched by file name is no longer lost to testRefByIdentifiers' definitive file-id short-circuit. The automatic/premium reference match is unchanged (tag, then fileMD5, then logicalFileName).
- Split InstallDriver.onDidInstallDependencies into isInstallComplete (the completion decision) and finalizeInstalledCollection (the postprocessing side-effect), so the decision is unit-testable without the staging-path / fs infrastructure.
- Add a test harness (makeApiHarness / makeDriverHarness in test-utils/builders) that drives the real InstallDriver through a fake IExtensionApi, plus behaviour-grouped InstallDriver.test.ts and collectionSkip.test.ts (covering the previously-flaky free-user matching).
- Expose IReferenceIdentifiers as a public type and reuse it for both testRefByIdentifiers and the skip identifiers, and drop the now-unemitted collection-mod-skipped from EVENTS.md.

Edit: Also added shared vitest fixtures: harnessTest (makeApi + automatic registry/mock teardown,
so a fake game can't leak between tests) and driverTest (makeDriver, kept separate to keep the
slow driver import out of api-only suites). InstallDriver/collectionSkip tests build through
them; the data builders stay plain functions. [c9d6974e0d7212ab1895c8a3d6a3eb50530fd8fe](https://github.com/Nexus-Mods/Vortex/pull/23537/changes/c9d6974e0d7212ab1895c8a3d6a3eb50530fd8fe)

part of LAZ-483

### DIFF
--- a/src/renderer/src/test-utils/builders.ts
+++ b/src/renderer/src/test-utils/builders.ts
@@ -7,6 +7,15 @@
  * and accepts a partial override - the Test Data Builder pattern - so a test only
  * states the fields it cares about.
  *
+ * The data builders (makeMod, makeRule, makeReference, makeSession, ...) are deliberately PLAIN
+ * FUNCTIONS, not vitest fixtures (test.extend). They are stateless, deterministic, and allocate
+ * nothing that needs teardown, and their whole value is the partial-override args - wrapping them
+ * as fixtures would only add indirection and lose that ergonomics. Fixtures earn their keep for
+ * per-test setup/teardown + laziness, which here applies only to the STATEFUL harness below
+ * (makeApiHarness / makeDriverHarness register a fake game in a worker-global registry that must
+ * be cleared between tests); that lifecycle is wrapped by the harnessTest / driverTest fixtures,
+ * not by these builders.
+ *
  * Test-only: nothing in the production tree imports this module.
  */
 import { EventEmitter } from "events";
--- a/src/renderer/src/test-utils/builders.ts
+++ b/src/renderer/src/test-utils/builders.ts
@@ -9,16 +9,35 @@
  *
  * Test-only: nothing in the production tree imports this module.
  */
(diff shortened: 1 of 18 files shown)


src/renderer/src/test-utils/builders.ts
erri120, line 1: You might want to look at https://vitest.dev/guide/test-context.html#extend-test-context

IDCs: Thanks for the test.extend pointer - added it.

I Left the plain data builders (makeMod, makeRule, ...) as-is though - stateless factories with nothing
to tear down, so fixtures only helped the stateful harness. Appreciate the nudge!

IDCs:
bloody barrel imports

erri120 (approved):

PR #23538, opened 2026-06-18
Suppress more false positive fingerprints being reported

Aragas:
For now just doing specific suppressions, but we might need to convert all `instanceof` calls if we won't find a better alternative


erri120 (changes requested):
I don't think this is the correct approach. `instanceof` works properly inside the same context, it only breaks for cross-context data. The solution isn't to change every `instanceof` check but to find the context boundaries and do proper hydration.

(also you can change how `instanceof` behaves using `static [Symbol.hasInstance](instance)` hackery`)

Aragas:
Good point, going then with the error contianer approach discussed

PR #23541, opened 2026-06-18
Added an error origin tracker to avoid error re-serialization

Aragas:
Each process (main/renderer) can now use a error origin tracker system.
Only renderer enables it for now.
> When `renderer` calls `main` and `main` calls `renderer` (callback), if the callback throws and error, it will be saved in the error origin tracker. Once it passes back to `main` and `renderer`, we will use the original error instead of the serialized one, saving the type information

### DIFF
--- a/src/shared/src/error-serialization.ts
+++ b/src/shared/src/error-serialization.ts
@@ -5,10 +5,11 @@
 // can branch on `err.name` (e.g. via isErrorOfType) rather than `instanceof`.
 //
 // When a context owns the live error (the renderer, for a callback error that
-// round-trips back to it), an optional ErrorOriginTracker passes it by
-// reference instead — the original object is handed back verbatim, preserving
-// identity, prototype and the real throw-site stack. This is the only path that
-// recovers the concrete type across the wire; everything else is a plain Error.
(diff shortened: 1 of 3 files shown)


src/shared/src/error-serialization.ts
```diff
@@ -12,6 +17,47 @@
...
+   * ref tokens so two trackers can't mis-resolve each other's errors.
+   */
+  readonly namespace: string;
+  /** Stash a live error and return a context-local id, or undefined to opt out. */
+  capture(err: Error): string | undefined;
+  /** Return the original for a context-local id (namespace already stripped), if still held. */
+  resolve(id: string): Error | undefined;
+}
+
+let originTracker: ErrorOriginTracker | undefined;
```
erri120: No global state in shared project. Makes it really hard to debug because this gets inlined.

erri120 (changes requested):

src/shared/src/error-serialization.ts
Aragas: Good point, fixed!

erri120 (approved):

Issue #20915, opened 2026-02-23
dotnetprobe fails on Windows 10 IoT Enterprise


wxMichael:
**Your Nexus Mods username**
wxMichael

**Describe the bug**
When a user is running Windows 10 IoT Enterprise, dotnetprobe.exe crashes with this error:
```
PS C:\Program Files\Black Tree Gaming Ltd\Vortex\resources\app.asar.unpacked\assets> .\dotnetprobe.exe
CLR: Assert failure(PID 1008 [0x000003f0], Thread: 7976 [0x1f28]): !AreShadowStacksEnabled() || UseSpecialUserModeApc()
    File: D:\a\_work\1\s\src\coreclr\vm\threads.cpp:7954 Image:
C:\Program Files\Black Tree Gaming Ltd\Vortex\resources\app.asar.unpacked\assets\dotnetprobe.exe
PS C:\Program Files\Black Tree Gaming Ltd\Vortex\resources\app.asar.unpacked\assets>
```
NET 9 itself is installed, but Vortex assumes it's not due to this app not exiting with code 0.

```
C:\Users\Michael>dotnet --list-runtimes
Microsoft.NETCore.App 9.0.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
Microsoft.WindowsDesktop.App 9.0.13 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

**To Reproduce**
1. Install Windows 10 IoT Enterprise in a VM (I have this set up if needed for testing)
2. Open a terminal
3. Run dotnetprobe.exe

**Expected behavior**
If NET is installed and functional on Windows 10 IoT Enterprise, Vortex ought to run.

**Platform (please complete the following information):**
 - OS: Windows 10 IoT Enterprise
 - Vortex 1.16.4

**Additional context**
Discord troubleshooting which led to this report:
https://discord.com/channels/215154001799413770/408252140533055499/1475463731298177170

IDCs:
@Aragas can you please have a look at this one - appears to be specific to that Windows version.

wxMichael:
Another user reported this same error on Windows 10 Pro.

xshadesxprayerx:
dont know if this will help i tried a bit of everything and repairing microsoft.net sdk worked for issue went away completely

StarMinx69:
yeah i had this issue too but a mod from the discord helped me with a "silly workaround" and it worked

[...]

Issue #20780, opened 2026-02-21
.NET 9 not being recognized as installed


DaddyPhantae:
**Your Nexus Mods username**
TheMegaPhantom

**Describe the bug**
During fresh install it was not able to detect .NET 9 even after being installed by vortex
there is no detailed error given

**Expected behavior**
vortex to work

**Platform (please complete the following information):**
OS is Win 10
Vortex is 1.16.4

**Additional context**
no error being given is kinda annoying lol

wxMichael:
I've done a clean install of Windows 10 in a VM and was able to install and launch Vortex 1.16.4.
I also repeated the test with a mix of NET 6 and NET 8 versions installed prior to Vortex and encountered no issues.

I'd suggest completely uninstalling all NET Runtimes, reboot, then run the Vortex installer.
After installing Vortex, if you open Command Prompt and run `dotnet --list-runtimes` you should see only these:
```
Microsoft.NETCore.App 9.0.13 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
Microsoft.WindowsDesktop.App 9.0.13 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

DaddyPhantae:
the prior NET runtimes in the screenshot you saw on discord (im the user they were talking about in there) were only installed in an attempt to get it to work after it failing with just the NET vortex comes with

i also tried to get the runtime from microsofts website and use that to repair the install, as well as running the installer from the vortex install location. i dont really want to have to reinstall vortex and NET but if its my only option so be it

wxMichael:
You don't need to uninstall Vortex or wipe any of its data, just run the latest installer over top of the existing installation.

DaddyPhantae:
reinstalling vortex seems to have fixed it, idk what went wrong originally

DecayingSec:
Fresh install of Windows, .NET 9 is installed. Tried reinstalling Vortex, tried reinstalling .NET 9. Vortex keeps saying it isn't installed.

Turokirill:
<img width="912" height="244" alt="Image" src="https://github.com/user-attachments/assets/256b962f-6d0b-4c5f-a25c-324b88223e71" /> I reinstalled and also executed commands to grant rights to the dotnet folder, everything worked on the old version of vortex with .net 6, decided to update vortex and it just doesn't find .net, although version 9 is worth it, and tried to reinstall with different rights.

[...]

PR #23546, opened 2026-06-20
Disable CET compatibility in dotnetprobe

nerfnet:
Some .NET runtimes (including 9) enable CET by default which can be problematic because it requires a very specific setup to work, which is not guaranteed on all systems(like mine). Setting compat to false allows everyone to run it without a problem.

This should also fix these issues:
#20915
#20780

Kind regards,

nerfnet

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -261,11 +261,11 @@ catalogs:
       specifier: 2.8.0
       version: 2.8.0
     '@nexusmods/fomod-installer-ipc':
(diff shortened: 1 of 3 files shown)


SorryForWhat:
worked, thanks!

Aragas:
The CET error will not only happen with `dotnetprobe`, we also have FOMOD, which is NativeAOT and most likely has CET enabled. I'm checking whether we distribute any other libraries with that flag

erri120 (approved):

PR #23550, opened 2026-06-22
collection completion tweaks + deterministic ref tags + free-user fixes (483 WIP)

IDCs:
- isInstallComplete reads the session's per-member terminal status (O(members)), replacing the O(rules x mods) findModByRef derivation that froze 2000-4000 mod collections; isTerminalMemberStatus shared across driver/selector/InstallManager.
- deterministicReferenceTag: stable tags from identity + install spec (exact keys on fileMD5, fuzzy on the repo data), gated by a referenceTagScheme marker, shortid fallback for legacy collections. O(1) tag index for did-install attribution; install spec stamped on every member so it re-matches its rule.
- freeUserDownloadPosition fixes the dialog's 101/100 overshoot and skip stalls.
- planDependencyErrorRecovery: leave skips/whole-install cancels, requeue while retries remain, else mark failed so a member can't dangle and block completion.
- Churn suite (driver at 2000-4000 members) + unit tests + shared builders.

part of LAZ-483

### DIFF
--- a/src/renderer/src/extensions/collections/types/ICollectionConfig.ts
+++ b/src/renderer/src/extensions/collections/types/ICollectionConfig.ts
@@ -4,6 +4,17 @@ import type { ICollection } from "./ICollection";
 export interface ICollectionConfig {
   recommendNewProfile: boolean;
   excludePluginRules: boolean;
+  // Identifies how this collection's member referenceTags were generated. Deterministic tags can
+  // only be applied to collections authored AFTER that change (already-published collections keep
+  // their random shortid tags and can never be retrofitted), so the scheme must be recorded in the
+  // config NOW to tell the two apart at install time:
+  //   absent          -> legacy: random shortid tags, matched only by the stamped tag.
+  //   "deterministic" -> tags derived from file identity + installSpec (deterministicReferenceTag),
+  //                      stable across re-install/restart and recomputable/indexable by the matcher.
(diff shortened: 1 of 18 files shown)


src/renderer/src/extensions/collections/types/ICollectionConfig.ts
```diff
@@ -4,6 +4,17 @@ import type { ICollection } from "./ICollection";
...
+  // only be applied to collections authored AFTER that change (already-published collections keep
+  // their random shortid tags and can never be retrofitted), so the scheme must be recorded in the
+  // config NOW to tell the two apart at install time:
+  //   absent          -> legacy: random shortid tags, matched only by the stamped tag.
+  //   "deterministic" -> tags derived from file identity + installSpec (deterministicReferenceTag),
+  //                      stable across re-install/restart and recomputable/indexable by the matcher.
+  // Set ONLY by the authoring path (alongside actually generating deterministic tags); the config
+  // defaults must leave it absent so legacy collections that fall back to the defaults are not
+  // mislabelled.
+  referenceTagScheme?: "deterministic";
```
erri120: Should we version this?

`referenceTagScheme?: "v1";`

I wouldn't exactly call MD5 a "stable identity" and we'd likely change how reference tags are constructed later on.

https://www.mscs.dal.ca/~selinger/md5collision/

erri120 (approved):

PR #23568, opened 2026-06-25
LAZ-483: reliable collection install completion

IDCs:
Integration branch for LAZ-483. Merges the full body of work into `master` with a merge commit to preserve the per-stage history.

## Problem
Collection installs visibly stall near the end (9X%) and never show as finished, even though the install completed in the background. Restarting Vortex "fixes" it, so users abandon installs and miss the optional-mods prompt. Root cause: two parallel tracking systems, and the UI trusted the unreliable in-memory `modsEx` cache (byte/count math + transient notifications) instead of the Redux install session.

## Approach
Fold the standalone `collections` extension into the renderer, make the Redux session the single source of truth for progress/completion, delete the `modsEx` cache, and add isolated tests. (Solution 3 from the ticket.)

## What's in this PR
Grouped by theme; stage PRs in parentheses.

**Source of truth (the visible fix)**
- Fold the collections extension and its session tracking into core (#23462, #23498).
- Track the install lifecycle on the Redux session, with InstallManager as the single writer (#23496, #23523).
- Switch progress / completion / optional-mods prompt to the session and remove the `modsEx` cache.

**Correctness & identity**
- Separate mod identity from install-spec (choices / fileList / patches) matching (#23481).
- Stable, deterministic member identifiers so installed mods are not needlessly re-downloaded (#23550).
- Fix `getCollectionModByReference` numeric-vs-text id namespacing (#23534).

**Durability & failure handling**
- Persist skip / failed decisions so completion survives a mid-install restart (#23465).
- Resync-from-reality action and immediate live-ignore (#23532).
- Unify the three skip paths through one place (#23537).
- Harden failure handling: a disk-full marks the member failed with a clear message, a failed download settles the member failed, and an orphaned on-disk archive no longer hangs a re-install (#23567).

**Free-user journey**
- Fix the declined-download path (could strand a member or mismatch an updated file id) and the progress counter overshoot (#23550).

**Performance**
- O(1) load-order rule resolution and mods-table row reuse so large collections install without UI stutter, plus closure/leak fixes (#23558).

**Tests & infra**
- Isolated harness driving the real install orchestration through a controllable fake; session state-machine / selector / reducer contract suites; churn suites at scale (thousands of members, update/downgrade, concurrent collections) (#23512, #23537, #23564).
- De-barrel the collections extension and drop blanket eslint-disables (#23543).
- Dedupe the dependency/rule model (#23510).
- Run bsdiff binary patching on a main-process worker thread (#23564).

## Testing
- Build, lint, typecheck, format green across all packages; collections + install-manager + churn suites pass.
- Manual validation: repeated installs of large collections across machines including a non-gamebryo game (Stardew Valley) - bar reaches 100%, no stall, completion and optional-mods prompts appear without a restart; free-user journey re-tested clean.
- bsdiff patched files verified byte-correct.

closes LAZ-483


.github/workflows/format.yml
erri120, line 1: These should be undone.

.github/workflows/main.yml
erri120, line 1: These should be undone.

erri120 (dismissed):

IDCs:
Won't be merging yet - we're probably going to build a separate artifact from the branch and run it through QA first to ensure functional parity, and that collections no longer stall under any scenario.

.github/workflows/format.yml
erri120: Still needs work.

.github/workflows/main.yml
erri120: Still needs work.

erri120 (approved):

PR #23571, opened 2026-06-25
FOMOD fixes

Aragas:
* Fixed IPC not having the .exe file distributed
* Cleaned up native folder when distributing

Closes LAZ-649

### DIFF
--- a/src/main/electron-builder.config.json
+++ b/src/main/electron-builder.config.json
@@ -60,7 +60,8 @@
     "!node_modules/@nexusmods/fomod-installer-native/src-native/**",
     "!node_modules/@nexusmods/fomod-installer-native/*.{h,lib,dll}",
     "!node_modules/@nexusmods/fomod-installer-native/build/*.{vcxproj,filters}",
-    "!node_modules/@nexusmods/fomod-installer-native/build/Release/*.{exp,iobj,ipdb,obj,lib,pdb}"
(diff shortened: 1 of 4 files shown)


src/main/electron-builder.config.json
```diff
@@ -53,7 +53,13 @@
...
     }
   ],
-  "files": ["**/*"],
+  "files": [
+    "**/*",
+    "!node_modules/@nexusmods/fomod-installer-native/build/**",
+    "!node_modules/@nexusmods/fomod-installer-native/bin/**",
+    "!node_modules/@nexusmods/fomod-installer-native/src-native/**",
+    "!node_modules/@nexusmods/fomod-installer-native/*.{h,lib,dll}"
+  ],
```
erri120, line 65: Don't need this and I really don't more hardcoded paths in configs than necessary.

IDCs (dismissed):

src/main/electron-builder.config.json
Aragas: изображение

Without that we ship redundant files with Native

Aragas: About 20MB of redundant files*

Aragas: We package those files because we're rebuilding the .node file with electron headers IIRC

Aragas: Something is wrong with this. I'll check whether the electron based .node files are shipped and correctly used

erri120 (approved):

PR #23586, opened 2026-06-29
Extract feature flag IPC code into FlagService

erri120:
Allows you to use feature flags from anywhere in the renderer, not just within a React tree.


Al12rs (approved):
Looks good, even if not super familiar with the codebase.

PR #23595, opened 2026-06-30
Add script to simplify E2E artifact downloading

erri120:
Script that massively simplifies downloading and extracting playwright reports from E2E CI runs.

Part of LAZ-648.


IDCs (dismissed):
looks fine but might be worth adding a scripts entry to the e2e package file to allow local invocation?

i.e.
```
"report": "tsx ci-report.ts"
```

IDCs (approved):

PR #23605, opened 2026-07-01
Report crashes when consent is still not available

Aragas:
We have a major gap where if Vortex crashed before we can access the state, where we can see whether analytics is enabled - the error will not be reported.
This is a huge issue, since we are losing critical errors that a user would need to report manually to be resolved.

Closes LAZ-638

### DIFF
--- a/src/renderer/src/util/errorHandling.ts
+++ b/src/renderer/src/util/errorHandling.ts
@@ -50,6 +50,7 @@ export function createErrorReport(
 ) {
   const userData = getVortexPath("userData");
   const reportPath = path.join(userData, "crashinfo.json");
+  const consentGiven = state !== undefined && isTelemetryEnabled(state);
   fs.writeFileSync(
     reportPath,
     JSON.stringify({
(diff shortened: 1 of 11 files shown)


src/renderer/src/util/errorHandling.ts
```diff
@@ -59,9 +60,10 @@ export function createErrorReport(
       reportProcess: process.type,
       sourceProcess,
       userData,
+      consentGiven,
     }),
   );
-  if (state !== undefined && isTelemetryEnabled(state)) {
+  if (consentGiven) {
     spawnSelf(["--report", reportPath]);
```
IDCs, line 67: Is this intentional? looks like it's still blocked behind `isTelemetryEnabled`

IDCs:
```
const consentGiven = state !== undefined && isTelemetryEnabled(state);
```

Aragas: Yea, basically what we are doing is we see whether the crash occurred after the state was initialized (so we can see whether telemetry is enabled or not). This way we can switch to the non-strict reporting if telemetry was enabled

IDCs (approved):

PR #23612, opened 2026-07-02
Fix collection plugins left unsorted or disabled after install (LAZ-497, LAZ-630)

IDCs:
Fixes several distinct causes of a collection install finishing with plugins left **unsorted** (`-1` in the load order column) or **disabled**, and makes plugin loading resilient to a single invalid plugin file.

**Not sorted**
- Durable "sort owed" marker (per-profile, on `persistent.transactions`) set at   install start, cleared only on a successful sort; drains on the next profile activation, so an interrupted install still sorts on next launch. Marker is only recorded for gamebryo games.
- Bound the post-deploy plugin-details wait so switching games mid-deploy can't hang the deployment and block all later sorts (resolves on game-switch cue or 30s timeout).

**Not enabled**
- Match the collection's stored plugin name against the installed file case-insensitively (curator casing is inconsistent).
- Read member rules from the re-read current mod, not a stale snapshot that could predate the `requires` rules.
- Key all `loadOrder` reducer lookups via `toPluginId`, so a mixed-case plugin no longer hits the insert branch and gets `loadOrder` reset to `-1`.

**Invalid plugins (LAZ-630)**
- A plugin LOOT can't parse no longer aborts the whole load order - drop the bad one(s), proceed with the rest, and surface one warning listing them.
- Validate all plugin headers with `ESPFile` in a single parallel pass and exclude invalid ones up front, instead of re-running a full load/sort per bad plugin (which stalled large profiles).

**Testing**
New unit tests (`gamebryo`, `loadOrder`, `invalidPlugins`, `InstallDriver`, `mapWithConcurrency`) plus manual testing against the Gate To Sovngarde collection.

closes LAZ-630
closes LAZ-497


src/renderer/src/util/mapWithConcurrency.ts
erri120, line 1: This should really just use `p-limit` which is already available through the main project: https://npmx.dev/package/p-limit

No need for even more DIY.

erri120 (changes requested):

src/renderer/src/util/mapWithConcurrency.ts
IDCs: Ok - just an FYI - `mapWithConcurrency` was extracted from `linkingDeployment` (it was an inline helper function) and it obviously controls how links get modified/removed/added - changing those is out of scope for this PR, so I'm restoring the inline helper function.

I'll modify the `plugin-management` extension to use `p-limit` though.

erri120 (approved):

PR #23618, opened 2026-07-02
QA-112: E2E test for deploying mods from the Mods list

chnexusmods:
## What

Adds `mods-deploy.spec.ts` — an E2E test for **QA-112 (Mods [9.5] — I can deploy mods from the Mods list)**, covering the manual deploy flow for both **free** and **premium** users:

1. Disable the Interface → Automation *auto-enable* and *auto-deploy* toggles (leaving auto-install on)
2. Download SMAPI (auto-installs into the Mods list, but stays **Disabled**)
3. Manually **enable** the mod
4. **Deploy** it from the Mods list and assert the "Mods deployed" confirmation

### Out of scope
The ticket's "launch the game and verify the menu changed" steps aren't automatable — the fake Stardew Valley fixture has no real launcher or renderer (consistent with `mods-manual.spec.ts`).

## Supporting POM changes

- **`settings.ts`** — `AUTOMATION_LABELS` + `automationToggle()` for the Automation toggles (class-based: the `Toggle` control exposes no role/aria/testid; see the comment for the recommended app-side a11y fix)
- **`modsPage.ts`** — `deployButton`, `playButton`, `statusButton`, `emptyState`, `deployedNotification`
- **`nexusModPage.ts`** — `modManagerDownload` (matches premium's **"Vortex" button** *or* free's **"Mod manager download" link**), `downloadModal`, `modalDownloadLink`
- **`navbar.ts`** — `homeButton` (exit the per-game workspace to reach global Settings)

## Notable fixes discovered while writing this

- **Premium download capture** — premium fires a direct `nxm://` navigation that never lands in the DOM, so the existing JS-hook/DOM-scan capture missed it. Now captured via a network-request listener, with the DOM scan as fallback.
- **HTML-entity-encoded nxm URLs** — decode `&amp;` → `&` before forwarding, otherwise Vortex's `download_link` request 403s.
- **`shell.openExternal` stub** — restored in `setupMainWindow` (dropped in a prior fixture refactor). Without it, login opened the OAuth page in the host's real browser. Fixes this for **all** login-using tests.

## Testing
- Both `free` and `premium` verified green (headless + Playwright UI).
- `tsc`, `eslint`, and `oxfmt` clean.

## Follow-ups (not in this PR)
- `mods.spec.ts` (QA-108) and `mods-manual.spec.ts` (QA-176) likely have the same latent premium breakage (link-only download locator) — left as-is per discussion.

### DIFF
--- a/packages/e2e/src/fixtures/vortex-app.ts
+++ b/packages/e2e/src/fixtures/vortex-app.ts
@@ -131,6 +131,16 @@ async function setupMainWindow(app: ElectronApplication, timeoutMs: number): Pro
     }),
   ]);
 
+  // Prevent Vortex from opening URLs in the host's real browser during tests
+  // (the OAuth login page, "Go Premium", external mod links, etc.). The login
+  // helper reads the OAuth URL from the login dialog and drives its own
+  // Playwright browser instead. All external opens funnel through
+  // shell.openExternal (renderer opn() → "shell:openUrl" IPC → main open.ts),
+  // so stubbing it here covers every path.
+  await app.evaluate(({ shell }) => {
+    shell.openExternal = () => Promise.resolve();
+  });
+
   return mainWindow;
 }
 
--- a/packages/e2e/src/fixtures/vortex-app.ts
+++ b/packages/e2e/src/fixtures/vortex-app.ts
@@ -131,16 +131,6 @@ async function setupMainWindow(app: ElectronApplication, timeoutMs: number): Pro
     }),
   ]);
 
-  // Prevent Vortex from opening URLs in the host's real browser during tests
-  // (the OAuth login page, "Go Premium", external mod links, etc.). The login
-  // helper reads the OAuth URL from the login dialog and drives its own
-  // Playwright browser instead. All external opens funnel through
-  // shell.openExternal (renderer opn() → "shell:openUrl" IPC → main open.ts),
-  // so stubbing it here covers every path.
-  await app.evaluate(({ shell }) => {
-    shell.openExternal = () => Promise.resolve();
-  });
-
   return mainWindow;
 }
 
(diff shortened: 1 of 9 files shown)


RedRanger14:
Feedback is:
Too many comments - I would say only keep the comments for complex things, at the moment there is some comments for selectors which is ott
It looks like this Claude context is missing a lot of the best practice stuff, we've made on e2e is missing?
There are lots bundled into steps when each step should be action + assertion

packages/e2e/src/tests/mods-deploy.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+          // download_link request is malformed and the server returns HTTP 403.
+          const url = (nxmUrl as string).replace(/&amp;/g, "&");
+          await vortexApp.evaluate(({ BrowserWindow }, u) => {
+            const target = BrowserWindow.getAllWindows().find((win) =>
+              win.webContents.getURL().includes("index.html"),
+            );
+            if (target === undefined) throw new Error("Vortex main window not found");
+            target.webContents.send("external-url", u, undefined, false);
+          }, url);
+        });
```
erri120, line 65: This is duplicate from `mods.spec.ts`. Should be refactored out into a helper method to simplify downloading mods in tests.

packages/e2e/src/fixtures/vortex-app.ts
```diff
@@ -131,6 +131,16 @@ async function setupMainWindow(app: ElectronApplication, timeoutMs: number): Pro
...
 
+  // Prevent Vortex from opening URLs in the host's real browser during tests
+  // (the OAuth login page, "Go Premium", external mod links, etc.). The login
+  // helper reads the OAuth URL from the login dialog and drives its own
+  // Playwright browser instead. All external opens funnel through
+  // shell.openExternal (renderer opn() → "shell:openUrl" IPC → main open.ts),
+  // so stubbing it here covers every path.
+  await app.evaluate(({ shell }) => {
+    shell.openExternal = () => Promise.resolve();
+  });
```
erri120: I removed this recently because global stubs are an anti-pattern. Tests should stub and cover these APIs on their own already.

erri120 (changes requested):

packages/e2e/src/fixtures/vortex-app.ts
chnexusmods: ok thanks I'll rework it

packages/e2e/src/tests/mods-deploy.spec.ts
chnexusmods: @erri120 I have updated this now

packages/e2e/src/fixtures/vortex-app.ts
chnexusmods: @erri120 I have updated this now

erri120 (approved):

PR #23620, opened 2026-07-02
feat: add Ready or Not game extension with variant conflict detection

rolling-codes:
## Summary

Adds a new game extension for **Ready Or Not** at `extensions/games/game-readyornot/`.

- Game discovery via Steam (`1144200`) and Epic Games Store
- PAK mod installation to `ReadyOrNot/Content/Paks/~mods`
- FMOD audio mod support (`.bank` files)
- Load order management using the alphanumeric prefix system (matches existing community extension behaviour)
- **Variant conflict detection** — warns the user when mutually exclusive mod variants are enabled at the same time

### Variant conflict detection

Vortex's existing file-level conflict detection doesn't catch a common modding problem: mods that are *variants* of the same feature and are never meant to coexist, but don't necessarily write to the same files.

This happens in two ways in the Ready or Not mod ecosystem:

**Intra-mod variants** — a single Nexus mod page ships multiple exclusive files (e.g. *No Stress for SWAT* ships Standard, Hardcore Officers, and Robocop variants). Installing two of them won't trigger a file conflict, but both will try to take effect in-game.

**Cross-mod conflicts** — multiple separate Nexus mod pages all modify the same gameplay system (e.g. the many player HP mods on Nexus — `350HP Mod`, `1000HP Mod`, `HP PLUS`, etc.). Installing two silently means only one wins.

The extension hooks into `profile-did-change` and `mods-enabled` events and shows a Vortex warning notification naming the specific conflicting mods.

The detection logic is split into two pure modules with no Vortex API dependency, making them straightforward to extend or unit test:

- `src/variantGroups.js` — curated list of exclusive variant groups (intra-mod and cross-mod)
- `src/conflictDetector.js` — stateless `detectConflicts(mods, groups)` function

### Notes

- Game registration code adapted from the community extension by [BeYkeRYkt](https://github.com/BeYkeRYkt/vortex_readyornot_extension) — credit included in source
- `assets/gameart.jpg` still needs to be added (binary file — happy to push separately if this PR approach is preferred)
- Variant group data in `variantGroups.js` can be extended by the community as new conflicting mod families are identified

### DIFF
--- a/extensions/games/game-readyornot/package.json
+++ b/extensions/games/game-readyornot/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "game-readyornot",
+  "version": "1.0.0",
+  "description": "Support for Ready Or Not",
+  "author": "Nexus Mods",
+  "license": "GPL-3.0",
+  "type": "commonjs",
+  "private": true,
+  "config": {
+    "game": "Ready Or Not"
+  },
+  "scripts": {
+    "_assets": "copyfiles -u 1 -f ./assets/* dist",
+    "build": "copyfiles -u 1 -f ./src/*.js dist && pnpm run _assets && pnpm extractInfo"
+  },
+  "devDependencies": {
+    "@nexusmods/vortex-api": "workspace:*"
+  },
+  "nx": {
+    "tags": [
+      "vortex:extension"
+    ],
+    "targets": {
+      "build": {
+        "inputs": [
+          "default",
+          "typescript",
+          "vortex-api"
+        ]
+      }
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


rolling-codes:
@claude

IDCs:
Hey @rolling-codes - I'm afraid this is not the correct process to submit a community game extension. We do not accept contributions into the bundled games repository unless discussed first.

In this case, there is already a community game extension available on the site as you noted - if the current implementation falls short, why not contribute to [BeYkeRYkt](https://github.com/BeYkeRYkt/vortex_readyornot_extension) repo directly?

PR #23630, opened 2026-07-06
LAZ-692: restore checkpointless downloads instead of failing to resume

IDCs:
`resume-download` throws `No checkpoint stored` for a paused download with no usable checkpoint (missing, or  completedRanges: []`), so collection installs and the Downloads page retry forever.

**Changed:**
- `IPCDownloadAdapter` restarts such a download under the same id instead of throwing: re-resolves the source (rebuilding the nxm url from stored nexus ids when `urls` were wiped), clears the partial/checkpoint, and starts fresh via a
new optional `downloadId` on `download:start`. `allowInstall` from the resume options is honored so the installer owns the install, and the poll no longer wipes stored urls.

- Consolidate nxm link construction into NXMUrl builders

**Testing**
Unit tests cover restore (missing/empty checkpoint), url reconstruction, `allowInstall`, start-failure, and the unchanged resume; full
renderer suite green. Dev-verified: pause-before-first-chunk resume, collection dependency, and checkpointed resume.

A refactor of the tests is required when merged into master to avoid the currently in-line harness.

Fixes LAZ-692

### DIFF
--- a/src/renderer/src/IPCDownloadAdapter.ts
+++ b/src/renderer/src/IPCDownloadAdapter.ts
@@ -41,6 +41,7 @@ import { downloadPathForGame } from "./extensions/download_management/selectors"
 import type { IDownload } from "./extensions/download_management/types/IDownload";
 import { knownGames } from "./extensions/gamemode_management/selectors";
 import type { IGameStored } from "./extensions/gamemode_management/types/IGameStored";
+import { nxmUrlFromDownload } from "./extensions/nexus_integration/NXMUrl";
 import { nexusIdsFromDownloadId } from "./extensions/nexus_integration/selectors";
 import { convertGameIdReverse } from "./extensions/nexus_integration/util/convertGameId";
 import { makeModAndFileUIDs } from "./extensions/nexus_integration/util/UIDs";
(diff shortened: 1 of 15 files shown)


src/renderer/src/IPCDownloadAdapter.ts
```diff
@@ -687,6 +736,112 @@ export class IPCDownloadAdapter {
...
+      if (revisionNumber === undefined) return undefined;
+      return `nxm://${ids.gameDomainName}/collections/${ids.collectionSlug}/revisions/${revisionNumber}`;
+    }
+
+    if (ids.modId !== undefined && ids.fileId !== undefined) {
+      return `nxm://${ids.gameDomainName}/mods/${ids.modId}/files/${ids.fileId}`;
+    }
+
+    return undefined;
+  }
```
erri120: This is like the 5th implementation of a method that constructs an `nxm` link that I've seen.

erri120 (dismissed):
I don't like that the file is now at 1000 lines but I guess that's what it takes to interface with Vortex weirdness.

src/renderer/src/IPCDownloadAdapter.ts
IDCs: I'll see if I can extract it into a shared utility function.

erri120 (approved):

PR #23635, opened 2026-07-06
Better diagnostics for download errors

Aragas:
We still don't know the roo cases, lets make sure the code is being passed
Closes LAZ-697

### DIFF
--- a/src/shared/src/types/errors.ts
+++ b/src/shared/src/types/errors.ts
@@ -6,7 +6,7 @@ export type DownloadErrorPayload =
   | { code: "precondition-failed"; url: URL }
   | { code: "protocol-violation"; url: URL }
   | { code: "is-html"; url: URL }
-  | { code: "fs-error"; path: string }
+  | { code: "fs-error"; path: string; errno?: string }
   | { code: "resolver-error" };
 
 export class DownloadError extends Error {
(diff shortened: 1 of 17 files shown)


src/shared/src/types/errors.ts
```diff
@@ -6,7 +6,7 @@ export type DownloadErrorPayload =
   | { code: "precondition-failed"; url: URL }
   | { code: "protocol-violation"; url: URL }
   | { code: "is-html"; url: URL }
-  | { code: "fs-error"; path: string }
+  | { code: "fs-error"; path: string; errno?: string }
```
erri120: I really despise this. Node's is incredibly inconsistent when it comes to error data and the error codes it provides are terrible. This is why I went to great lengths for the new FS abstraction to use proper meaningful errors:

https://github.com/Nexus-Mods/Vortex/blob/3b483b6f9153d6729156c3d6cfc7f30ad61f240c/src/main/src/filesystem/backend.ts#L31-L91

Aragas: We would need to add support for `FileSystemError` inside the downloader, because right not it's not properly setup. It seems to be a way bigger PR comparing to this. I mainly wanted to get diagnostics immediately to see whether we have any errors that should be handled.

So the question is, do you want to properly add support for `FileSystemError` instead of doing this diagnostics PR? I'm fine with going the proper route here, this is not a critical thing AFAIK

Aragas: @erri120 I did exactly that, what's you opinion on this approach?

PR #23636, opened 2026-07-06
Fixed error reporting for transient errors

Aragas:
Also fixed the inverted condition that was hiding real errors
Closes LAZ-699

### DIFF
--- a/src/renderer/src/extensions/nexus_integration/util/tracking.tsx
+++ b/src/renderer/src/extensions/nexus_integration/util/tracking.tsx
@@ -16,6 +16,20 @@ import { truthy } from "../../../util/util";
 import { getGame } from "../../gamemode_management/util/getGame";
 import { nexusGameId } from "./convertGameId";
 
+/**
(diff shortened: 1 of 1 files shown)


src/renderer/src/extensions/nexus_integration/util/tracking.tsx
```diff
@@ -16,6 +16,20 @@ import { truthy } from "../../../util/util";
 import { getGame } from "../../gamemode_management/util/getGame";
 import { nexusGameId } from "./convertGameId";
 
+/**
+ * Rate-limit, timeout, protocol and Nexus API errors are expected/transient
+ * conditions, not Vortex bugs, so they must not be surfaced with a "report"
+ * button (nor recorded as actionable errors in telemetry).
+ */
+function allowReportTrackingError(err: unknown): boolean {
+  return (
```
IDCs, line 25: Could theoretically be a shared utility func - I could've swore we had one 🤔

IDCs (approved):

PR #23652, opened 2026-07-08
Fixed loot top bar overflow

Aragas:
изображение

Also fixed nx cache invalidation

Closes LAZ-238

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -36,7 +36,8 @@
       "inputs": [
         "default",
         "typescript",
-        "!{projectRoot}/src/**/*.test.{ts,tsx}",
+        "{projectRoot}/**/*",
+        "!{projectRoot}/**/*.test.{ts,tsx}",
         { "dependentTasksOutputFiles": "**/dist/**" }
       ],
       "outputs": ["{projectRoot}/dist/"]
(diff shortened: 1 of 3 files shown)


nx.json
```diff
@@ -36,7 +36,8 @@
       "inputs": [
         "default",
         "typescript",
-        "!{projectRoot}/src/**/*.test.{ts,tsx}",
+        "{projectRoot}/**/*",
```
erri120: Two things:

1) This doesn't actually fix the issue you had. Assuming you had a cache issue with the SCSS file that also got modified in this PR then this change doesn't actually fix that. This is because `extensions/gamebryo-plugin-management/package.json` overwrites the inputs for the `build` task. The cache got invalidated because `nx.json` changed, not because the issue got fixed.
2) Including every file defeats the purpose of caching on only selective files.

The "proper" solution is to add a new named input `stylesheets` that includes `{projectRoot}/*.{scss,css}` and adding it as an input to `extensions/gamebryo-plugin-management/package.json`.

erri120 (changes requested):

nx.json
Aragas: Yea, sorry for that. Claude was telling me that this is an idiomatic solution, but I was _really_ sceptical here, so I wanted you to double check that

erri120 (approved):

PR #23657, opened 2026-07-09
Don't display `__vortex_tmp_` if possible, use the friendly name instead

Aragas:
Fixed download not being deleteable because it was not properly cancelled

Closes LAZ-719

### DIFF
--- a/src/renderer/src/extensions/download_management/util/downloadNames.ts
+++ b/src/renderer/src/extensions/download_management/util/downloadNames.ts
@@ -19,9 +19,6 @@ export function nameFromUrl(input: string | undefined): string | undefined {
 
   try {
     const pathname = new URL(input).pathname;
-    if (!truthy(pathname)) {
-      return undefined;
-    }
     return decodeURI(path.basename(pathname));
   } catch (err) {
     return undefined;
--- a/src/renderer/src/extensions/download_management/util/downloadNames.ts
+++ b/src/renderer/src/extensions/download_management/util/downloadNames.ts
@@ -35,7 +35,7 @@ export function nameFromUrl(input: string | undefined): string | undefined {
  */
 export function friendlyDownloadName(download: IDownload): string | undefined {
   const localPath = download.localPath;
-  if (truthy(localPath) && !isTempDownloadName(localPath)) {
+  if (localPath && !isTempDownloadName(localPath)) {
     return localPath;
   }
 
--- a/src/renderer/src/extensions/download_management/util/downloadNames.ts
+++ b/src/renderer/src/extensions/download_management/util/downloadNames.ts
@@ -43,7 +43,7 @@ export function friendlyDownloadName(download: IDownload): string | undefined {
     download.modInfo?.nexus?.fileInfo?.name,
     nameFromUrl(download.urls?.[0]),
   ];
-  const friendly = candidates.find(truthy);
+  const friendly = candidates.find(Boolean);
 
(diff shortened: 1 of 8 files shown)


src/renderer/src/extensions/download_management/util/downloadNames.ts
```diff
@@ -0,0 +1,53 @@
+import * as path from "path";
+
+import { truthy } from "../../../util/util";
```
erri120: 🔫😃 no

src/renderer/src/extensions/download_management/util/downloadNames.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  const candidates: Array<string | undefined> = [
+    download.modInfo?.name,
+    download.modInfo?.meta?.logicalFileName,
+    download.modInfo?.meta?.fileName,
+    download.modInfo?.nexus?.fileInfo?.name,
+    nameFromUrl(download.urls?.[0]),
+  ];
+  const friendly = candidates.find(truthy);
```
erri120:
```suggestion
  const friendly = candidates.find(Boolean);
```

src/renderer/src/extensions/download_management/util/downloadNames.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (input === undefined) {
+    return undefined;
+  }
+
+  try {
+    const pathname = new URL(input).pathname;
+    if (!truthy(pathname)) {
+      return undefined;
+    }
+    return decodeURI(path.basename(pathname));
```
erri120, line 21:
```suggestion
    const pathname = new URL(input).pathname;
    return decodeURI(path.basename(pathname));
```

src/renderer/src/extensions/download_management/util/downloadNames.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/**
+ * Name to show for a download. Prefers the real on-disk name; while the temp
+ * placeholder is still in use, falls back to metadata or the URL so the user
+ * doesn't see __vortex_tmp_*.
+ */
+export function friendlyDownloadName(download: IDownload): string | undefined {
+  const localPath = download.localPath;
+  if (truthy(localPath) && !isTempDownloadName(localPath)) {
```
erri120:
```suggestion
  if (localPath && !isTempDownloadName(localPath)) {
```

erri120 (changes requested):

src/renderer/src/extensions/download_management/util/downloadNames.ts
Aragas: I kid you not, I had a dilemma about "what looks better, `const friendly = candidates.find(Boolean);` or the `truthy`" one. I did pick truthy because I wanted to double check whether `Boolean` is the better approach if if there's something entirely different
I do remember that truthy was banned from new code, xD

src/renderer/src/extensions/download_management/util/downloadNames.ts
Aragas: Agree, that doesn't make sense since it will be throwing anyway

src/renderer/src/extensions/download_management/util/downloadNames.ts
Aragas: Should have judged better, fixed

erri120 (dismissed):

erri120 (approved):

PR #23658, opened 2026-07-09
Fetch adult content flag from downloads

erri120:
Closes LAZ-681.

Couldn't find a mod to test this.

### DIFF
--- a/src/renderer/src/extensions/health_check/utils/fileRequirements/installedFiles.ts
+++ b/src/renderer/src/extensions/health_check/utils/fileRequirements/installedFiles.ts
@@ -222,17 +222,14 @@ export function makeDownloadedFileHydrator(
 }
 
 /**
- * Build the display shape for one installed file from its Vortex mod. The
- * adult-content flag isn't part of a mod's own attributes, but its originating
- * download (linked via `archiveId`) carries the same Nexus modInfo used by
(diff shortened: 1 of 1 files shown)


Aragas (approved):

Aragas:
Something like [this](https://www.nexusmods.com/skyrimspecialedition/mods/184803)? Looks only partially NSFW, hahaha!

Aragas (changes requested):
Sec

src/renderer/src/extensions/health_check/utils/fileRequirements/installedFiles.ts
```diff
@@ -260,6 +267,7 @@ export function makeInstalledFileHydrator(
   const state = api.getState();
   const gameId = activeProfile(state)?.gameId;
   const mods = gameId ? (state.persistent.mods[gameId] ?? {}) : {};
+  const downloads = state.persistent.downloads?.files ?? {};
```
Aragas, line 267: Do we need to pass the whole downloads here? Since we have the mod object lower, we could just calculate isAdult and pass at as a parameter?

Aragas: Wrong comment line, meant 282

Aragas (approved):

PR #23671, opened 2026-07-13
test: QA-246 mod uninstall e2e tests (two-step and one-step)

RedRanger14:
Linear ticket: https://linear.app/nexus-mods/issue/QA-246/mods-97-i-can-uninstall-a-mod
Covers Mods [9.7]: remove dialog checkbox defaults, cancel path, uninstall keeping the archive, archive-only removal, and combined remove+delete-archive. Adds ConfirmRemovalDialog POM and row-scoped ModsPage locators.

### DIFF
--- a/packages/e2e/src/tests/mods-uninstall.spec.ts
+++ b/packages/e2e/src/tests/mods-uninstall.spec.ts
@@ -1,0 +1,168 @@
+import type { ElectronApplication, Page } from "@playwright/test";
+
+/**
+ * QA-246: Mods [9.7] — I can uninstall a mod.
+ *
+ * Scenario 1 (two steps): remove the mod (archive kept), then remove again to
+ * delete the archive — including a Cancel pass to prove nothing is removed.
+ * Scenario 2 (one step): tick "Delete Archive" alongside "Remove Mod" and
+ * remove both at once.
+ *
+ * Setup installs SMAPI first (the target mod depends on it), then the target
+ * mod. The ticket's "verify the mod changes the main menu" step is skipped:
+ * the fake-game fixture has no real launcher, so install/uninstall state is
+ * asserted via the Mods list instead (same approach as QA-176).
+ *
+ * Uninstall behaviour is tier-independent (tier only affects the download
+ * interstitial, covered by QA-108), so this runs as freeUser only.
+ */
+import { test, expect } from "../fixtures/vortex-app";
+import { downloadModViaModManager } from "../helpers/modDownload";
+import { Timeouts } from "../helpers/timeouts";
+import { freeUser } from "../helpers/users";
+import { ModsPage } from "../selectors/modsPage";
+import { NavBar } from "../selectors/navbar";
+import { ConfirmRemovalDialog } from "../selectors/removeDialog";
+
+const SMAPI_MOD_URL = "https://www.nexusmods.com/stardewvalley/mods/2400";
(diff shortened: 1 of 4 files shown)


packages/e2e/src/tests/mods-uninstall.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+  await test.step("Open the Mods page", async () => {
+    const navbar = new NavBar(vortexWindow);
+    await navbar.modsLink.click();
+    const modsPage = new ModsPage(vortexWindow);
+    await expect(modsPage.row(SMAPI_NAME)).toBeVisible({ timeout: Timeouts.NETWORK });
+  });
+
+  await test.step("The target mod is installed and enabled", async () => {
+    const modsPage = new ModsPage(vortexWindow);
+    await expect(modsPage.statusButtonInRow(TARGET_MOD_NAME)).toHaveText(/enabled/i, {
```
Marcin-NM: Does it have to be a regexp? We could keep those statuses in POM too.

packages/e2e/src/tests/mods-uninstall.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      await expect(dialog.deleteArchiveCheckbox).not.toBeChecked();
+    });
+
+    await test.step("Cancel closes the dialog", async () => {
+      await dialog.cancelButton.click();
+      await expect(dialog.root).toBeHidden();
+    });
+
+    await test.step("The mod is still installed and enabled", async () => {
+      await expect(modsPage.statusButtonInRow(TARGET_MOD_NAME)).toHaveText(/enabled/i);
```
Marcin-NM: same^

packages/e2e/src/tests/mods-uninstall.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      await expect(dialog.root).toBeVisible();
+    });
+
+    await test.step("Remove the mod, keeping the archive", async () => {
+      await dialog.removeButton.click();
+      await expect(dialog.root).toBeHidden();
+    });
+
+    await test.step("The mod is uninstalled but its archive remains", async () => {
+      await expect(modsPage.statusButtonInRow(TARGET_MOD_NAME)).toHaveText(/uninstalled/i, {
```
Marcin-NM: same^

Marcin-NM: or even there could be a helper that asserts status?

packages/e2e/src/tests/mods-uninstall.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    await test.step("Remove the mod and its archive", async () => {
+      await dialog.removeButton.click();
+      await expect(dialog.root).toBeHidden();
+    });
+
+    await test.step("The mod is gone from the Mods list", async () => {
+      await expect(modsPage.row(TARGET_MOD_NAME)).toBeHidden({ timeout: Timeouts.NETWORK });
+    });
```
Marcin-NM, line 166: There's no assertion for archives being removed?

Marcin-NM (changes requested):

packages/e2e/src/tests/mods-uninstall.spec.ts
RedRanger14: That is an assertion

Marcin-NM (approved):

PR #23672, opened 2026-07-13
remove dead cross-process IPC from notifications

IDCs:
Drops the electron import so unit tests no longer depend on the electron binary being installed (was failing InstallDriver.test.ts on CI at import time).


erri120 (approved):

IDCs:
Don't care for CodeQL

PR #23697, opened 2026-07-15
test: QA-103 manually set game location to manage a game

RedRanger14:
Covers Games [#8.8]: unmanaged game listed, 'Game not discovered' dialog shown, manual folder selection accepted, game becomes active and listed as Managed, and a mod installs for it. Extracts the showOpenDialog stub into helpers/dialogs and adds Managed/Unmanaged section and dialog locators to the GamesPage POM. https://linear.app/nexus-mods/issue/QA-103/games-88-manually-set-a-games-location-to-successfully-manage-it

Store selection is skipped by the app itself under VORTEX_E2E, and no UAC prompt occurs for user-writable fake game dirs.


Marcin-NM (dismissed):

tools/dotnetprobe/nuget.config
```diff
@@ -0,0 +1,13 @@
...
+  machine-level NuGet configuration. Machines with no configured sources
+  (seen on dev machines and CI runners) otherwise fail restore with NU1100,
+  which used to silently ship a Vortex build without dotnetprobe.
+-->
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+</configuration>
```
Aragas, line 13: What was the issue encountered? I don't remember having a need for a machine wide configuration, but I don't mind this

packages/e2e/src/tests/game-management.spec.ts
```diff
@@ -55,3 +62,103 @@ test.describe("Game Management", () => {
...
+      fakeGame = undefined;
+    }
+  });
+
+  test("[QA-103] user can manually set a game location to manage it", async ({
+    vortexApp,
+    vortexWindow,
+    nexusPage,
+  }) => {
+    fakeGame = setupFakeGame("stardewvalley");
```
IDCs: There's significant overlap here with `manageGame` (`games.ts`) - might be worth extracting the common code and re-use here and in `manageGame`. They're already diverging, so if we change something in the UI you'll have to modify the loccators in both files.

IDCs (changes requested):
Functionally these changes are fine - but they're partially duplicated code and will be a maintenance nightmare going forward if we keep duplicating and diverging for each test we write.

packages/e2e/src/tests/game-management.spec.ts
RedRanger14: Reworked, QA-103 now calls manageGame directly, and the dialog assertions moved into the helper so the fixture path and the test share one implementation. The test only keeps its ticket-specific assertions.

tools/dotnetprobe/nuget.config
RedRanger14: From what I could tell, the failure was NU1100 on restore. My machine had no NuGet sources configured at all (dotnet nuget list source showed "No sources found"), so the restore failed, and because the build script swallowed the error, Vortex got built without the probe. I suspect the platform-cluster runner might have the same issue but I haven't been able to confirm that. I'm honestly not sure how my machine ended up with no sources, so this felt like the safest way to make the probe build not depend on machine config. It should only apply to dotnetprobe, so other .NET builds shouldn't be affected, but happy to change the approach if you'd rather handle it differently.

IDCs (dismissed):

tools/dotnetprobe/nuget.config
IDCs: I think the platform-cluster runner is an intentionally minimal ubuntu runner which never provisions .NET SDK; this obviously means that workflow will always fail until the runner gets updated to include the SDK OR we update the e2e workflow to provision .NET SDK ourselves on linux environments.

My suggestion is to have it provisioned on the runner rather than us modifying this workflow and porbably others in the future whenever that runner is being used with .NET builds.

IDCs (approved):

PR #23699, opened 2026-07-15
improve load order page performance for large mod counts (LAZ-717)

IDCs:
Virtualize the file-based load order list (via a standalone ListWindow helper) and cut the per-drag work that made dragging unusable with 1000+ entries: stable react-dnd collector, memoized rows, precomputed row position/locked-count, id-indexed UpdateSet lookups.

Debounce the BG3 auto-export and heal stale pak-cache mod links so managed mods stop showing as "Not managed by Vortex". Add an opt-in uniformRowHeight flag for custom-renderer games to virtualize.

closes LAZ-717

### DIFF
--- a/etc/vortex.api.md
+++ b/etc/vortex.api.md
@@ -2867,7 +2867,11 @@ interface IItemRendererProps {
     // (undocumented)
     invalidEntries?: IInvalidResult[];
     // (undocumented)
+    lockedEntriesCount?: number;
+    // (undocumented)
     loEntry: ILoadOrderEntry_2;
+    // (undocumented)
+    position?: number;
     // (undocumented)
     setRef?: (ref: any) => void;
 }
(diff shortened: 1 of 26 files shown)


erri120:
Might be too difficult for a virtualized list but would be great to see a component test for that.

IDCs:
A full drag-through-a-virtualized-list component test is awkward, so I split the windowing and reorder logic to small pure modules that test cleanly without fighting jsdom layout, and there's a component test for the mounting behaviour.

ListWindow - visible-range math, overscan, spacer padding, and the drag-clamp that keeps a dragged row rendered while it's scrolled outside the window (so its drag source doesn't unmount mid-drag).
dragReorder - the move/reorder logic and the hover midpoint guard.
DraggableList.test.tsx - asserts a 1000-item virtualized list only mounts a small window of rows, and every row when virtualization is off.

I stopped short of driving an actual DnD in a component test, but react-dnd's HTML5 backend needs real layout and pointer events that happy-dom doesn't provide, so a faithful end-to-end reorder belongs in a Playwright E2E rather than here I think.

erri120 (approved):

PR #23725, opened 2026-07-20
fire `app_game_exited` when the game closes, not when the loader exits (LAZ-803)

IDCs:
For a launch that hands off to the game through a script extender or mod loader (SKSE, F4SE, SMAPI), the launched loader exits within a second or two of spawning the game, so app_game_exited reported a few-second session while the game was still running.

Such launches now watch the game's own executable in session.base.toolsRunning (which ProcessMonitor already tracks) rather than the loader, and emit app_game_exited when the game process ends, with a duration reflecting real play time. A launch counts as running the game when it is the game itself, a defaultPrimary launcher, or the user's configured primary tool; a plain utility tool keeps watching its own process. When the game process never appears, a bounded timeout emits a best-effort exit.

closes LAZ-803

### DIFF
--- a/src/renderer/src/reducers/session.ts
+++ b/src/renderer/src/reducers/session.ts
@@ -2,6 +2,7 @@ import * as path from "path";
 
 import * as actions from "../actions/session";
 import type { IReducerSpec } from "../types/IExtensionContext";
+import type { IRunningTool } from "../types/IState";
 import { addUniqueSafe, deleteOrNop, pushSafe, removeValue, setSafe } from "../util/storeHelper";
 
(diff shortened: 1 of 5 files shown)


Aragas (dismissed):

Aragas:
The Modern UI (Tool bar?) has a similar code, we track whether the game is running or not and show a spinner. Perhaps the codebase can be combined?

IDCs:
I partially consolidated the functionality - but keep in mind that the existing functionality is a react hook which doesn't match what the mixpanel gameanalytics functionality is tackling.

Aragas (approved):

PR #23728, opened 2026-07-20
API events rework

erri120:
Reworks how API events are typed:

- Central event map that can be enhanced with `declare module` syntax
- Compatibility for un-typed events defaulting to `unknown`

Reworks internal event dispatching, fixing many many hidden bugs.

**Important**: This has a breaking change, instead of `PromiseBB` the methods `emitAndAwait` and `withPrePost` now return `Promise`. Any consumers using Bluebird specific methods like `.tap` will get a runtime error.


src/renderer/src/extensions/download_management/index.ts
```diff
@@ -53,6 +53,12 @@ import Settings from "./views/Settings";
 import ShutdownButton from "./views/ShutdownButton";
 import SpeedOMeter from "./views/SpeedOMeter";
 
+declare module "../../types/IExtensionContext" {
+  interface ApiEvents {
+    "set-download-games": (dlId: string, gameIds: string[], fromMetadata?: boolean) => void;
+  }
+}
+
```
Aragas, line 61: Why do we need the module declaration? Because we're not using the `vortex-api` package?

Aragas: Oh, this is an extension event that download_management provides?

erri120: The event comes from this extension, with declare module syntax it's clear who owns the event while still making it available throughout the codebase.

Aragas (dismissed):
That's an interesting approach!

IDCs (dismissed):

Aragas (approved):

PR #23731, opened 2026-07-20
restore downloads from mega.nz / Google Drive (LAZ-808)

### LINKED ISSUE #23723: Restore ability to download from mega
**Your Nexus Mods username**
madtisa

**Describe the bug**
Somewhere between v2.0.1 and the latest version (currently v2.3.0), vortex lost ability to download files from mega

**To Reproduce**

1. Open vortex > Downloads > Click to enter URL
2. Paste https://mega.nz/file/Gw4AACqJ#6O_OiA5n7RBTJI8D6nJoOMXdf2dbpQZ6fis9g0esGqA or any other mega url and click `Download`
3. Wait for timeout (Request timed out)

Image

**Expected behavior**
Opened embedded browser page to download from mega

**Logs**

**Screenshots**

- OS: Windows 10
- Vortex Version: 2.3.0
- Game: Karryn's Prison or Noita

**Additional context**
Not specific to any game extensions

IDCs:
Pasted download URLs are routed through the embedded browser. When the browser hands back a blob (the file it has already saved to the temp folder), the IPC download adapter adopts it as a finished download: it moves the file into the game's download folder under a free name, hashes it, and registers it, taking the same completion path as a network download so it emits did-finish-download, fires the start-download callback, and triggers install.

This makes collection installs and manual downloads work for hosts that deliver files through client-side JavaScript rather than a fetchable URL (mega.nz, Google Drive, and similar), covering manual pastes, collection browse dependencies, and community extensions alike.

Sorry @erri120 - tried not to inflate the adapter any further but it's the only location that caters for all blob style downloads regardless of path.

closes LAZ-808

### DIFF
--- a/src/renderer/src/extensions/download_management/util/downloadNames.ts
+++ b/src/renderer/src/extensions/download_management/util/downloadNames.ts
@@ -50,22 +50,18 @@ export function friendlyDownloadName(download: IDownload): string | undefined {
 }
 
 /**
- * Returns fileName, or the first "name.n.ext" variant not already present in dlPath, so an existing
- * download is never overwritten.
+ * Returns fileName if it is free in dlPath, otherwise a timestamped variant ("name.<ms>.ext"), so an
+ * existing download is never overwritten.
  */
 export async function freeDownloadName(dlPath: string, fileName: string): Promise<string> {
+  const taken = await access(path.join(dlPath, fileName)).then(
+    () => true,
+    () => false,
+  );
+  if (!taken) {
+    return fileName;
+  }
   const ext = path.extname(fileName);
   const base = path.basename(fileName, ext);
-  let candidate = fileName;
-  let counter = 0;
-  while (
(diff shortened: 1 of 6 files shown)


src/renderer/src/extensions/download_management/util/downloadNames.ts
```diff
@@ -47,3 +48,24 @@ export function friendlyDownloadName(download: IDownload): string | undefined {
...
+    await access(path.join(dlPath, candidate)).then(
+      () => true,
+      () => false,
+    )
+  ) {
+    counter += 1;
+    candidate = `${base}.${counter}${ext}`;
+  }
+  return candidate;
+}
```
erri120, line 67: I'd prefer we didn't do an async FS lookup in a while loop. Instead of counting numbers, just use current timestamp.

erri120 (approved):

PR #23733, opened 2026-07-20
Added native crash handling

Aragas:
Tested, we'll be now getting reports like "PreviousSessionCrash: ACCESS_VIOLATION (0xc0000005) in electron.exe+0x398fe0". Keeping it minimal for now.
If there will be a need to have more info, will be added.
Should handle renderer/main crashes. 

* There are no pure JS minidump handlers, only with native code. Decided to write a mini parser, since we only check a few sections of the minidump. Controversial, can switch to a proper lib


src/main/src/minidump.ts
erri120, line 1: Electron doesn't provide a minidump reader package?

Aragas: There is one, but it requires a native dependency https://github.com/electron/node-minidump
My idea was that as long as our usage is minimal, the pure JS implementation _could_ be maintainable, but I'm not sure this would be the best strategy, so I would like to hear other opinions

erri120: Then it's fine. Other issue: have you tested this on Linux?

src/main/src/minidump.test.ts
erri120, line 1: Praser tests should parse a real minidump file.

src/main/src/minidump.ts
Aragas: Nope, but yea, that should be a good idea

Aragas: The minidump should have the same format. We'll need to add mapping for Linux specific codes. Are you able to provide a minidump for main and one for renderer crashes?

erri120: Don't have anything on hand, do you know of a reliable way to crash the processes?

Aragas: For renderer it's pretty simple - `DevTools console → process.crash()`
For main it might be a bit confusing - `pnpm -F @vortex/main run start:debug`, then open a chromium browser and type `chrome://inspect` and connect to `http://localhost:9229`
I usually use `Open dedicated DevTools for Node`. From there you should be able to call `process.crash()`. Just make sure that the JS Context is not `internal` at the top, otherwise it won't see `process`

erri120 (approved):

PR #23747, opened 2026-07-22
hash files on a main-process worker pool (LAZ-774)

IDCs:
Move MD5 file hashing off the renderer thread so finalizing a large download no longer blocks the UI.

Add a hash worker service (src/main/src/hash): a worker_thread pool that streams a file and returns its hex digest over a hash:compute IPC channel (window.api.hash.compute). It spawns lazily, runs pool-size jobs in parallel, isolates a crashed worker, and surfaces failures as VortexError. Route the download-finalize hash and checksum.fileMD5 / genMd5Hash through it; a Buffer already in renderer memory stays local.

Verified in a dev build: a 836 MB download finalizes off-thread in ~6.7s during peak collection installation which thrashes the disk (downloading/extracting/etc) with the UI fully interactable/responsive. Previously it would stutter/freeze + finalizing downloads would queue up to 40 minutes at times, affecting collection installation speed.

Also fixed bsdiff worker shutdown not being called on app quit.

closes LAZ-774

### DIFF
--- a/src/main/src/hash/compute.ts
+++ b/src/main/src/hash/compute.ts
@@ -1,0 +1,39 @@
+/**
+ * Pure file-hashing core, kept separate from worker.ts (a worker_thread entry
+ * point with parentPort side effects) so it can be unit-tested directly.
+ */
+
+import { createHash, getHashes } from "node:crypto";
+import * as fs from "node:fs";
+
+const supportedAlgorithms = new Set(getHashes());
+
+/** True if the algorithm name is one crypto.createHash accepts on this build. */
+export function isSupportedAlgorithm(algorithm: string): boolean {
+  return supportedAlgorithms.has(algorithm);
+}
+
+/**
+ * Stream a file and compute its hex digest. Rejects on an unsupported algorithm
+ * or any read error. numBytes is the number of bytes hashed (the file size).
+ */
+export function hashFileStream(
+  algorithm: string,
+  filePath: string,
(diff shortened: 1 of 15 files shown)


src/shared/src/types/ipc.ts
```diff
@@ -278,6 +278,12 @@ export interface InvokeChannels {
   "bsdiff:create": (oldPath: string, newPath: string, patchPath: string) => Promise<void>;
   "bsdiff:apply": (oldPath: string, patchPath: string, outputPath: string) => Promise<void>;
 
+  // file hashing, run on a main-process worker_thread (only the path crosses IPC)
+  "hash:compute": (
+    algorithm: string,
```
erri120: Let's do a proper closed set here.

src/main/src/hash/compute.ts
```diff
@@ -0,0 +1,39 @@
...
+    const hash = createHash(algorithm);
+    let numBytes = 0;
+    const stream = fs.createReadStream(filePath);
+    stream.on("data", (chunk: string | Buffer) => {
+      hash.update(chunk);
+      numBytes += chunk.length;
+    });
+    stream.on("end", () => resolve({ hash: hash.digest("hex"), numBytes }));
+    stream.on("error", reject);
+  });
```
erri120: Use Node.js pipeline feature.

erri120 (changes requested):

erri120 (approved):

PR #23748, opened 2026-07-22
Cleanup extension manager extension

erri120:
Cleanup and promisefies the entire extension in preparation for the backend deprecation.

Part of LAZ-690.


src/renderer/src/extensions/extension_manager/installExtension.ts
```diff
@@ -224,296 +199,217 @@ function isLocaleCode(input: string): boolean {
...
+  // the check in isLocaleCode is extremely unreliable because it will fall back to
+  // iso on everything. Was it always like that or was that changed in a recent
+  // node release?
+  const [language, country] = languageDirectory.name.split("-");
+  if (!languageExists(language) || (country !== undefined && !countryExists(country))) {
+    throw new DataInvalid("Directory isn't a language code");
+  }
+
+  const languageEntries = await readdir(
+    path.join(extPath, path.join(languageDirectory.parentPath, languageDirectory.name)),
```
IDCs: Double join here - `path.join(languageDirectory.parentPath, languageDirectory.name)` should be sufficient.

IDCs:
@erri120 - been testing this manually on my end and it seems to work just fine.
Tested:
- community game extension installation
- community game extension update
- tool extensions
- translations
- themes
- optional extensions
- installation through extensions browser.

What I would suggest though, is adding unit tests for the validation functions at the very least (would've caught that double path join for translations).

I'm assuming you were planning to add the unit-tests in a future PR?

IDCs (dismissed):
Would be good to see unit tests for this piece of work, but I'm assuming you're still refactoring and will add them in an upcoming PR so going to approve.

IDCs (approved):
nice

PR #23751, opened 2026-07-22
ensure we support linux for vortex-paths test-setup

IDCs:
I need to go through the collections tests and make sure to mock the paths locally so we can remove the vortex-paths test-setup stuff. But fixing this for linux in the file directly temporarily.

### DIFF
--- a/src/renderer/test-setup.ts
+++ b/src/renderer/test-setup.ts
@@ -1,5 +1,6 @@
 // NOTE(erri120): yes, the library is called "jest-dom" but it works for vitest as well with this import:
 // https://www.npmjs.com/package/@testing-library/jest-dom#with-vitest
+import os from "node:os";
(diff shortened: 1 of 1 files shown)


src/renderer/test-setup.ts
```diff
@@ -8,27 +10,28 @@ import { ApplicationData } from "./src/applicationData";
 
 // Every path keyed off a single absolute test root so getVortexPath-backed selectors
 // (downloadPathForGame, etc.) resolve real strings instead of throwing in tests.
-const TEST_ROOT = "C:\\vortex-test";
+const TEST_ROOT = process.platform === "win32" ? "C:\\vortex-test" : "/vortex-test";
```
erri120: Should be a temp directory.

erri120 (approved):

PR #23756, opened 2026-07-22
Bump node-api module to fix modsByUid not returning all results

Al12rs:


src/renderer/src/extensions/nexus_integration/eventHandlers.ts
```diff
@@ -942,16 +946,20 @@ export function onGetModRequirements(
...
-          validUids,
+        // Query must include modId to map results back to mods. modsByUid returns
+        // at most MODS_BY_UID_PAGE_SIZE mods per call, so split the UIDs into
+        // batches of that size, fetch them concurrently, and concatenate.
+        const modRequirements = getModRequirementsInfo();
+        const batches: string[][] = [];
+        for (let offset = 0; offset < validUids.length; offset += MODS_BY_UID_PAGE_SIZE) {
+          batches.push(validUids.slice(offset, offset + MODS_BY_UID_PAGE_SIZE));
+        }
+        const fetchedBatches = await Promise.all(
```
IDCs: Use `Promise.allSettled` here - otherwise one failed fetch dumps all other fetches.

IDCs (changes requested):

src/renderer/src/extensions/nexus_integration/eventHandlers.ts
IDCs: Rejected fetches should probably be logged here rather than propagated upstream.

Al12rs:
Will change this to use the node-nexus-api chunking instead once it's merged.

IDCs (approved):

PR #23759, opened 2026-07-23
Settings & Preferences: new Page + Tabs layout

insomnious:
Migrates the Settings and Preferences pages to the new `Page` + Tabs layout (header pictogram + title/subtitle), and moves the Interface and Vortex tab controls to design-system components. Download/Mods content and toggles are intentionally left for follow-ups.

electron_Nm7nzmUaws
electron_cMWH08NVds
electron_iy3OBReBr1
electron_zogjfHtVHM


src/renderer/src/extensions/settings_application/SettingsVortex.tsx
```diff
@@ -39,42 +42,48 @@ class SettingsVortex extends ComponentEx<IProps, IComponentState> {
 
     const restartNoti =
       multiUser === oldMultiUser ? null : (
-        <HelpBlock>
-          <Alert>
+        <div className="flex items-center gap-x-4 rounded-lg border border-info-weak bg-info-950 p-3">
+          <Typography appearance="strong" brand="neutral-translucent" className="grow">
```
r1chm8: I'd really like to get into the habit of not setting values on component when they are the default.

appearance defaults to "strong"
brand defaults to "neutral" (not used here but is in other places)

The reasoning being, if we ever want to do a sweeping change of the defaults it won't be consistent.

We should add a cluade rule to the frontend agent to omit props when they equal the default value.

r1chm8 (changes requested):

src/renderer/src/extensions/settings_application/SettingsVortex.tsx
insomnious: That's fair. Do you want to add that or shall I add that rule in this PR ?

r1chm8: If you don't mind doing it in this PR that would be great, you can get claude to update the agent file for you, and then remove any default props, thanks.

src/renderer/src/extensions/settings_application/SettingsVortex.tsx
```diff
@@ -39,42 +42,48 @@ class SettingsVortex extends ComponentEx<IProps, IComponentState> {
...
-            componentClass="select"
-            onChange={this.selectMode}
-            value={multiUser ? "on" : "off"}
-          >
-            <option value="on">{t("Shared")}</option>
-            <option value="off">{t("Per-User")}</option>
-          </FormControl>
-          {restartNoti}
+          <div className="flex flex-col items-start gap-y-2">
+            <Typography as="span" typographyType="body-md">
```
r1chm8: still have default props being set :sad-user: `typographyType="body-md"`

src/renderer/src/extensions/settings_interface/SettingsInterface.tsx
```diff
@@ -152,44 +152,48 @@ class SettingsInterfaceImpl extends ComponentEx<IProps, {}> {
...
-              }
-              return prev;
-            }, [])}
-          </FormControl>
-
-          <ControlLabel>
-            {t("When you select a language for the first time you may have to restart Vortex.")}
-          </ControlLabel>
+          <div className="flex flex-col items-start gap-y-2">
+            <Typography as="span" typographyType="body-md">
```
r1chm8: Here too

r1chm8 (approved):

IDCs (approved):

PR #23762, opened 2026-07-23
Fix stale Mod Requirements by switching to a session cache rather than persistent storage

Al12rs:
Closes LAZ-810

- Switch mod requirements to a session cache: moves mod-requirement data out of persisted Redux state (where it was never invalidated, causing stale data) into a session/memory cache.
- Store mod files data in the memory cache rather than Redux state, removing the associated session actions, reducers, and selectors. This also seems to have improved performance.
- Skip fetching unused data
- Reduce modsByUid query complexity by limiting the requirements page size in the GraphQL query.
- Removed now-dead health-check session plumbing (actions/reducers/selectors) and updated DetailView/types accordingly.
- Fix missing Mod Requirements fetches due to query cap by upgrading node-nexus-api https://github.com/Nexus-Mods/node-nexus-api/pull/41


Al12rs:
Closes LAZ-810

IDCs (approved):

PR #23768, opened 2026-07-23
fix: missing thumbnails for downloaded files by hydrating them

Al12rs:
The thumbnails could be missing for newly downloaded mods since the mod info would be populated, but not the thumbnails.
This makes sure the thumbnails data is hydrated when generating the health checks.

### DIFF
--- a/src/renderer/src/extensions/health_check/checks/fileRequirementsCheck.test.ts
+++ b/src/renderer/src/extensions/health_check/checks/fileRequirementsCheck.test.ts
@@ -80,7 +80,7 @@ function ref(fileUID: string, enabled = true, emitRequirements = true): IInstall
 }
 
 function downloadedRef(fileUID: string): IDownloadedFileRef {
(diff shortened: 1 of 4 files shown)


IDCs (changes requested):
So logic looks alright - what I would like to see in a bugfix PR are unit tests - `installedFiles` has no coverage currently.

At the very least you should be adding unit tests for `makeDownloadedFileHydrator` covering the fallback chains in `toDownloadedFile`; someone re-ordering a couple of fallbacks could potentially break this functionality.

* unknown fileUID/download removed from state
* contains_adult_content: false + details true - stays false (someone re-ordering a couple of fallbacks or swapping ?? for || would silently break this)
* etc

IDCs:
Basically for bugfixes the minimum level of tests should always be covering regression.

IDCs (dismissed):

IDCs (approved):

PR #23774, opened 2026-07-24
Feature - Add filter to Version dropdown for mods with multiple versions

ChemGuy1611:
Add filter to Version dropdown to only show mods with multiple versions in Vortex.

## Summary

Adds a new preset option, "Multiple versions installed", to the Version column
filter on the mods page. When selected, the table shows only mods that have
more than one version installed at the same time (i.e. multiple entries
sharing the same Nexus mod ID, such as when a mod is installed as separate
variants instead of being upgraded in place).

## Changes

`src/renderer/src/extensions/mod_management/util/VersionFilter.tsx`

1. Added a new entry to `PRESET_OPTIONS`:
   ```ts
   { value: "multi-version", label: "Multiple versions installed" }
   ```

2. Extended `VersionFilter.matches()` to accept the `state` parameter (already
   supported by the `ITableFilter` interface but previously unused here), and
   added a `multi-version` branch: it reads the row's `attributes.modId`,
   looks up all mods for the active game in `state.persistent.mods`, and
   counts how many share that `modId`. If more than one match, the row passes
   the filter.

### DIFF
--- a/src/renderer/src/extensions/mod_management/util/VersionFilter.test.ts
+++ b/src/renderer/src/extensions/mod_management/util/VersionFilter.test.ts
@@ -1,0 +1,322 @@
+import { describe, expect, it } from "vitest";
+
+import { makeApiHarness, makeMod, makeProfile } from "../../../test-utils/builders";
+import type { IApiHarness } from "../../../test-utils/harnessTypes";
+import type { IMod, IModAttributes } from "../types/IMod";
+import VersionFilter from "./VersionFilter";
+
+const GAME_ID = "skyrimse";
+const OTHER_GAME_ID = "fallout4";
+const PROFILE_ID = "profile-skyrimse";
+const OTHER_PROFILE_ID = "profile-fallout4";
+
+/**
+ * A harness seeded with the mods each game has installed, active on GAME_ID.
+ * VersionFilter resolves the game through the active profile, so `activeProfileId` is
+ * what decides which mod list the counts come from.
+ */
+function seed(mods: Record<string, Record<string, IMod>>): IApiHarness {
+  const harness = makeApiHarness({
+    profiles: { [PROFILE_ID]: makeProfile({ id: PROFILE_ID, gameId: GAME_ID }) },
+    mods,
+  });
+  harness.setState((draft) => {
+    draft.settings.profiles.activeProfileId = PROFILE_ID;
+  });
+  return harness;
+}
+
+/**
+ * An installed mod that reports the given Nexus mod id. `modId` is typed loosely
+ * because ICommonModAttributes declares it a number while the store also holds the
+ * string form (from meta lookups) and the nullish forms these tests cover.
+ */
+function installed(id: string, modId: unknown): IMod {
+  return makeMod({ id, state: "installed", attributes: { modId } as IModAttributes });
+}
+
+function byId(...mods: IMod[]): Record<string, IMod> {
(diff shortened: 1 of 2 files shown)


IDCs (dismissed):
Hey @ChemGuy1611 - is this your first Vortex PR? congrats!

Couple of issues with it:
1. Slow with lots of mods. This runs once per table row, and each run loops over every mod again - with 2,000 mods that's 4 million checks, repeated every refresh. Count how many mods share each modId once, cache it, and look it up per row. activeGameId(state) doesn't need to be per-row either.
2. We're moving away from the store helpers i.e `getSafe` - use optional chaining instead: `state.persistent.mods?.[gameId] ?? {}`
3. Uninstalled mods are not filtered out - accept only mods that have the "installed" state.
4. I'm afraid this is a band-aid fix, useful - but the actual problem here is the fact that the mods weren't grouped together correctly.

ChemGuy1611:
@IDCs - Updated per your feedback. Also removed a few other instances of getSafe that were already in the file. 

For context, the goal of this is to make it easier to find which mods have had multiple versions installed for the purpose of removing old versions and archives. Requested on the forum: https://forums.nexusmods.com/topic/13542191-would-be-cool-if-the-versions-dropdown-had-the-ability-to-select-all-mod-files-with-multiple-versions-for-easy-cleanup-of-older-versions/?do=getNewComment

I also fixed so that previously installed archives are counted, which is the intended behavior. Built the app and tested, seems to be working correctly.

electron_5kppRrTcZL

electron_32626dOmTs

IDCs:
Hey @ChemGuy1611

The only remaining issue is the downloads scan. `installed` is set on install and never cleared on uninstall ([IDownload.ts:160](https://github.com/Nexus-Mods/Vortex/blob/master/src/renderer/src/extensions/download_management/types/IDownload.ts#L160)), so any mod updated in place counts 2 - the current mod plus the leftover archive it replaced. That's the normal lifecycle, so "Multiple versions installed" would match most of the list instead of side-by-side variants.

Counting only installed mods matches your summary and drops ~35 lines. The mods identity check alone keeps the memoisation, so `mCachedState` and `mCachedDownloads` both go away.

Four cache fields become two:

```
  // installed versions per Nexus mod id, cached so we don't rescan the mod
  // list for every row we're asked to match
  private mCachedMods: Record<string, IMod> | undefined;
  private mVersionCounts: Record<string, number> = {};

  public matches(filter: any, value: any, state: IState): boolean {
```
And versionCounts becomes:

```
  /**
   * how many versions of each mod are installed, keyed by Nexus mod id. Rebuilt only
   * when the mod list it was derived from actually changed, so a single filter pass
   * walks the list once rather than once per row.
   */
  private versionCounts(state: IState): Record<string, number> {
    const gameId = activeGameId(state);
    const mods = state.persistent.mods?.[gameId] ?? {};

    if (mods !== this.mCachedMods) {
      this.mCachedMods = mods;

      const counts: Record<string, number> = {};
      for (const mod of Object.values<IMod>(mods)) {
        const modId = mod.attributes?.modId;
        if (mod.state === "installed" && modId !== undefined) {
          counts[modId] = (counts[modId] ?? 0) + 1;
        }
      }
      this.mVersionCounts = counts;
    }

    return this.mVersionCounts;
  }
```

ChemGuy1611:
@IDCs - Focused to only installed mods, per your change request.

IDCs (approved):

PR #23040, opened 2026-05-06
Upgrade Downloads Graph Dependencies

Sewer56:
fixes LAZ-199

## Summary

`DownloadGraph` blocks the React 17 migration because it depends on 
`react-resize-detector@4.2.3` and `recharts@1.8.6`, both of which only support
React 16 as a peer dependency.

This PR updates/replaces those dependencies without changing Downloads page
behaviour (expand/collapse, resize handling, recent speed history, bandwidth-limit line).

See LAZ-199 for details.

### Dependency updates
- **Remove** `react-resize-detector` from catalog, `src/main/package.json`, and `src/renderer/package.json`
- **Upgrade** `recharts` from `^1.8.5` → `^2.15.4` (React 17 support since 2.0.3)

### `DownloadGraph.tsx`
- Replace `react-resize-detector` with native `ResizeObserver`
- Update `setRef` to store container ref and observe resize via `ResizeObserver`
- Remove `truthy` import (no longer needed)

### `charts.scss`
- Update `.recharts-area` selectors: use `.recharts-area-area` / `.recharts-area-curve` class names instead of `:first-child` / `:last-child` pseudo-selectors (recharts 2.x changed markup structure)

### Test coverage
- Add `DownloadGraph.test.tsx` covering:
  - `ResizeObserver` triggers chart width update
  - `disconnect()` called on unmount
  - Bandwidth limit label rendered when `maxBandwidth` constrains the graph

### Config
- Add `vitest.config.mts` glob to `vitest.config.ts` projects

## Acceptance criteria

- [x] No active dependency path for the Downloads graph uses `react-resize-detector@4.2.3` or `recharts@1.8.6`
- [x] Graph still renders above the downloads table, expands/collapses, resizes with container, updates from `speedHistory`, shows bandwidth-limit line
- [x] Current no-animation behaviour preserved
- [x] Build, test, lint, format pass
- [x] Manual smoke-test of Downloads page graph

PR #23775, opened 2026-07-24
Upgrade React to 17.0.2

insomnious:
fixes LAZ-198, LAZ-199, LAZ-200, LAZ-201
Supersedes #23040 (its Downloads-graph changes are included here, reconciled with current master).

## Summary

Upgrades the workspace and app from `react`/`react-dom` 16.14.0 to **17.0.2**. React 17 is the deliberate bridge release (no 18/19 changes here): it preserves the current app model while unblocking later upgrades and modern library adoption.

Because extensions don't bundle React — the bundler externalizes host deps and `extensionRequire` resolves them from the host at runtime — this is a single atomic switch for the host and every extension. All extension-facing surfaces are preserved: the react-select v1 API (via `ReactSelectWrap` remap), react-dnd 14, react-bootstrap 0.33.

## Changes

1. **Downloads graph (LAZ-199):** `react-resize-detector` removed in favour of native `ResizeObserver`; `recharts` 1.8.6 → ^2.15.4; `charts.scss` selectors updated for recharts-2 markup; adds `DownloadGraph.test.tsx`.
2. **react-select kept at 1.3.0 (LAZ-200):** peer range capped at ^16, handled via `peerDependencyRules.allowedVersions` (incl. its `react-input-autosize` dependency). API-frozen and remapped for extensions; real replacement remains LAZ-248 (Headless UI).
3. **react-sortable-tree → @nosferatu500 fork (LAZ-201):** `@nosferatu500/react-sortable-tree@4.4.0` + `@nosferatu500/theme-file-explorer@3.0.21` (last versions compatible with react-dnd 14). The fork pins `react-dnd@14.0.4` exact, so an override forces the host's 14.0.5 — a single react-dnd instance is load-bearing (LAZ-731) and verified in the lockfile. Only consumer (`OverrideEditor.tsx`) migrated; vendored tree CSS refreshed from the fork.
4. **Core bump (LAZ-198):** catalog `react`/`react-dom` 17.0.2, `@types/react` ^17.0.93 / `@types/react-dom` ~17.0.26 (overrides 16 → 17), `why-did-you-render` 10.0.1 → 7.0.1 (10.x requires React 19 and was already peer-broken; 7.0.1 is the newest with React 17 support). **Zero source changes needed.** Also adds a `bbcode-to-react` peer allowance (git fork, peers cap at ^16, uses only createElement-era APIs).
5. **ContextMenu fix:** React 17 delegates events at the react root instead of `document`, so `RootCloseWrapper`'s document-level close listeners now mounted while the opening right-click was still bubbling — the menu closed itself instantly. Listener registration is deferred one tick, restoring React 16 behaviour; regression test added.

## Verification

- Build (230 nx tasks), 1558 renderer unit tests, lint, format all green; `pnpm install` emits zero react peer warnings.
- Lockfile carries zero React-16 residue; single `react`/`react-dnd` instance verified down to the symlink graph.
- Playwright E2E: 25/25 credential-free tests pass on React 17.
- Full manual smoke pass: Downloads graph, OverrideEditor (Manage File Conflicts) on the fork, all react-select surfaces, the React-17 event-delegation risk set (context menus, portal dropdowns, tooltips, webview, overlay portal), bundled extensions (gamebryo-plugin-management, documentation, issue-tracker), soak (install → deploy → purge), and third-party extensions Bannerlord (BUTR) + Support Stats. `DEBUG_REACT_RENDERS` (wdyr 7.0.1) verified live.
- Two days of log monitoring during testing: no new React warnings (the only React warning observed, `validateDOMNesting`, predates the upgrade).

## Known items / follow-ups

- `flatpak/generated-sources.json` is stale and currently **cannot** be regenerated: the tooling requires `yarn.lock` (see warning in `docs/flatpak/maintenance.md`). Needs a pnpm-aware generation ticket before a release ships from this.
- ~~`recharts` is `^2.15.4` while the rest of the react-adjacent catalog is exact-pinned — can tighten to `2.15.4` if preferred.~~ ✅ Resolved: pinned to exact `2.15.4` (`react`/`react-dom` were already exact-pinned at `17.0.2`).
- ~~Third-party impact expectation: minimal. The event-delegation change is the main thing for authors doing manual `document`-level listeners; this branch also removes `react-resize-detector` and unscoped `react-sortable-tree`(+theme) from the extension-resolvable module surface and bumps recharts across a major — those break at require-time for any extension using them directly.~~ ✅ Verified: extension analysis confirms no extension imports `react-resize-detector` or `react-sortable-tree` — the removals are safe. Residual risks unchanged: the recharts 1→2 major bump and manual `document`-level listeners (not covered by the analysis).
- `renderer.tsx`'s react-sortable-tree error-swallow kept (comment updated); candidate for deletion since the only consumer sets `canDrag={false}`.


pnpm-workspace.yaml
```diff
@@ -203,23 +204,20 @@ catalog:
...
   react-i18next: 11.18.6
   react-markdown: 6.0.3
   react-overlays: 1.2.0
   react-redux: 7.2.9
-  react-resize-detector: 4.2.3
   react-select: 1.3.0
-  react-sortable-tree: 2.8.0
-  react-sortable-tree-theme-file-explorer: 2.0.0
-  recharts: 1.8.6
+  recharts: ^2.15.4
```
IDCs: Lets pin this to an exact version; should probably do the same for react and react-dom

IDCs (changes requested):

pnpm-workspace.yaml
IDCs: So based on the lock file:
react + react-dom -> 17.0.2
recharts -> 2.15.4

src/renderer/src/controls/ContextMenu.tsx
```diff
@@ -155,12 +155,22 @@ function MenuAction(props: IMenuActionProps) {
 }
 
 class RootCloseWrapper extends React.Component<{ onClose: () => void }, {}> {
+  // React 17 delegates events at the react root instead of document, so this
+  // component mounts while the click/contextmenu that opened the menu is
+  // still bubbling toward document; registering the close listeners
+  // synchronously would make that same event close the menu immediately.
+  // Defer registration until the current event has finished dispatching.
+  private mAttachTimer: NodeJS.Timeout;
```
erri120: `setTimeout` is provided by both Node.js via `nodeIntegration` and via global `window.setTimeout`. The DOM method returns `number` while the Node.js version returns `NodeJS.Timeout`. Even though the IDE will suggest the `setTimeout` method, if you run `typeof setTimeout(() => { })` you'll get `number`.

To prevent weirdness, infer the return type:

```typescript
  private mAttachTimer: ReturnType<typeof setTimeout>;
```

Or explicitly use the DOM method:

```typescript
window.setTimeout(() = { });
```

erri120:
Verified with latest extension analysis that no extension is importing `react-resize-detector` or `react-sortable-tree`, safe to delete the dependencies.

erri120 (dismissed):

IDCs (dismissed):

IDCs (approved):

erri120 (approved):

PR #23781, opened 2026-07-27
fix renderer memory leak crashing installs on large mod libraries (LAZ-834)

### LINKED ISSUE #23776: vortex crashes regularly, my assumption due to having too many mod files
**Your Nexus Mods username**
fractalbase

**Describe the bug**
vortex regularly crashes now, when I attempt to install mods, or attempting to remove old versions of mods.

**To Reproduce**
just attempt to use vortex with the amount of mod files I have (maybe about 6000?)

**Expected behavior**
vortex shouldn't crash.

**Logs**

**Screenshots**
N/A

Windows 11 Pro
Latest Vortex version
Skyrim Special Edition

IDCs:
Fixes a renderer memory leak that crashes Vortex during installs on large mod libraries (#23776: 39 OOM kills in ~24h on a 4852-mod Skyrim setup, ~3.3 GB against a 4 GB ceiling).

Three faults compound:

- **Per-mod checks are not scoped to a game.** The runner enumerates mods for the ACTIVE game, so the `xrebirth-*` checks walk whatever game is managed. `IHealthCheck` gains `gameId`: the registry runs such a check only while that game is active, and discards a result it holds for another.
- **The walk is unbounded.** `runPerModCheck` walks in waves of `MOD_WALK_CONCURRENCY` (16), holding one wave's file lists rather than the whole library's. A mod whose staging folder cannot be read fails on its own.
- **A timed-out run is not cancelled.** The registry aborts an `AbortSignal` on timeout and holds the check's concurrency slot until the body settles, so a re-trigger waits rather than stacking another pass.

16 is measured: the walk is filesystem-bound and flat from 8 to 128 concurrency (~1-2s for 4900 mods / 280K files either way), so the cap costs nothing.

Tests cover each fault and use no fixed sleeps — the harness exposes `hasSettled()` for `vi.waitFor`, and the fan-out probe runs on microtasks rather than timers.

### Public API

Additive and optional; existing extensions compile unchanged (`etc/vortex.api.md` regenerated).

```ts
type HealthCheckFunction = (api, signal?: AbortSignal) => Promise<IHealthCheckResult>
type PerModCheckFunction = (api, mod, signal?: AbortSignal) => Promise<IHealthCheckResult>
interface IHealthCheck { gameId?: string }
```

### Known limitations

- A timed-out check's partial result is discarded in favour of `Health check timed out`; the runner labels it `(stopped after N of M mods)` but nothing consumes it.
- The 30s budget does not scale with library size.

closes #23776
closes LAZ-834

### DIFF
--- a/etc/vortex.api.md
+++ b/etc/vortex.api.md
@@ -1602,8 +1602,8 @@ enum HealthCheckCategory {
 // @public (undocumented)
 type HealthCheckFixFunction = (api: IExtensionApi) => Promise<void>;
 
-// @public (undocumented)
-type HealthCheckFunction = (api: IExtensionApi) => Promise<IHealthCheckResult>;
+// @public
+type HealthCheckFunction = (api: IExtensionApi, signal?: AbortSignal) => Promise<IHealthCheckResult>;
 
 // @public (undocumented)
 enum HealthCheckSeverity {
@@ -2698,6 +2698,7 @@ interface IHealthCheck {
     extensionName?: string;
     // (undocumented)
     fix?: HealthCheckFixFunction;
+    gameId?: string;
     // (undocumented)
     id: string;
     // (undocumented)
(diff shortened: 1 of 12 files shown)


erri120:
Might want to check out `p-queue` which I added for the downloader.

erri120 (approved):

PR #23790, opened 2026-07-27
Add Health Check analytics tracking (LAZ-551, Phase 1)

r1chm8:
Instruments the Health Check feature with Mixpanel analytics per LAZ-551. This is Phase 1 — all UI-driven events (page, mod-requirement path, file-requirement path, premium upsell). Phase 2 (scan lifecycle + install started/completed/failed) will follow in a separate PR.

### What's included

Global context
- Registers the active game (game_id, numeric Nexus id) and profile_id as Mixpanel super properties, cleared when no game is active — so every event carries game/profile scope without per-event plumbing. (game_name intentionally excluded per the data team.)

Tracking hook                                                                                                 
- New useHealthCheckTracking(api) — one typed callback per event, built on a pure createHealthCheckTracker factory (unit-testable without React, and reusable outside components in Phase 2).
- Pure mappers: issueTypeForCheck (file → warning, mod → suggestion) and resolutionTypeForCategory.

Also                                                                                                          
- Fixed the file-requirement 1-click install buttons to use the monitor-download icon (were using the tray icon), matching the mod path.

### DIFF
--- a/src/renderer/src/extensions/analytics/index.ts
+++ b/src/renderer/src/extensions/analytics/index.ts
@@ -2,13 +2,16 @@ import * as os from "os";
 
 import { getErrorMessageOrDefault } from "@vortex/shared";
 
-import type { IExtensionContext } from "../../types/IExtensionContext";
-import { getCPUArch } from "../../util/nativeArch";
+import type { IExtensionContext } from "@/types/IExtensionContext";
+import { getCPUArch } from "@/util/nativeArch";
+
+import { activeGameId, activeProfileId } from "../../util/selectors";
 import { setAnalytics } from "./actions/analytics.action";
 import { HELP_ARTICLE, PRIVACY_POLICY } from "./constants";
 import AnalyticsMixpanel from "./mixpanel/MixpanelAnalytics";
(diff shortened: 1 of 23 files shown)


Al12rs:
I think there are a couple of events that are dangling:
`trackIssueExpanded` and `trackEnableAllClicked` are never called, so
`health_check_issue_expanded` and `health_check_enable_all_clicked` never fire. Both
`ListingRow`s call `onOpen()` untracked.

r1chm8:
> I think there are a couple of events that are dangling: `trackIssueExpanded` and `trackEnableAllClicked` are never called, so `health_check_issue_expanded` and `health_check_enable_all_clicked` never fire. Both `ListingRow`s call `onOpen()` untracked.

We don't actually have UI to expand an issue, DetailView fires `health_check_detail_viewed` on mount. So the navigation is captured — just under a different event name.  I will delete the `trackIssueExpanded` event.

Also there's no "Enable all (N)" button, and enabling is per-file via enable_this_version_clicked, which does fire. I will delete this event too.

Al12rs (approved):

PR #23795, opened 2026-07-28
Make the Health Check issue identity ambient (LAZ-551, cleanup)

r1chm8:
## Summary

Cleanup on the LAZ-551 Health Check analytics: the issue an event belongs to is now ambient rather than threaded through every layer by hand. Includes one bug fix and one new event found on the way.

**Stacked on #23790 and the Phase 2 PR** — base is `laz-551-p2`, so this diff is the cleanup only. Merge those first.

### What changes on the wire

Mostly nothing — the refactor is payload-identical, pinned by tests. Two deliberate exceptions:

| Change | Why |
|---|---|
| `issue_id` value corrected for hidden file requirements | It was wrong; see commit 4 |
| New `health_check_unhide_all_clicked` | The spec had no counterpart to `hide_all_clicked`; see commit 3 |

Both are raised on LAZ-551 for the data team.

## Why

Both checks emit the same `health_check_*` event names, so every issue-scoped event has to carry `issue_id` and `check_id` to stay separable. After the review feedback that added `check_id`, the same fact was restated at ~27 call sites:

| Duplication | Count |
|---|---|
| `useHealthCheckTracking(api)` | 8 components |
| `checkNameForCheck(entry.checkId)` locals | 4 components + 2 inline |
| `issue_id: entry.id, check_id: checkName` in event calls | ~27 |
| `tracking={{ api, issueId, checkId, ... }}` prop objects | 6 render sites |
| Interfaces existing only to carry the above | 2 |

A miss is invisible at compile time on the optional-scope events, and an analytics dimension that was never sent cannot be backfilled — so this is the fact most worth making impossible to get wrong. Commit 4 is exactly that failure, found by review while this was in flight.

## Commit 1 — ambient issue identity

A tracking context supplies the identity:

- `HealthCheckTrackingProvider` — the tracker, wrapping the page.
- `IssueProvider` — one listing entry, wrapping each row and the detail page.
- `useIssueTracking()` — the tracker with `issue_id` / `check_id` already applied. **Throws outside an `IssueProvider`** rather than emitting an event missing its dimensions.
- `useIssue()` / `useOptionalIssue()` — the identity itself, for the events that also carry `issue_type` and for handing an install action its issue.

Call sites went from

```ts
const checkName = checkNameForCheck(entry.checkId);
const { trackOneClickInstallClicked } = useHealthCheckTracking(api);
trackOneClickInstallClicked({ issue_id: entry.id, check_id: checkName, mod_id, ... });
```

to `trackOneClickInstallClicked({ mod_id, ... })`.

Two details worth a reviewer's eye:

- **`IssueTracker` drops the cross-check and optional-identity methods** via a mapped type. Without that, calling `trackPremiumModalShown` through `useIssueTracking()` would compile and silently lose the identity — the exact failure this change exists to prevent.
- **The page-level premium surfaces render outside any `IssueProvider`**, so they read no identity and emit without `issue_id` / `check_id` — exactly the cross-check contract agreed with the data team. The component tree now enforces what was a manual convention.

Also here: `PremiumModal` / `PremiumBanner` lose their `tracking={{...}}` prop objects and both tracking interfaces; the install actions take one snake_case `IssueIdentity` instead of a camelCase twin; `issueTypeForCheck` gets the same exhaustive `Record<HealthCheckId, _>` as `CHECK_NAMES`, so a third check is a build error rather than a silent collision.

`IssueIdentity` follows the existing `ModAnalyticsIdentity` naming, which leaves "scope" meaning only the React region an `IssueProvider` covers.

## Commit 2 — fold single-consumer helpers

Several helpers were exported from shared modules but only ever used by one file:

- `issueFor` and the tracker-binding logic → into the tracking context, their only caller.
- `useHealthCheckTracking` → inlined into `HealthCheckPage`; it was a one-line `useMemo` with one call site. With the hook gone the module holds only the factory, so `useHealthCheckTracking.ts` is renamed `healthCheckTracker.ts`.
- `countActiveIssues` → into the api layer that runs the scan.
- `modEntryId` / `isModHidden` → into `ModRequirementsContent`, deleting `modRequirementEntries.ts`.

## Commit 3 — track bulk unhide (new event)

`hide_all_clicked` had no counterpart, so restoring hidden issues left no trace — not even per-issue `issue_unhidden` events, since bulk hide and unhide both bypass the row handlers.

That biases a KPI listed on the ticket: **hide rate — % of issues hidden (indicates low-value or noisy issues)**. Ten issues hidden and then restored read identically to ten left hidden.

`health_check_unhide_all_clicked` carries `issue_count_unhidden`, mirroring `hide_all_clicked`, and is a cross-check aggregate so it carries no issue identity.

## Commit 4 — `issue_id` stability fix 🐞

**Found in review — the most serious issue here.**

A file requirement's listing id gains a `::hidden` suffix once dismissed, and Phase 1 reported that id as `issue_id`. So hiding an issue emitted `uid:download-replace` and unhiding it emitted `uid:download-replace::hidden`. Since `issue_unhidden` can only ever fire from a dismissed entry, **every file-requirement unhide event carried a key that joined to nothing**. The same applied to `detail_viewed` and any resolution action taken from the Hidden tab.

Mod requirements were unaffected (`modEntryId` has no suffix), so the two checks also disagreed about whether `issue_id` was stable — and `collectInstallAll` already used the unsuffixed form, so install events on a partially dismissed file already disagreed with that file's own hide/unhide events.

**The suffix is not removed.** It is correct as a row key: a partially dismissed file contributes both a showing and a dismissed entry for the same category, and both `HealthCheckPage`'s React key and `HealthCheckDetailPage`'s `find(c => c.id === entry.id)` need to tell them apart. Removing it would trade a data bug for a UI bug.

Instead the two concerns are separated:

- `IHealthCheckEntry.id` — row key, documented as possibly encoding mutable UI state.
- `IHealthCheckEntry.issueId?` — stable analytics identity, defaulting to `id`.
- `issueFor` prefers `issueId`; the helper splits into `fileEntryId(uid, category)` (stable) and `fileRowKey(uid, category, hidden)` (row key).

Only file content sets `issueId`. Because of commit 1 there was a single derivation point to change rather than 27 call sites.


Al12rs:
`IssueIdentity` tripped me up on first read: the name suggests a unique id for the report,
something you could compare to detect the report changed. But `issue_id` looks like it's
meant to do the opposite, it leaves report contents out (`${sourceFileUID}:${category}`) so
the same file and category keep one value as the report changes underneath, which is what
lets the funnel join and what the `::hidden` commit was protecting. And `check_id` is more
of a facet than an identity.

So maybe `IssueScope`? It'd match the vocabulary already around it (`IssueProvider`,
`useIssue`, `IssueTracker`). Type-only rename, plus the `identity` field in `IIssueValue` and
the `identity?` params on the install actions.

Might be worth a doc line as well, since a rename on its own probably won't stop the same
misreading: that it's a join key, insensitive to report contents by design, with
`IHealthCheckEntry.id` being the one that does track report state.

src/renderer/src/extensions/health_check/views/content/fileRequirementEntries.ts
```diff
@@ -26,15 +26,23 @@ export const isFileEntryHidden = (
...
 /**
- * Listing-entry id for one source file's requirements of a given category — also the
- * issue_id the analytics events report, so the bulk-install path can attribute an install
- * to the same entry the listing shows.
+ * The issue one source file's requirements of a given category represent — the same
+ * whether they are showing or dismissed, so it is what the analytics events report as
+ * `issue_id` and what the bulk-install path attributes an install to.
  */
-export const fileEntryId = (
+export const fileEntryId = (sourceFileUID: string, category: FileRequirementCategory): string =>
```
Al12rs: This might be more accurate as `fileIssueId` . It produces entry.issueId, while fileRowKey produces entry.id, so the current name points at the wrong one of the two. Two call sites plus the doc reference.

r1chm8:
> `IssueIdentity` tripped me up on first read: the name suggests a unique id for the report, something you could compare to detect the report changed. But `issue_id` looks like it's meant to do the opposite, it leaves report contents out (`${sourceFileUID}:${category}`) so the same file and category keep one value as the report changes underneath, which is what lets the funnel join and what the `::hidden` commit was protecting. And `check_id` is more of a facet than an identity.
> 
> So maybe `IssueScope`? It'd match the vocabulary already around it (`IssueProvider`, `useIssue`, `IssueTracker`). Type-only rename, plus the `identity` field in `IIssueValue` and the `identity?` params on the install actions.
> 
> Might be worth a doc line as well, since a rename on its own probably won't stop the same misreading: that it's a join key, insensitive to report contents by design, with `IHealthCheckEntry.id` being the one that does track report state.

Agreed on the doc, that's the part that actually prevents the misread, so I'll add it either way:                                                                                                   

```ts                                                                                                         
/**                                                                                                           
 * Which issue an event belongs to, and which check surfaced it, the join key that                           
 * ties a user's events together across the funnel.                                                           
 *                                                                                                            
 * Deliberately insensitive to report contents: it is `${sourceFileUID}:${category}`,                         
 * so it stays put as requirements come and go underneath, and as the issue is hidden                         
 * and restored. `IHealthCheckEntry.id` is the row key that tracks report state.                              
 */
```

On the name I'd push back. It was IssueScope and was renamed this morning for the mirror-image reason: "scope" meant both this and the region an IssueProvider covers ("Scopes everything below it to one listing entry"). Going back re-creates that. And "identity" already means stable-under-change, what you hit sounds more like a content fingerprint, which would be a revision or digest.

You're right that check_id is a facet, not part of the id, the two checks' ids don't collide, so it isn't load-bearing. That weakens "identity", but weakens "scope" equally.

I will add the doc line now, name unchanged. If it still reads wrong afterwards I'd go to IssueKey rather than back to IssueScope.

src/renderer/src/extensions/health_check/views/content/fileRequirementEntries.ts
r1chm8: I agree, nice spot

Al12rs (approved):

PR #23797, opened 2026-07-28
GitHub Actions Packaging and Release Overhaul

Aragas:
Reworked the Packaging flow
1. The package action now only builds Vortex.
It was previously also creating the GitHub Release and publishing the artifacts
2. package.yml workflow uploads the artifacts, creates the Releases and uploads to R2
3. package.yml workflow does not have nightly anymore, nightly creates the artifacts now
4. nightly.yml workflow builds and uploads the artifacts separately
5. upload-release-to-r2.yml workflow is now a manual fallback


erri120 (changes requested):
Please re-use what we already have:
- no separate pnpm workspace
- rolldown for building
- oxlint with shared config for linting
- tsc with shared strict config for typechecking
- vitest with shared config for testing

.github/actions/prepare-nexusmods-release/dist/index.js
erri120, line 1: You can use `.gitattributes` to mark the file as generated/blob so GitHub and git don't try to diff anymore.

Aragas:
> Please re-use what we already have:
>     * no separate pnpm workspace
>     * rolldown for building
>     * oxlint with shared config for linting
>     * tsc with shared strict config for typechecking
>     * vitest with shared config for testing

My assumption was that we don't want to mix infrastructure code with the application code? We'll be introducing new dependencies to the app pnpm workspace this way. It's a better practice than to split?

erri120 (approved):

PR #23810, opened 2026-07-29
Increase feature flag intervals

erri120:
- GET request for feature flags: 60 minutes instead of 60 seconds
- POST request for metrics: 15 minutes instead of 60 seconds

I put the metrics on 15 minutes because they are lazy already, meaning we don't do a POST request if we have nothing to send.

Closes LAZ-792.


IDCs (dismissed):

erri120:
Need to update the tests, they use hardcoded values.

IDCs (approved):

PR #23821, opened 2026-07-29
validate analytics event properties before sending

IDCs:
MixpanelEvent.properties is Record<string, any> and most event classes forward caller input verbatim, so nothing stopped a malformed payload reaching Mixpanel. A collections_installation_failed row arrived in 2.4.0 carrying only {"0":"4",...} — a bare id string spread into an object — with none of its real properties.

Add zod contracts for the collection-install and mod-install families, validated at the single send boundary in trackEvent. Unknown keys are stripped rather than rejected; nothing read from download metadata is required, since those ids come through optional chaining and parseInt and are legitimately absent for bundled and direct downloads. Garbage is caught by requiring at least one recognized property to survive. Events without a contract are still sent unvalidated.

part of LAZ-862

### DIFF
--- a/src/renderer/src/extensions/analytics/mixpanel/MixpanelAnalytics.ts
+++ b/src/renderer/src/extensions/analytics/mixpanel/MixpanelAnalytics.ts
@@ -8,6 +8,7 @@ import type {
 } from "../../nexus_integration/types/IValidateKeyData";
 import { MIXPANEL_PROD_TOKEN, MIXPANEL_DEV_TOKEN } from "../constants";
 import { analyticsServiceLog } from "../utils/analyticsLog";
+import { validateEventProperties } from "./eventSchemas";
 import type { MixpanelEvent } from "./MixpanelEvents";
 
 class MixpanelAnalytics {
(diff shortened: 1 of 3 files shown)


erri120 (changes requested):
With this change we now have three different places where the define a event contract:

1) Mixpanel Lexicon
2) This new runtime schema
3) The existing `MixpanelEvent` type and classes implementing it for typechecking

The issue is that these are not linked in any way. Making a change to the Mixpanel Lexicon doesn't impact our event classes, changing the runtime schema doesn't give us typechecking issues during development and changing the event classes won't pass validation without changing the new schema as well.

Since the linked ticket LAZ-862 is about malformed data reaching Mixpanel, I suggest checking if Mixpanel has a "strict" mode that rejects data not fitting the Lexicon. While we can change things in Vortex, users would actually have to run the latest version to send "valid" data, old versions of Vortex would continue sending malformed data.

As for what we can do inside Vortex, if we want to make sure we're only sending valid data as part of a schema, both the runtime and typechecking schemas should be the same, ie using `zod` and exporting both the Zod schema as a JavaScript object and a TypeScript type so that we get validation on all stages.

IDCs:
Nope - governance/workarounds for this will have to happen on the mixpanel/server side.

PR #23826, opened 2026-07-29
Anchor Settings pickers left so their menus stay inside the pane

insomnious:
The Picker default `placement="right"` opens its menu leftward, which the Settings pane's overflow clipping cuts off — the language, multi-user, and update-channel pickers appeared hidden behind the sidebar. Anchor them `left` (same idiom as the Design System demos); Headless UI v2's auto-positioning will retire the prop later.

Adds an e2e regression test that opens the language picker and hit-tests the menu's corners.

Screenshot 2026-07-29 212730

Screenshot 2026-07-29 212832

### DIFF
--- a/packages/e2e/src/tests/settings.spec.ts
+++ b/packages/e2e/src/tests/settings.spec.ts
@@ -26,6 +26,42 @@ test.describe("Settings - Interface Tab", () => {
     });
   });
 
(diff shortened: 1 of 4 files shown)


r1chm8 (approved):

r1chm8:
When we update headless these elements will be content aware and we won't need to set placement.

PR #23833, opened 2026-07-30
gamebryo-plugin-management: Fix plugin IDs for LOOT on case-sensitive filesystems

caseif:
This tweaks the logic for deriving plugin names to pass to LOOT to preserve case so that it's able to actually load the plugin files on case-sensitive filesystems. I haven't tested this on Windows as I don't have an install readily available, so it would be very much appreciated if anyone is able to make sure this doesn't blow anything up there.

### DIFF
--- a/extensions/gamebryo-plugin-management/src/autosort.ts
+++ b/extensions/gamebryo-plugin-management/src/autosort.ts
@@ -18,6 +18,7 @@ import { gameDataPath, gameSupported, nativePlugins, pluginPath } from "./util/g
 import { missingGroupFixes } from "./util/groups";
 import { invalidPluginsFromError } from "./util/invalidPlugins";
 import { downloadMasterlist, downloadPrelude } from "./util/masterlist";
+import toLootPluginName from "./util/toLootPluginName";
 import toPluginId from "./util/toPluginId";
 
 const MAX_RESTARTS = 3;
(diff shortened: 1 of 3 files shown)


IDCs (commented):
`toPluginId` _is_ the normalization path - so no, this isn't going in.

caseif:
@IDCs The trouble is that LOOT expects filesystem paths, so at least AFAICT passing normalized (lowercase) IDs is incorrect and only happens to work because of Windows's case-insensitivity.

Would it be better to modify the normalization logic itself to resolve filesystem names? I thought it made more sense to use a separate path specifically for LOOT because introducing externalities into a normalization function felt kind of smelly, but it's possible I'm missing knock-on effects from deviating from normalized IDs in this case.

IDCs:
@caseif - you're absolutely correct - Win32 case insensitivity is currently the saving grace of this extension (at least for Windows) and changing it will have knock-on effects (even outside the bundled extensions). Although your PR worked around that for libloot specifically, it added yet another band-aid fix which we would have to review once we start looking at linux implementations specifically. Our target is not to workaround faults in the codebase, it's to explicitly eliminate them.

I wouldn't recommend tackling this just yet as we're planning to change the file system approach entirely through the API. Please understand that we appreciate people trying to help us out to get this out the door, but any workarounds are future maintenance overhead which we will have to tackle down the road.

caseif:
@IDCs That's totally understandable, I appreciate the clarification!

PR #23834, opened 2026-07-30
gamebryo-plugin-management: Search for appdata path in game Wine prefix on Linux

caseif:
This adds a Linux-specific path to `gamebryo-plugin-management` when locating the game's appdata contents. It now searches for files in the game-specific Wine prefix under `compatdata` if `LOCALAPPDATA` is undefined. This allows it to update `plugins.txt` rather than silently failing.

There currently isn't any guard to ensure that the game is specifically running under Proton, but this should harmlessly fail unless the user has previously run the game through Proton in which case the wrong files will be used. This isn't ideal, but it would still silently fail even with a guard since there's no mechanism for loading native appdata files on Linux currently.

### DIFF
--- a/extensions/gamebryo-plugin-management/src/util/gameSupport.ts
+++ b/extensions/gamebryo-plugin-management/src/util/gameSupport.ts
@@ -1,3 +1,4 @@
+import * as nodeFs from "fs";
 import * as path from "path";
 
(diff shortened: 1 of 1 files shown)


IDCs:
Hey @caseif - I'm currently refactoring the extension and this PR is no longer applicable - closing.

PR #23853, opened 2026-07-30
Upgrade React to 18.3.1 (legacy render)

insomnious:
## Summary

This went a lot smoother than expected. The reason is that React 18 has a built-in compatibility mode: keep rendering through the old `ReactDOM.render`, which we do, and the app runs with React 17 behavior, bug for bug. No concurrent rendering, no `createRoot`, no `StrictMode`, no timing changes. This PR swaps which React we ship without changing how anything renders. The flip to `createRoot`, where the actual new behavior lives, is a separate later release so we never land a new React and new rendering semantics in the same update.

This is the lightest touch possible. The real refactor work (replacing `findDOMNode`, migrating off legacy context, dealing with react-bootstrap) is what React 19 will eventually force, and none of it is in here. Being current on 18 makes 19 less urgent: we can schedule that work instead of doing it under pressure.

It was also quick because the React 17 upgrade did the structural work. There is one shared React copy for the host and every extension (extensions don't bundle their own), one version catalog, and the peer rules were already tamed. Flipping the catalog to 18.3.1 installed clean on the first try.

For extensions: published extensions keep working, unmodified. Unlike the React 17 upgrade, nothing was added to or removed from the set of modules extensions can require. I verified this at runtime with three third-party extensions. Authors will notice one thing, which is that their next `npm install` pulls `@types/react` 18 and may surface a few TypeScript errors on rebuild (mostly "declare your `children` prop"). All three error classes with fixes are documented in a new MIGRATION.md section. At runtime, nothing changes for them.

Why bother at all: React 18 unblocks `@headlessui/react` v2 for frontend work, keeps us on supported versions, and 18.3's dev-mode warnings point at exactly the code React 19 would break. Free early warning while we run 18.

## Changes

1. Regenerate stale API report (first commit). `etc/vortex.api.md` had drifted on master; regenerating it separately keeps this branch's API diff purely React-driven.
2. Catalog bump: `react`/`react-dom` → 18.3.1, `@types/react` ^18.3.0 / `@types/react-dom` ~18.3.0, overrides in lockstep, the four React-17 peer escape hatches re-pointed at 18. Zero new escape hatches needed; single React instance and no 17 residue in the lockfile.
3. @testing-library/react 12 → 16 (12 caps at React 17), plus `IS_REACT_ACT_ENVIRONMENT` in the vitest setup. The whole renderer suite passed without touching a single test.
4. @types/react 18 fallout, smaller than feared: the implicit-`children` removal boils down to `children?: React.ReactNode` on 14 shared props interfaces, `StatelessComponent` → `React.FC` (3 sites), and a few `ReactNode` casts. All 50 extension projects compiled clean with zero changes.
5. Typed class `context` declarations. React 18 types `Component.context` as `unknown`; `ExtensionProvider`, `FormFeedback`, and `Modal` now declare their context shape instead of casting.
6. API report regenerated. The React-driven public API delta is ~19 lines, all loosening (`ComponentType<PropsWithChildren<...>>` on registration surfaces). Nothing an existing extension compiles against breaks.
7. Known React 18 deprecation warnings demoted to `console.warn` (dev builds only; production React strips them). Besides log noise, an un-demoted `console.error` during first render would arm the "Vortex failed to start" terminate timer in `MainWindow` before the window shows.
8. Docs: MIGRATION.md "React 18" section for extension authors (with an npm `--legacy-peer-deps` note for react-select's stale peer range), and AGENTS-FRONTEND.md updated (it still said React 16.14).

## Verification

- Build, lint, and typecheck green across all 152 projects. 3305/3305 tests across 253 files. Playwright e2e: 23 passed, and the 3 failures are all confirmed pre-existing on master (stale selectors; the code involved is byte-identical on both branches).
- A from-scratch sample extension compiled against the new types produces zero React-18-introduced errors. The identical setup on React 17's types has one more.
- Full manual smoke (15 areas, log-monitored throughout): boot and game switching, mods table batch ops, downloads, a FOMOD wizard, a complete collection install, deploy/purge, every drag-and-drop surface (gamebryo load order, dependency lines, groups editor, BUTR load order), dialogs including menus inside modals, notifications, profiles, dashboard, live HMR. Community extensions exercised for real: Bannerlord (BUTR), Support Stats, Subnautica 2.
- Log audit over the whole session: three deduped React warnings, all known and benign (`defaultProps` from react-markdown, `unmountComponentAtNode`, and `validateDOMNesting`, which predates React 17). Zero React 18 regressions found.

## Known items / follow-ups

- createRoot is not in this PR on purpose. It's where React 18's real behavior change lives (auto-batching), and it gets its own later release with a focused audit. A spike confirmed the app boots and runs under `createRoot`, and RTL 16 already renders the entire test suite through it.
- The `renderer:api` script has undeclared build prerequisites (`@vortex/shared`, `exe-version`, `icon-extract`); it needs an nx task dependency.
- react-markdown 6.0.3 triggers React's `defaultProps` deprecation; it goes on the React 19 dependency-sweep list.

### DIFF
--- a/src/renderer/src/extensions/mod_load_order/views/LoadOrderPage.tsx
+++ b/src/renderer/src/extensions/mod_load_order/views/LoadOrderPage.tsx
@@ -33,11 +33,13 @@ interface IBaseState {
   loading: boolean;
   updating: boolean;
   sortType: SortType;
-  itemRenderer: React.ComponentType<{
-    className?: string;
-    item: ILoadOrderDisplayItem;
-    onRef: (ref: any) => any;
-  }>;
+  itemRenderer: React.ComponentType<
+    React.PropsWithChildren<{
+      className?: string;
+      item: ILoadOrderDisplayItem;
+      onRef: (ref: any) => any;
+    }>
+  >;
 }
 
 export interface IBaseProps {
@@ -422,7 +424,7 @@ class LoadOrderPage extends ComponentEx<IProps, IComponentState> {
                       apply={this.onApply}
                     />
                   </FlexLayout.Flex>
-                  <FlexLayout.Flex>{infoPanel}</FlexLayout.Flex>
+                  <FlexLayout.Flex>{infoPanel as React.ReactNode}</FlexLayout.Flex>
                 </FlexLayout>
               </DNDContainer>
             </PanelX.Body>
--- a/src/renderer/src/extensions/mod_load_order/views/LoadOrderPage.tsx
+++ b/src/renderer/src/extensions/mod_load_order/views/LoadOrderPage.tsx
@@ -394,7 +394,16 @@ class LoadOrderPage extends ComponentEx<IProps, IComponentState> {
       refresh: () => this.mForceUpdateDebouncer.schedule(),
     });
 
-    const infoPanel = typeof res === "string" ? <DefaultInfoPanel infoText={res} /> : res;
+    // The declared return type of createInfoPanel is string | ComponentType, but
+    // extensions in the wild return ready-made elements, so handle all three shapes.
+    const infoPanel =
+      typeof res === "string" ? (
+        <DefaultInfoPanel infoText={res} />
+      ) : typeof res === "function" ? (
+        React.createElement(res)
+      ) : (
+        res
+      );
 
     const sorted =
       this.state.sortType === "ascending"
@@ -424,7 +433,7 @@ class LoadOrderPage extends ComponentEx<IProps, IComponentState> {
                       apply={this.onApply}
                     />
                   </FlexLayout.Flex>
-                  <FlexLayout.Flex>{infoPanel as React.ReactNode}</FlexLayout.Flex>
+                  <FlexLayout.Flex>{infoPanel}</FlexLayout.Flex>
                 </FlexLayout>
               </DNDContainer>
             </PanelX.Body>
--- a/src/renderer/test-setup.ts
+++ b/src/renderer/test-setup.ts
@@ -1,3 +1,6 @@
+// React 18: opt tests into act() environment (required by RTL 13+)
+(globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
+
 // NOTE(erri120): yes, the library is called "jest-dom" but it works for vitest as well with this import:
 // https://www.npmjs.com/package/@testing-library/jest-dom#with-vitest
 import os from "node:os";
--- a/src/renderer/test-setup.ts
+++ b/src/renderer/test-setup.ts
@@ -1,6 +1,3 @@
-// React 18: opt tests into act() environment (required by RTL 13+)
-(globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
-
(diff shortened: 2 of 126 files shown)


src/renderer/test-setup.ts
```diff
@@ -1,3 +1,6 @@
+// React 18: opt tests into act() environment (required by RTL 13+)
+(globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
```
erri120: > Eventually, we expect testing libraries will configure this for you automatically. For example, the [next version of React Testing Library has built-in support for React 18](https://github.com/testing-library/react-testing-library/issues/509#issuecomment-917989936) without any additional configuration.

https://react.dev/blog/2022/03/08/react-18-upgrade-guide#configuring-your-testing-environment

Aragas:
With new react code if we already have `import * as React from "react";` we should just keep using `React.*` references. Otherwise we should actually add the imports correctly via `import { PropsWithChildren, ... } from "react";`

I see that there are a few places where we added old style references like `src/renderer/src/types/IExtensionProvider.ts`, ideally we should not do that

src/renderer/src/extensions/mod_load_order/views/LoadOrderPage.tsx
```diff
@@ -422,7 +424,7 @@ class LoadOrderPage extends ComponentEx<IProps, IComponentState> {
                       apply={this.onApply}
                     />
                   </FlexLayout.Flex>
-                  <FlexLayout.Flex>{infoPanel}</FlexLayout.Flex>
+                  <FlexLayout.Flex>{infoPanel as React.ReactNode}</FlexLayout.Flex>
```
Aragas: As I understand the issue is because of this code
`const infoPanel = typeof res === "string" ? <DefaultInfoPanel infoText={res} /> : res;`
We should replace the string with an inline react component, so we won't need `as` casting. Same issues were found in other places

insomnious:
> With new react code if we already have `import * as React from "react";` we should just keep using `React.*` references. Otherwise we should actually add the imports correctly via `import { PropsWithChildren, ... } from "react";`
> 
> I see that there are a few places where we added old style references like `src/renderer/src/types/IExtensionProvider.ts`, ideally we should not do that

I've updated one and left the other. Left a note in agents-frontend as well about not mixing styles and 'fix when we touch' comes into play here.

src/renderer/src/extensions/mod_load_order/views/LoadOrderPage.tsx
insomnious: Removed the need for the cast. This was simplest way that won't affect that API

Aragas (dismissed):

src/renderer/test-setup.ts
insomnious: I've set `globals: true` in `vitst.config.mts` and turned off the manual flag. I'll do a separate PR to remove all the now redundant `cleanup()`

Aragas (approved):

PR #23858, opened 2026-07-30
Health Check fixes and polish

Al12rs:
- `skipDisabledRequirements` is now always sent, so the mod-level query no longer varies with the file-requirements Unleash flag. A flag flip re-runs only the file check instead of both.
- A resolved detail page returned only once every check had finished; it now waits on its own.
- "Adult", "Installed", "Enabled", "Disabled" and the "And" divider were hardcoded English; now translated, with the local divider replaced by the shared one.
- Dropped a stale comment claiming the `install-uninstalled` category isn't produced.

Closes LAZ-852

### DIFF
--- a/src/renderer/src/extensions/nexus_integration/util/graphQueries.ts
+++ b/src/renderer/src/extensions/nexus_integration/util/graphQueries.ts
@@ -5,8 +5,6 @@ import type {
   IModFileQuery,
 } from "@nexusmods/nexus-api";
 
-import { FlagService } from "@/FlagService";
-
 const revisionInfo: IRevisionQuery = {
   id: true,
   revisionNumber: true,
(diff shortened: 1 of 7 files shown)


src/renderer/src/extensions/nexus_integration/util/graphQueries.ts
```diff
@@ -233,15 +231,13 @@ export const MOD_FILE_INFO: Partial<IModFileQuery> = {
...
+ * requirement has superseded are always skipped, so the mod-level check can never report a
+ * requirement the file-level one owns.
  */
 export function getModRequirementsInfo(): IModRequirementsQuery {
-  const flagEnabled =
-    FlagService.instance?.getFlag("vortex-file-requirements-health-check") !== undefined;
-
-  const baseQuery: IModRequirementsQuery = {
+  return {
+    $filter: { skipDisabledRequirements: true },
```
erri120, line 240: This is no longer needed?

erri120 (dismissed):

erri120 (approved):

PR #23873, opened 2026-08-03
Vortex UI fixes - Laz 873

r1chm8:
**Games search bar** — a clear `×` appears inside the field once it has a value, subdued until hovered, matching the design. Clearing empties the field and returns focus to the input, and the input only takes the extra right padding while the button is there so text can't run underneath it. `Search` moves to `components/search/` with a colocated `Search.test.tsx` (7 tests), matching the shape of health_check's `entry_actions`.

**Input loses its `size` prop** — `sm` is the only size now, so those metrics fold into `.nxm-input` and the type scale drops to `body-sm`. `size="sm"` comes off all eight call sites; `JumpToPage` was the only input still on the old default, so the pagination page field gets smaller.

**Page content cap widens to `max-w-8xl`** — Tailwind's container scale stops at `7xl`, so this adds `--container-8xl` (88rem, keeping its 8rem step) and points `PageContent`'s cap at it.

**Health check** — the premium upsell banner no longer shows while the visible panel is on its empty state, which also stops the shown event firing there. The page's eye, check-circle and cog icons move to their outline variants, matching the row actions and the success pill.

Also drops a duplicated react-i18next stub: six test files had it in two slightly different shapes, now `test-utils/i18nMock`.


insomnious (dismissed):

src/renderer/src/test-utils/i18nMock.ts
erri120, line 1: This highlights an issue about our testing method: finding elements by text contents. This is fine for E2E tests because they emulate the full user experience but as you've seen we run into issues for unit tests. The translation keys should not be used to look up elements because they shouldn't be clear form text and can be assigned to multiple elements.

If you want to get a single element, the best approach is still with IDs or roles are fine as well.

erri120 (changes requested):

Aragas (dismissed):

src/renderer/src/test-utils/i18nMock.ts
r1chm8: I've removed the mock and added testid's everywhere it was being used.

erri120 (approved):

PR #23884, opened 2026-08-04
Health Check refresh fixes

Al12rs:
- Health check should now correctly refresh on login state change
- Health check should now correctly refresh on enabled state change
- Health check should now correctly debounce multiple triggers, instead of running multiple checks at the same time.
- Health check should now correctly refresh on mods being removed
- Health check should now correctly refresh on downloads being removed or added

### DIFF
--- a/src/renderer/src/extensions/health_check/core/HealthCheckRegistry.ts
+++ b/src/renderer/src/extensions/health_check/core/HealthCheckRegistry.ts
@@ -19,11 +19,18 @@ import type { HealthCheckId } from "../types";
 import { runPerModCheck } from "./perModRunner";
 
 export class HealthCheckRegistry {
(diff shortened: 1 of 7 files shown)


src/renderer/src/extensions/health_check/core/HealthCheckRegistry.ts
erri120, line 1: Generic Debouncer we have in the shared project doesn't work for this?

Al12rs: I had a look at that, it would have required some changes to the debouncer to know whether the things was currently running. Easier to add these few lines here.

erri120 (approved):

Issue #13, opened 2017-02-27
Current Source - XML Installers getting stuck


DuskDweller:
There's a few XML Installers getting stuck installing **after** selecting the install options: 

http://www.nexusmods.com/fallout4/mods/4640/?

http://www.nexusmods.com/fallout4/mods/4772/?

http://www.nexusmods.com/fallout4/mods/10654/?

PR #23905, opened 2026-08-07
Fix renderer OOM crash loop during large collection installs

### LINKED ISSUE #23904: Renderer OOM crash loop when installing/updating large Cyberpunk 2077 collection (2852 files) — restarts endlessly reporting "collection incomplete"
**Nexus Mods username**
DoomerDGR8

**Description**
Updating a large Cyberpunk 2077 collection ([p0qfwm](https://www.nexusmods.com/games/cyberpunk2077/collections/p0qfwm/), revision #62, 2852 required files) to a new profile causes the Vortex renderer process to crash with an explicit out-of-memory reason roughly 2 minutes into every install/resume attempt. Vortex then auto-restarts, reports the collection as incomplete, and automatically resumes the install — which repeats the same expensive dependency resolution and crashes again after a similar interval. This loop has already run 12+ times (`resume_count: 12` in the attached log) and burned over 2 hours of cumulative install time (`total_duration_ms: 8419790`) without completing.

The crash is confirmed in the log, not a guess:
```
[ERRO] [MAIN] [diag] render-process-gone {"exitCode":-536870904,"reason":"oom"}
[ERRO] [MAIN] render process gone {"exitCode":-536870904,"reason":"oom"}
```

Most of my previously-downloaded archives (from an earlier, working install of revision #54) are being correctly matched against the new revision's manifest — I can see many `"downloaded as dependency"` matches in the log — so this doesn't look like a matching/detection failure. It looks like the sheer size of the dependency graph (2852 entries) exhausts renderer memory before it can finish resolving/installing them, well before getting through even a fraction of the list.

**To Reproduce**
1. Have Cyberpunk 2077 installed (GOG) with a large mod archive collection already downloaded from a previous, successfully-installed revision of a Nexus collection.
2. Reinstall/verify the base game vanilla to confirm it's clean.
3. In Vortex, create a new profile for Cyberpunk 2077.
4. Fetch the updated revision of the same collection (in my case, revision #62 of `p0qfwm`, up from a previously-completed #54) and start/resume the install into the new profile.
5. Wait roughly 1–2 minutes.
6. Vortex's window closes/crashes entirely (confirmed via task manager and the log's `render-process-gone` / `reason: oom` entries), then relaunches on its own and reports the collection as incomplete, auto-resuming into the same crash loop.

**Expected behavior**
The collection installer should be able to resolve a large dependency list (2852 files) against already-present downloads without exhausting renderer memory, and it should not need to restart and re-walk the entire dependency resolution from scratch on every resume. Ideally it should also surface a proper "out of memory" error dialog rather than silently dying and restarting.

**Logs**
See attached `vortex.log`. Key evidence (line numbers as attached):
- `"starting install of collection" {"totalMods":2852,"missing":2852}`
- `"collections_installation_resumed" {"collection_id":"468847","revision_id":"754418","required_total":2852,"resume_count":12,"total_duration_ms":8419790}`
- `"determined unfulfilled dependencies" {"count":2852,"errors":0}`
- `[ERRO] [MAIN] [diag] render-process-gone {"exitCode":-536870904,"reason":"oom"}`
- Startup of the *next* session also shows `"reported native crash from previous session {"dumps":1}"`, confirming this has happened repeatedly before this capture.

I'll attach all `.log` files from `%appdata%/vortex` as requested.

[vortex.log](https://github.com/user-attachments/files/30832455/vortex.log)

**Screenshots**
N/A — the crash is silent from the UI's perspective; Vortex simply closes (visible in Task Manager disappearing entirely) and reopens showing "collection incomplete." No error dialog is shown.

- OS: Windows 11 (build 10.0.26300, x64)
- Vortex Version: 2.5.0-beta.2
- Game: Cyberpunk 2077 (GOG)

**Additional context**
- This same collection completed successfully at revision #54. The problem only appeared when updating to revision #62, which appears to be substantially larger.
- Downloads/mods are stored on a separate drive (`E:\Downloads\Games\Vortex\cyberpunk2077`), game install on another (`G:\Games\GOG\Cyberpunk 2077`) — not on the OS/Vortex install drive.
- Possibly related open issue with a very similar symptom (Vortex idles, then closes completely, reopens reporting "collection incomplete," repeats) on a different game/older version: [#23715](https://github.com/Nexus-Mods/Vortex/issues/23715). That report doesn't include an explicit OOM reason code — mine does, which may help narrow down the cause.
- I'd rather not delete my existing ~150GB+ of downloaded mod archives and start over if avoidable, since most of them are matching correctly against the new revision.

DoomerDGR8:
Fixes #23904

## Problem

Installing or updating a very large collection (reproduced with Cyberpunk 2077 [p0qfwm](https://www.nexusmods.com/games/cyberpunk2077/collections/p0qfwm/) rev 62 — **2852 required files** — against a library of ~2900 existing archives) crashes the renderer with `render-process-gone {"reason":"oom"}` roughly 60–120s into every attempt. Vortex silently relaunches, reports the collection incomplete, and each resume recomputes everything and crashes identically — an endless loop (`resume_count: 12`, ~2.3h, net progress ≈ 0 in the issue's log).

## Root cause

The dependency pipeline contains several `O(refs × downloads/mods)` hot paths that allocate fresh objects per probe. At 2852 references × ~3000 archives that is ~8.5 million probes per pass, and several passes run repeatedly:

1. **Per-probe allocation in download matching** — `lookupFromDownload()` (`util/dependencies.ts`) built a fresh lookup object (plus two `Set`s and an identifier bundle in `findDownloadByRef`) for every download on every probe. Called per rule during resolution, per dependency in the post-phase ready-download scan (`doInstallDependenciesPhase`'s `finally`, which begins right at the `done installing dependencies` log line — matching the observed crash ~30s later), and again in requeue/stall-rescue scans.
2. **Unbounded poll-tick scan** — `pollAllPhasesComplete` ticks every 500ms and called `driveSelectedOptionals` → `selectedOptionalRules`, which runs a `findModByRef` scan over **all installed mods** per candidate optional rule, on every tick, for the whole install. The cheap session-status filter ran *after* the expensive scan.
3. **Retained metadata** — every dependency kept its full metadb lookup-result list alive for the entire install, though only `lookupResults[0]` is ever consumed.
4. **O(n²) rule updates** — `updateRules` dispatched `removeModRule` + `addModRule` per dependency; each dispatch clones the collection mod's full 2852-entry rules array and pushes a separate diff through the persist pipeline (the `persist:diff` ~1/s churn in the log).

The allocation rate outruns GC and the renderer heap dies. Small collections never notice because every term is small.

## Changes

- `util/dependencies.ts`
  - Memoize `lookupFromDownload` and the `findDownloadByRef` identifier bundle per download object via `WeakMap` (redux state objects are immutable, so object identity is a safe key).
  - Retain only `lookupResults[0]` per dependency (the only entry consumers read).
  - Resolve rules in bounded batches of 50, with a DEBG `dependency resolution batch done {resolved, total}` line so future user logs show resolution progress.
  - `tagDuplicates` skips the collateral scan for deps without a lookup result.
- `InstallManager.ts`
  - `updateRules` collects all rule updates into **one** `batchDispatch` (DEBG `batch updating dependency rules {count}`); `updateModRule` shares the same logic.
  - `driveSelectedOptionals` filters by session status (`pending`) **before** the per-rule `findModByRef` scan.
- **Visible OOM handling** (issue's "silently dying and restarting" point): on `render-process-gone` with `reason === "oom"`, the main process writes a `renderer-oom.json` marker to userData (`MainWindow.ts`); on next startup the collections extension consumes it and shows an error notification instead of silently restarting into the same crash (`collections/index.ts`). The collection stays paused until explicitly resumed.

## Verification

- **New stress test** (`util/dependencies.stress.test.ts`): 3000 refs × 3000 downloads, the crash shape. Heap growth **~234 MB → ~10 MB** with the memoization; suite runs ~6× faster.
- **Unit tests** for the memoization identity/invalidation semantics; full `mod_management` + `collections` renderer suites pass (37 files / 506 tests).
- **Real-world repro**: packaged this branch and resumed the exact failing collection (p0qfwm rev 62, resume #13). Resolution completed all 2852 refs in ~74s at a steady ~50 refs/s with 0 errors, sailed past the point where all 12 previous attempts OOMed, and proceeded into downloads/installs normally (124 existing archives matched, fresh downloads + installs running in parallel).
- Small-collection behavior unchanged: batching degenerates to a single batch below 50 rules; single-dep rule updates take the same code path as before.

## Not addressed (possible follow-up)

Resolution still restarts from zero on every resume (`missing: 2852` each time). With this fix a re-resolution is cheap enough not to OOM, but persisting resolution progress across resumes would still save time on very large collections.

### DIFF
--- a/src/main/src/MainWindow.ts
+++ b/src/main/src/MainWindow.ts
@@ -1,4 +1,4 @@
-import { appendFileSync } from "node:fs";
+import { appendFileSync, writeFileSync } from "node:fs";
 import * as path from "path";
 import { pathToFileURL } from "url";
 
@@ -176,6 +176,25 @@ class MainWindow {
           exitCode: details.exitCode,
           reason: details.reason,
         });
+
+        if (details.reason === "oom") {
+          // leave a marker for the relaunched renderer: an OOM during e.g. a large
+          // collection install would otherwise silently restart into the exact same
+          // crash. The collections extension picks this up on startup and warns the
+          // user instead of letting the install resume into a crash loop.
+          try {
+            writeFileSync(
+              path.join(app.getPath("userData"), "renderer-oom.json"),
+              JSON.stringify({
+                reason: details.reason,
+                exitCode: details.exitCode,
+                timestamp: Date.now(),
+              }),
+            );
+          } catch {
+            // diagnostics must never throw
+          }
+        }
 
         // hard renderer crashes never reach the JS error handlers, so this
         // is the only place they can be reported
--- a/src/renderer/src/extensions/collections/index.ts
+++ b/src/renderer/src/extensions/collections/index.ts
@@ -1,3 +1,4 @@
+import { readFileSync, unlinkSync } from "node:fs";
 import * as path from "path";
 import { pathToFileURL } from "url";
 
@@ -1236,8 +1237,51 @@ async function checkVoteRequest(api: IExtensionApi): Promise<number> {
   return TIME_BEFORE_VOTE - elapsed;
 }
 
+/**
+ * checks for the marker the main process writes when the renderer died with an
+ * out-of-memory crash (see MainWindow render-process-gone handler). If the previous
+ * session OOMed while a collection install was incomplete, resuming right away would
+ * very likely crash the same way again - so instead of silently restarting into a
+ * crash loop, surface a visible error and leave the collection paused until the user
+ * explicitly resumes.
+ */
+function checkRendererOomMarker(api: IExtensionApi) {
+  const markerPath = path.join(getVortexPath("userData"), "renderer-oom.json");
+  let marker: { timestamp?: number } | undefined;
+  try {
+    marker = JSON.parse(readFileSync(markerPath, "utf8"));
+    unlinkSync(markerPath);
+  } catch {
+    // no marker (the usual case) or unreadable - nothing to report
+    return;
+  }
+
+  // ignore stale markers (e.g. crash long ago, unrelated to this session)
+  const MARKER_MAX_AGE_MS = 30 * 60 * 1000;
+  if (typeof marker?.timestamp !== "number" || Date.now() - marker.timestamp > MARKER_MAX_AGE_MS) {
+    return;
+  }
+
+  log("warn", "previous session ended in a renderer OOM crash", {
+    crashedAt: new Date(marker.timestamp).toISOString(),
+  });
+
+  api.sendNotification({
+    id: "renderer-oom-collection-install",
+    type: "error",
+    title: "Vortex ran out of memory",
+    message:
+      "The previous session crashed because Vortex ran out of memory. " +
+      "If this happened while installing a large collection, resuming it " +
+      "immediately may crash again - consider restarting Vortex before resuming, " +
+      "and please report the issue if it persists.",
+  });
+}
+
 function once(api: IExtensionApi, collectionsCB: () => ICallbackMap) {
   const { store } = api;
+
+  checkRendererOomMarker(api);
 
   const applyCollectionModDefaults = new Debouncer(() => {
     const gameMode = selectors.activeGameId(state());
--- a/src/renderer/src/extensions/mod_management/InstallManager.ts
+++ b/src/renderer/src/extensions/mod_management/InstallManager.ts
@@ -3461,10 +3461,19 @@ class InstallManager {
           }
         : undefined;
     const stagingPath = installPathForGame(state, session.gameId);
+    // filter by session status FIRST: this runs on every 500ms poll tick and
+    // selectedOptionalRules performs a findModByRef scan over ALL installed mods per
+    // candidate rule. Doing that for the full rule set of a several-thousand-mod
+    // collection on every tick generated enough allocation churn to OOM the renderer;
+    // the cheap session-status lookup reduces the expensive scan to genuinely
+    // pending optionals (typically none or a handful).
+    const pendingRules = (collectionMod.rules ?? []).filter(
+      (rule) => session.mods[modRuleId(rule)]?.status === "pending",
+    );
     const pending = selectedOptionalRules(
-      collectionMod.rules ?? [],
+      pendingRules,
       state.persistent.mods[session.gameId] ?? {},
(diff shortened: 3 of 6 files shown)


DoomerDGR8:
I was able to fully install the collection that was failing for me earlier today and the previous leftover huge files from an older version also help preventing too much re-downloading.

IDCs:
Thanks for this, and especially for the log on #23904. That log is what made the root cause findable, so this was useful even though we're closing the PR.

We profiled each change against the shape from #23904 (2852 references, ~3000 local archives). Dependency resolution retains about 13.5 MB there, against a 4 GB renderer heap cap, so it isn't where the memory goes. The memoisation didn't reduce retained heap (13.5 to 13.9 MB); the batched gather didn't change it at all, since `gatherDependencies` already bounds
concurrency and that limiter's whole queue retains under 2 MB; trimming `lookupResults` saved 8 MB. The stress test's "234 MB to 10 MB" is measuring GC scheduling rather than allocation, because `global.gc` is undefined under our vitest config.

The real cause only appears on a revision update, which is what you were doing (54 to 62). Your archives carry revision 54's reference tags, so every dependency takes the tag reconciliation branch in `InstallManager`, which rewrites the rule and emits no log line. That's why your log shows only 124 "downloaded as dependency" entries for 2852 dependencies. Each rewrite dispatches `removeModRule` plus `addModRule`, the persist middleware computes a diff per dispatch, and because `computeStateDiff` treats arrays as leaves each diff carries a full copy of the 2852 entry rules array (~462 KB). Nothing coalesces them, so the same state path accumulates thousands of whole array snapshots: 5704 operations, roughly 2.6 GB.

That's also why your build stopped crashing. Batching the two dispatches into one halves the queued volume, 2635 MB to 1318 MB, which drops it under the cap. Real, but a factor of two on a quantity that grows with the square of collection size, so it moves the failure point from roughly 2850 members to about 5700 rather than removing it.

The fix needs to be in the middleware instead, potentially coalescing queued operations per path so repeated writes cost the size of the state rather than size times write count. That covers every caller, not just rules. (I'm looking into it)

Closing with thanks. The report and log were the valuable half, and the fix will reference them.

PR #23915, opened 2026-08-10
refactor(header): build the top nav buttons on Button (LAZ-904)

r1chm8:
The header's four icon buttons — panel toggle, notifications, help and the avatar — were a bespoke `IconButton` plus a bespoke `ActionButton`. Both are gone; each is now the shared `Button` with `appearance="weak" brand="neutral"`, which already reproduced what they hand-rolled: the same hover surface, the same treatment while a menu or popover is open, and, since the size scale shifted, the same 28px box.

Each keeps its tooltip by wrapping the trigger rather than nesting inside it. That ordering matters: Headless UI returns focus to its button after the click that opens a surface, and a tooltip rendered within the trigger reads that as a cue to reappear — on top of the panel it just opened. The three that open something suppress the tooltip while it is up, reading the state from the `aria-expanded` Headless UI passes into the `as` component, so no caller has to thread it down.

`title` becomes `aria-label` throughout. On a `Button` it would have reached the DOM as the native tooltip and shown twice.

The notification badge and the avatar are what stopped these being plain `Button`s, so each keeps a small wrapper: the badge is a sibling of the button rather than a child, since `Button` builds its own content, and the avatar rides in the `leftIcon` slot as an `Image`, which brings the inset border and a broken-image fallback the bare `img` never had.

Not converted: the window controls. They already match the Figma, and `Button` has no `danger` brand for the red close hover, no size on its scale for a 44px title-bar control, and would need its radius overridden to sit flush.


insomnious:
image
image

They are slightly diff than the original, so I'm assuming @captainsandypants is happy with the subtle shift. Avatar button is now square and not round for example. Icon is a diff size for the help.

I'm happy with the consolidation of random buttons. I've approved as long as Laurence is happy

insomnious (approved):

PR #23918, opened 2026-08-10
Added showItemInFolder

Pickysaurus:
Added new shell command `showItemInFolder`. When triggered this will attempt to open the folder containing the passed file, with the target file selected. 

On awkward systems (like some Linux configs) it _should_ fallback to simply opening the folder. 

This is not used in Vortex proper right now, but will be useful in cases where we want to show the user a specific file on disk. This could include a mod file, a plugin, or even the Vortex log files.


Pickysaurus:
@IDCs Had a look at this today and the changes to vortex.api.md don't appear to be correct. 

I should note this file is auto-generated when building Vortex and was not edited directly.

src/main/src/open.ts
```diff
@@ -32,3 +32,19 @@ export function openFile(filePath: string): void {
...
+  } catch (err) {
+    log("error", "failed to show item in folder. Falling back to opening folder.", {
+      filePath,
+      resolvedPath,
+      error: err,
+    });
+    const parentDir = path.dirname(resolvedPath);
+    openFile(parentDir);
+  }
+}
```
IDCs, line 38: Plain path instead of the file:// link, and the try/catch can go (this call never throws):

```suggestion
export function showItemInFolder(filePath: string): void {
  shell.showItemInFolder(path.resolve(filePath));
}
```

src/shared/src/types/ipc.ts
```diff
@@ -232,6 +232,9 @@ export interface RendererChannels extends RendererCallbackChannels {
   /** Opens the file using the default application for the file extension */
   "shell:openFile": (filePath: string) => void;
 
+  /** Opens the file in its parent folder using the file explorer */
+  "shell:showItemInFolder": (filePath: string) => void;
```
IDCs: Matching the preload.ts wording - this selects the file rather than opening it:

```suggestion
  /** Opens the OS file manager with the file selected */
  "shell:showItemInFolder": (filePath: string) => void;
```

IDCs (approved):

PR #23924, opened 2026-08-11
add mod-list snapshot telemetry (LAZ-701)

IDCs:
Emit a per-game mod-list snapshot to the internal v3 telemetry ingest endpoint (POST /v3/telemetry/mod-lists) on game launch and on deploy, gated on analytics consent and being logged in.

The snapshot is a flat payload: instance_id, game_id, captured_at, vortex_version, and one entry per installed mod (source, mod_id, file_id, version, enabled). mod_id/file_id are Nexus UIDs, null for non-Nexus mods. The server reads the account id from the auth token, so no user_id is sent.

Request types are generated from a vendored internal OpenAPI fragment (nexus-api-v3/schema/internal/telemetry.yaml) via a second codegen output, so the payload is checked against the endpoint schema at build time. A typed internal v3 client (createNexusV3InternalClient / submitModLists) carries the request and shares auth and transport with the public client.

closes LAZ-701

### DIFF
--- a/packages/nexus-api-v3/schema/internal/telemetry.yaml
+++ b/packages/nexus-api-v3/schema/internal/telemetry.yaml
@@ -1,0 +1,157 @@
+# openapi/specs/telemetry/telemetry_openapi.yaml (path + components)
+# openapi/specs/shared/components.yaml (ProblemDetails, referenced by the error responses)
+
+openapi: 3.0.3
+info:
+  title: Telemetry — Vortex mod-list ingestion (internal)
+  version: 1.0.0
+
+paths:
+  /telemetry/mod-lists:
+    post:
+      tags:
+        - telemetry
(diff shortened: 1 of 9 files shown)


packages/nexus-api-v3/schema/internal/telemetry.yaml
erri120, line 1: This isn't exposed on the normal openapi.yaml file?

IDCs: Nope, backend decided to keep it internal and share the spec fragment. More info [here](https://app.notion.com/p/LAZ-701-Next-steps-and-work-breakdown-3a61a4be2f4c80428b55e9499b2fae68) if interested.

insomnious (approved):
Checked end to end, my mod lists are showing up in Clickhouse

PR #23926, opened 2026-08-11
Consolidate extension queries

erri120:
Part of LAZ-690.

Cleans up and consolidates all queries done on the various extension states.

### DIFF
--- a/extensions/extension-dashlet/src/ExtensionsDashlet.tsx
+++ b/extensions/extension-dashlet/src/ExtensionsDashlet.tsx
@@ -243,7 +243,7 @@ class ExtensionsDashlet extends ComponentEx<IProps, IExtensionsDashletState> {
 function mapStateToProps(state: types.IState): IConnectedProps {
   return {
     extensionState: state.app.extensions,
-    extensions: state.session.extensions.available,
+    extensions: (state.session as any).extensions.available,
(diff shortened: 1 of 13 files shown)


extensions/extension-dashlet/src/ExtensionsDashlet.tsx
```diff
@@ -243,7 +243,7 @@ class ExtensionsDashlet extends ComponentEx<IProps, IExtensionsDashletState> {
 function mapStateToProps(state: types.IState): IConnectedProps {
   return {
     extensionState: state.app.extensions,
-    extensions: state.session.extensions.available,
+    extensions: (state.session as any).extensions.available,
```
IDCs, line 246: Can probably extend the session hive in IState to avoid `any` here.

erri120: I moved the session type for extensions into a renderer-local definition. This module doesn't have access to that type anymore which is why it needs any. Not much of an issue as this extension will likely be yeeted soon anyways, either removed entirely or internalized

IDCs: Nevermind - spotted the declaration.

IDCs (approved):

PR #23932, opened 2026-08-11
Linear pipeline #2

Aragas:
Continuation of https://github.com/Nexus-Mods/Vortex/pull/23928

Added a summary in the action so we can see what was included into the change. Also fixed some issues with the original code.

Examples:
* v2.6.0-beta.1 result - https://github.com/Nexus-Mods/Vortex/actions/runs/31503574184
* v2.5.0 result - https://github.com/Nexus-Mods/Vortex/actions/runs/31506966319
* v2.5.0-beta.1 result - https://github.com/Nexus-Mods/Vortex/actions/runs/31506940774
* v2.5.0-beta.2 result - https://github.com/Nexus-Mods/Vortex/actions/runs/31506953798


.github/workflows/linear-release.yml
```diff
@@ -64,28 +66,118 @@ jobs:
...
+            git log "$range" --merges --format='%H' | while read -r sha; do
+              subject=$(git show -s --format='%s' "$sha")
+              num=$(printf '%s' "$subject" | sed -nE 's/^Merge pull request #([0-9]+).*/\1/p')
+              if [ -z "$num" ]; then
+                continue
+              fi
+              title=$(git show -s --format='%b' "$sha" | sed -n 1p)
+              echo "- [#$num]($REPO_URL/pull/$num) ${title:-$subject}"
+            done
+          } >> "$GITHUB_STEP_SUMMARY"
```
IDCs: I think that once a workflow step goes over 10 LoC - it's time to separate the functionality into a script that can be tested bespoke of the workflow. If anything goes wrong here and the reasoning behind it is not obvious, we would have to do exactly that anyway.

Aragas: This will introduce more lines, but I believe that this will be more readable and better testable this way

Aragas: * I've switched to javascript

.github/actions/linear-release/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "@vortex/linear-release-action",
+  "version": "0.0.0",
+  "private": true,
+  "type": "commonjs",
+  "description": "GitHub Action: resolve the commit-scan range for the Linear release sync and write the run summary.",
+  "scripts": {
+    "build": "node ./build.mjs",
+    "test": "pnpm exec vitest run",
+    "typecheck": "pnpm tsc",
+    "lint": "pnpm exec oxlint --quiet",
```
IDCs, line 12: Just want to confirm that this isn't by accident - you didn't include an `.oxlintrc.json` file. So the built-in defaults will be running.

Obviously that means that you're going to be linting the `dist/` folder too.

.github/actions/linear-release/dist/index.js
```diff
@@ -0,0 +1,37 @@
...:`CONNECT`,path:e.host+`:`+e.port,agent:!1,headers:{host:e.host+`:`+e.port}});e.localAddress&&(i.localAddress=e.localAddress),i.proxyAuth&&(i.headers=i.headers||{},i.headers[`Proxy-Authorization`]=`Basic `+new Buffer(i.proxyAuth).toString(`base64`)),m(`making CONNECT request`);var a=n.request(i);a.useChunkedEncodingByDefault=!1,a.once(`response`,o),a.once(`upgrade`,s),a.once(`connect`,c),a.once(`error`,l),a.end();function o(e){e.upgrade=!0}function s(e,t,n){process.nextTick(function(){c(e,t,n)})}function c(i,o,s){if(a.removeAllListeners(),o.removeAllListeners(),i.statusCode!==200){m(`tunneling socket could not be established, statusCode=%d`,i.statusCode),o.destroy();var c=Error(`tunneling socket could not be established, statusCode=`+i.statusCode);c.code=`ECONNRESET`,e.request.emit(`error`,c),n.removeSocket(r);return}if(s.length>0){m(`got illegal response body from proxy`),o.destroy();var c=Error(`got illegal response body from proxy`);c.code=`ECONNRESET`,e.request.emit(`error`,c),n.removeSocket(r);return}return m(`tunneling connection has established`),n.sockets[n.sockets.indexOf(r)]=o,t(o)}function l(t){a.removeAllListeners(),m(`tunneling socket could not be established, cause=%s
```
IDCs, line 1: Surely `./dist` should be gitignored.

Aragas: ./dist should not be ignored because GitHub does not compile the javascript at runtime, it expects it to be compiled before

IDCs (dismissed):

IDCs (approved):

PR #23935, opened 2026-08-12
Rework `ExtensionInfo` parsing and `IExtensionState` building

erri120:
Part of LAZ-690.

This PR has potential breaking changes:

- `state.app.extensions` object is now keyed on generated short-ids meaning any previous by-name or by-bullshit lookup will fail. The previous PR #23926 and this PR should've changed all lookups to proper queries.
- the `info.json` file is no-longer the source of truth and Vortex stops updating it on disk

### DIFF
--- a/src/renderer/src/ExtensionManager.ts
+++ b/src/renderer/src/ExtensionManager.ts
@@ -38,6 +38,7 @@ import {
 import { suppressNotification } from "./actions/notificationSettings";
 import { setExtensionLoadFailures } from "./actions/session";
 import { setOptionalExtensions } from "./extensions/extension_manager/actions";
+import { parseExtensionInfo } from "./extensions/extension_manager/extensionInfo";
 import type { IModReference, IModRepoId } from "./extensions/mod_management/types/IMod";
 import { IPCDownloadAdapter } from "./IPCDownloadAdapter";
 import { log } from "./logging";
@@ -738,7 +739,7 @@ class ExtensionManager {
   // Pending actions to dispatch when setStore() is called (renderer-only architecture)
   private mPendingDisables: string[] = [];
   private mPendingRemoves: string[] = [];
-  private mPendingAdds: Array<{ extId: string; info: IExtension }> = [];
+  private mPendingAdds: IExtensionState[] = [];
   // Extension-registered persistors for custom hives (e.g., loadOrder -> plugins.txt)
   private mExtensionPersistors: {
     [hive: string]: { persistor: IPersistor; debounce: number };
@@ -974,8 +975,8 @@ class ExtensionManager {
     });
     this.mPendingRemoves = [];
 
-    this.mPendingAdds.forEach(({ extId, info }) => {
-      store.dispatch(addExtension(extId, info));
+    this.mPendingAdds.forEach((state) => {
+      store.dispatch(addExtension(state));
     });
     this.mPendingAdds = [];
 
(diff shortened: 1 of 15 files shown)


IDCs (commented):
We're still checking if an extension is disabled by name in two locations:

https://github.com/Nexus-Mods/Vortex/blob/4de5971d6e350f0e71295cd55098b3a405868e5c/src/renderer/src/ExtensionManager.ts#L2907
https://github.com/Nexus-Mods/Vortex/blob/4de5971d6e350f0e71295cd55098b3a405868e5c/src/renderer/src/ExtensionManager.ts#L2923

src/renderer/src/ExtensionManager.ts
```diff
@@ -3083,7 +3096,19 @@ class ExtensionManager {
     );
     for (const ext of scanned) {
       if (!loadedFromState.has(ext.name)) {
```
IDCs, line 3101: So when `loadUserExtensions` is called; we skip disabled plugins from state. Later down the line we scan the plugins folder for extensions that have not been added to the state (such as disabled extensions); but the scan will pick up the disabled extension and re-add it here; I've been encountering duplicate extension entries and I believe this is where it's coming from. (compounded by the remaining name lookups)

src/renderer/src/ExtensionManager.ts
```diff
@@ -2777,30 +2807,13 @@ class ExtensionManager {
...
     if (indexPath !== undefined) {
-      let info: IExtension = {
-        name: "",
-        author: "",
-        description: "",
-        version: "",
-      };
-      try {
-        info = JSON.parse(
+      const info = parseExtensionInfo(
```
IDCs: Parsing the `info.json` file should still be inside a try/catch block. If JSON.parse fails or the file is malformed/invalid types, it's going to take down the renderer.

```
2026-08-12T11:58:54.681Z [ERRO] [RENDERER] error setting up renderer {"message":"[\n  {\n    \"expected\": \"string\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"version\"\n    ],\n    \"message\": \"Invalid input: expected string, received number\"\n  }\n]","name":"ZodError","stack":"ZodError: [\n  {\n    \"expected\": \"string\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"version\"\n    ],\n    \"message\": \"Invalid input: expected string, received number\"\n  }\n]\n    at parseExtensionInfo (C:\\Projects\\vortex210\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\extensions\\extension_manager\\extensionInfo.ts:15:25)\n    at ExtensionManager.loadDynamicExtension (C:\\Projects\\vortex210\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\ExtensionManager.ts:2810:38)\n    at ExtensionManager.loadUserExtensions (C:\\Projects\\vortex210\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\ExtensionManager.ts:3083:24)\n    at ExtensionManager.prepareExtensions (C:\\Projects\\vortex210\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\ExtensionManager.ts:3055:27)\n    at new ExtensionManager (C:\\Projects\\vortex210\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\ExtensionManager.ts:906:29)\n    at init (C:\\Projects\\vortex210\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\renderer.tsx:550:16)"}
```

IDCs (commented):
`IExtensionState` has no `id` - `emitExtensionInstalled` uses that for `extension_id` in its mixpanel event which will now always be `undefined`

src/renderer/src/extensions/extension_manager/installExtension.ts
```diff
@@ -222,49 +224,6 @@ export async function validateExtension(extPath: string): Promise<void> {
   await Promise.all([stat(path.join(extPath, "index.js")), stat(path.join(extPath, "info.json"))]);
 }
 
-export async function validateInstall(extPath: string, info?: IExtension): Promise<ExtensionType> {
```
IDCs, line 225: Removing installation validation now allows us to install any random archive as an extension.

src/renderer/src/ExtensionManager.ts
erri120: Added try/catch back but with no default value fallback, I really don't want to load invalid extensions.

src/renderer/src/ExtensionManager.ts
erri120: Should be fixed with the query fix.

src/renderer/src/extensions/extension_manager/installExtension.ts
erri120: Re-added the error.

IDCs (approved):

PR #23941, opened 2026-08-13
Let the user choose which actions sit on a toolbar (LAZ-915)

r1chm8:
https://linear.app/nexus-mods/issue/LAZ-915/toolbar-pinning

image

 - A toolbar opts into pinning with a `pinningId` and keeps the user's choices in `settings.toolbars`, keyed like `settings.tables`. Only  decisions are stored, so an action follows its own `pinned` default until the user overrides it — and a default we change later still moves for everyone with no opinion.
- An action needs an `id` to be decided about, since labels are translated. Registered actions take theirs from the untranslated `registerAction` title and ask to be pinned through the options they already pass; on the mods bar that's History, Categories and Manage Rules beside Install From File, Open and Check for Updates.
- Where pinning is offered the overflow menu is permanent and holds the full list, so an unpinned action, or a pinned one that didn't fit, is one click away. Each row carries its own pin toggle: a `role="button"` span, since the row is already a button, tabbable on the row the keyboard is on, and still usable on a disabled row.
- `pinned` previously meant "never collapses", which nothing used. Only the pinning path touches the store, so a toolbar without it subscribes to nothing. Most of the `etc/vortex.api.md` diff is the toolchain catching up on deprecation tags and inlined types.


extensions/mod-dependency-manager/src/reducers.ts
```diff
@@ -62,8 +62,14 @@ export const sessionReducer: types.IReducerSpec = {
...
-    [actions.setHasUnsolvedConflicts as any]: (state, payload) =>
-      util.setSafe(state, ["hasUnsolvedConflicts"], payload.hasUnsolvedConflicts),
+    // The count is what the mods toolbar says in the Manage Rules tooltip; the boolean
+    // stays derived from it for what already reads that.
+    [actions.setUnsolvedConflictsCount as any]: (state, payload) =>
+      util.setSafe(
+        util.setSafe(state, ["unsolvedConflictsCount"], payload.count),
+        ["hasUnsolvedConflicts"],
+        payload.count > 0,
+      ),
```
erri120: `setSafe` is deprecated and should not be used for new code.

erri120 (approved):

PR #23962, opened 2026-08-19
source the extension catalog from the v3 extensions endpoint (LAZ-690)

IDCs:
The catalog comes from GET /v3/vortex/extensions through the typed @vortex/nexus-api-v3 client, fetched at startup and on manual refresh. IAvailableExtension is a local model mapped at one wire boundary; game domain/name resolve from the local Nexus games list, and a game claimed by more than one extension keeps the most endorsed claimant.

- browse dialog drops download/endorsement counts and fetches descriptions per selection via mod info
- translation locales derive from the mod name until the endpoint carries a locale field
- catalog dependency lookup matches by extension name
- readExtensibleDir and the info.json disk scan are removed; theme and locale listings read their base directories directly
- extension_manager/util.ts uses native promises

closes LAZ-690

### DIFF
--- a/etc/vortex.api.md
+++ b/etc/vortex.api.md
@@ -666,47 +666,20 @@ interface IAttributeState {
 
 // @public
 interface IAvailableExtension {
-    // (undocumented)
     author: string;
     // (undocumented)
-    dependencies?: {
-        [key: string]: any;
-    };
-    // (undocumented)
-    description: {
-        short: string;
-        long: string;
-    };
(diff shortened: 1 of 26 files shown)


insomnious:
It's looking great and I couldn't see any regressions. 
- downloaded and installed a new extension from the browser
- changed theme
- changed language
- checked dedupe logic of most endorsements win (as a fallback)

erri120 (approved):

PR #23969, opened 2026-08-19
Persist show bundled extension option in redux instead of in-memory

Aragas:
The show bundled extension option will be reset when we go from one page to another and then back, since the option is not persisted in the redux session

Closes LAZ-933


IDCs (changes requested):
I'll be honest - I wouldn't touch anything extension manager related before LAZ-690 gets merged in.

IDCs:
I reconsidered - this should be fine.

IDCs (approved):

PR #23975, opened 2026-08-20
Fixed table reducer crash on persisted state without `attributes`

Aragas:
Fixes fingerprints b89ffde8 1706359b
Closes LAZ-974

### DIFF
--- a/src/renderer/src/reducers/tables.ts
+++ b/src/renderer/src/reducers/tables.ts
@@ -8,7 +8,8 @@ interface TableAttributeState {
 }
 
 interface TableState {
-  attributes: Record<string, TableAttributeState>;
+  // optional: pre-2.6.0 persisted entries may only contain filter/groupBy/collapsedGroups
+  attributes?: Record<string, TableAttributeState>;
   filter?: Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


src/renderer/src/reducers/tables.ts
```diff
@@ -8,7 +8,8 @@ interface TableAttributeState {
 }
 
 interface TableState {
-  attributes: Record<string, TableAttributeState>;
+  // optional: pre-2.6.0 persisted entries may only contain filter/groupBy/collapsedGroups
```
IDCs, line 11: I'm not sure about this one - if a filter/grouping functor exists, the attributes should exist too. Otherwise something went wrong with attribute registration. Which table is causing this?

Aragas: When we do the registration like this `tables.downloads = { attributes: {...}, filter: {...} }` we don't write the Redux state, it's only persisted if the user does an action - toggles a column or sorts.
Before we used setSafe, which was not caring about whether the state existed or not, but the new code assumes that it exists, so it fails

IDCs (approved):

PR #23993, opened 2026-08-24
updater 4/6: push updater status to the renderer

insomnious:
Merges after #23992.

The renderer used to poll for update status once, 15 seconds after startup. Anything that resolved later was missed until the next launch. Main now pushes every status change to all windows and the renderer renders what it's told. Progress broadcasts are throttled to whole percents.

### DIFF
--- a/src/main/src/extensions/autoupdater/index.test.ts
+++ b/src/main/src/extensions/autoupdater/index.test.ts
@@ -20,7 +20,7 @@ const { autoUpdaterMock, appMock, ipcMock, resolveUpdateMock } = vi.hoisted(() =
       on: vi.fn(),
       removeListener: vi.fn(),
     },
-    ipcMock: { handle: vi.fn(), on: vi.fn() },
+    ipcMock: { handle: vi.fn(), on: vi.fn(), send: vi.fn() },
     resolveUpdateMock: vi.fn(),
   };
 });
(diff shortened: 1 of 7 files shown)


erri120 (changes requested):
The downloader and uploader use a pull-based model where the renderer pulls updates periodically to keep IPC chatter down and to not overwhelm the renderer with updates. It's a much simpler approach as you don't need to deal with reactive updates and you don't have to litter the code with push methods (this PR adds 13 calls to `broadcastStatus` at various places in the code).

I highly recommend keeping the pull model, especially since it doesn't seem like there was a major issue with it that couldn't be fixed.

insomnious:
Fair point, it pulls now. Main keeps the state and the renderer polls `updater:get-status`, same as the downloader and uploader poll their progress. 200ms while a check or download is running or just after a button press, nothing at all otherwise. Every updater action starts in the renderer (launch check, the 4h timer, button presses), so the loop is woken just before anything can happen and an idle Vortex makes no updater IPC.

One wrinkle the transfers don't have: the UI reacts to transitions, not just the latest value. A manual check that finds nothing is `checking` for ~300ms then `idle`, and the "up to date" toast needs to have seen the `checking`. So main numbers each snapshot and a poll asks for everything since the last one it saw. No push channel left.

The transport and the state machine touch the same files, so instead of rebasing 28 commits over a rewritten transport I've folded this into #23994 and retargeted it. Closing this one. The conversion is da76d0233 on that PR, and your review here is linked from its description so it doesn't get lost.

PR #23995, opened 2026-08-24
updater 5/5: docs

insomnious:
Merges after #23994 (describes what the stack ships).

Writes down everything the stack introduced: `docs/updater.md` (the update flow, what auto-downloads and what doesn't, downgrades, with a state diagram) and `docs/updater-testing.md` (mock feed, test builds, env vars, the unsigned-build caveats). Also the overdue packaging cleanup. `docs/packaging/windows.md` now documents commands that exist, `CONTRIBUTE.md` becomes `CONTRIBUTING.md` so GitHub surfaces it, and AGENTS gets a packaging guide plus links to all of it.


AGENTS-PACKAGING.md
erri120, line 1: I'd prefer if we do less AGENTS files and instead use `docs/` for both human and agent documentation.

insomnious: It'll be in a follow up PR, along with other docs changes to bring up to date

Aragas (dismissed):

PR #24000, opened 2026-08-24
replace the database error classes with VortexError kinds (LAZ-747)

IDCs:
database:locked and database:open-failed are declared via module augmentation next to their throw sites in LevelPersist, and startup error handling branches on data.kind.

Narrowing from unknown goes through parseError, which returns an existing VortexError unchanged and correctly typed; a bare instanceof produces VortexError<any> with an untyped data (microsoft/TypeScript#17473). @vortex/shared/errors re-exports the error base, so the augmentation target its docs name resolves the kind map. database:open-failed carries only the path - the underlying reason rides on cause, which the startup dialog reads. Stale FileSystemError doc references cleaned up.

Side-effect: errorToReportableError stringifies object props as JSON, with util.inspect for circular values; template interpolation flattened VortexError.data to "[object Object]" in the crash-report details.

part of LAZ-747

### DIFF
--- a/src/main/src/Application.ts
+++ b/src/main/src/Application.ts
@@ -392,7 +392,7 @@ class Application {
         }
 
         app.quit();
-      } else if (isVortexError(err) && err.data.kind === "database:locked") {
+      } else if (isVortexError(err, "database:locked")) {
         dialog.showErrorBox(
           "Startup failed",
           "Vortex seems to be running already. " +
@@ -400,7 +400,7 @@ class Application {
         );
 
         app.quit();
-      } else if (isVortexError(err) && err.data.kind === "database:open-failed") {
+      } else if (isVortexError(err, "database:open-failed")) {
         dialog.showErrorBox(
           "Startup failed",
           `Vortex couldn't open its application database at:\n\n` +
--- a/src/main/src/Application.ts
+++ b/src/main/src/Application.ts
@@ -2,12 +2,11 @@ import { mkdirSync, statSync } from "node:fs";
 import { readFile, writeFile, rm, stat } from "node:fs/promises";
 import path from "node:path";
 
-import { getErrorCode, getErrorMessageOrDefault, unknownToError } from "@vortex/shared";
+import { getErrorCode, getErrorMessageOrDefault, parseError, unknownToError } from "@vortex/shared";
 import type { IParameters, ISetItem } from "@vortex/shared/cli";
 import {
   DataInvalid,
   DocumentsPathMissing,
-  isVortexError,
   ProcessCanceled,
   UserCanceled,
 } from "@vortex/shared/errors";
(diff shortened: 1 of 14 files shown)


Aragas:
That's a nice round number! 🔥

src/main/src/Application.ts
```diff
@@ -391,21 +392,21 @@ class Application {
         }
 
         app.quit();
-      } else if (err instanceof DatabaseLocked) {
+      } else if (isVortexError(err) && err.data.kind === "database:locked") {
```
Aragas: What if we do something like `isVortexError(err, "database:locked")`?

Aragas: Or perhaps `isVortexError<"database:locked">(err)` (not sure if it's even legal, but I think Florian did something similar?)

IDCs: Yeah, cleaner - I'll have a look

src/shared/src/api/errors.ts
```diff
@@ -1 +1,2 @@
+export * from "../errors/base";
```
erri120, line 1: `isVortexError` should be exported next to `parseError` which is in the root export.

src/main/src/store/LevelPersist.ts
```diff
@@ -24,21 +24,14 @@ function traceWritesEnabled(): boolean {
...
-    this.path = persistPath;
-    this.cause = cause;
+// Database kinds are main-process-specific, declared next to their throw sites.
+declare module "@vortex/shared/errors" {
+  interface VortexErrorKindMap {
+    /** Another process holds the database lock (usually a second Vortex instance). */
+    // eslint-disable-next-line @typescript-eslint/no-empty-object-type -- kinds without payload use {} like the core catalog
+    "database:locked": {};
+    /** Opening the database failed for a reason other than lock contention. */
+    "database:open-failed": { path: string; reason: string };
```
erri120: Error message as `reason` in data is an anti-pattern, should be removed.

src/shared/src/errors/base.ts
```diff
@@ -192,6 +192,27 @@ export class VortexError<out K extends VortexErrorKind = VortexErrorKind> extend
...
+ * typing `data` as `any`.
+ *
+ * Matches live instances only: an error rehydrated from the IPC wire is a
+ * plain `Error` (with `data` reattached) and does not match until LAZ-749
+ * rebuilds wire errors as real `VortexError` instances.
+ *
+ * @public
+ */
+export function isVortexError(err: unknown): err is VortexError;
+export function isVortexError<K extends VortexErrorKind>(
```
erri120:
```suggestion
export function isVortexError<K extends VortexErrorKind = VortexErrorKind>(
```

Should be a generic default.

erri120 (changes requested):

src/shared/src/errors/base.ts
erri120: From the meeting: replace `isVortexError` with `parseError`

erri120 (approved):

PR #24003, opened 2026-08-24
stop reconcilers wiping records when a read comes back short (LAZ-978)

### LINKED ISSUE #23981: Invalid App State
**Your Nexus Mods username**
BottomlessThought23

**Describe the bug**
The short version is, my app state is completely messed up. 
There are three main symptoms:
1. When I open my Skyrim profile for the first time (either by switching to it or launching vortex with it as the active profile), it brings up a "Mods changed on disk" error, citing a large list of mods. Pressing "Apply Changes" doesn't affect this behavior whatsoever. 

2. There is a "ghost mod" at the top of my Skyrim modlist, which states it has never been installed, and has no information about it at all. Attempting to remove this does nothing at all. It previously listed an archive name (unified water), but deleting that archive did not remove the ghost mod. 

3. My Cyberpunk profile, which previously had about 350 mods installed, now shows those mods as "Disabled". It also lists another 300 mods that I have downloaded in the past but subsequently deleted, as "Uninstalled."

**To Reproduce**
Well, I can't be completely sure, but my working theory is that I broke the app state by doing this:

1. Install Skyrim on drive C:, install mods.
2. Uninstall Skyrim.
3. Reinstall Skyrim on drive D:, install mods (I actually used a collection I created)
4. Boom, broken app state. 

**Expected behavior**
I know you don't like it when we skip these, but I really don't know what to say about this... I expected it to not cause so many problems switching the location of the game. I expected to just be able to continue modding. 

**Logs**
[vortex2.log](https://github.com/user-attachments/files/31297300/vortex2.log)
[vortex3.log](https://github.com/user-attachments/files/31297301/vortex3.log)
[vortex4.log](https://github.com/user-attachments/files/31297297/vortex4.log)
[vortex.log](https://github.com/user-attachments/files/31297298/vortex.log)
[vortex1.log](https://github.com/user-attachments/files/31297299/vortex1.log)

**Screenshots**
Image

Image

- OS: Windows 10
- Vortex Version: 2.5.0
- Game: Skyrim, Cyberpunk

**Additional context**
Some mods which appear in the downloads folder are not showing up in the modlist. Even odder, installing these mods from Nexus brings up the "mod already installed" dialogue and adds the mod to the list as if it was always there. 

It's like it's partially stuck in an older version of the app state. The version of the app state from the last time I had Skyrim on the D drive, maybe? (That's complete speculation).
I say partially, because I've been able to install new mods without issue, and they remain installed when closing and re-opening vortex. So it's able to alter the current state, there's just a part of it that's not updating... or something. 

(It does also bring up the "Invalid App State" dialogue when I launch vortex, but clicking Repair there doesn't seem to fix anything)

IDCs:
A prefix read that misses keys, or a folder that isn't there, made the
reconcilers treat absence as "the user deleted everything" and persist it. On
the reporter's store ~142k of 309k live rows were unreadable while intact on disk.

Amplification only - the store-level cause is deliberately not fixed here, see
docs/investigation/state-store-key-order-corruption.md.

- `refreshMods` / `refreshDownloads` don't create or reconcile against a folder
  they can't see, and keep records when it reads empty
- `sanitizeProfile` keeps a profile's enabled-flags when a game's mod table is
  absent - they exist nowhere else
- `setModArchiveId` / `setModInstallationPath` don't mint a record for an unknown
  modId, which is what made archive rebinds leave undeletable no-name rows
- Removal keys off the map key, so a record whose `id` leaf was lost is deletable
- In-flight downloads stay out of the reconciliation
- `getAllKVs` prefix reads no longer drop keys above a literal `zzzzzzzzzzz`
  bound, which hid every non-ascii mod name
- The "Data corrupted" retry clears the stale `getInitMetadata` IPC handler

closes #23981
closes LAZ-978


erri120 (changes requested):
I strongly discourage putting the LevelDB validation and repair functionality into this repository. This is a huge maintainability risk and potential source for tons of bugs that we now take on as responsibility.

This functionality should be in an upstream package. Maybe even in https://github.com/Nexus-Mods/duckdb-level-pivot which we own entirely.

IDCs:
I'll admit I felt the same way halfway through writing the validation logic and I should've stopped there, going to chunk down this PR to stopping the reconcilers on this repo and move the validation/repair functionality to level_pivot.

erri120 (dismissed):

erri120 (approved):

PR #24011, opened 2026-08-25
fix(ui): let the classic notifications popover be closed

r1chm8:
The popover's visibility was `items.length > 0` — whether it had anything to show — so the button under it only ever chose whether notifications past their display time stayed listed. Closing worked by accident: the list emptied as things expired. A notification that never expires therefore had nothing to close it, and since APP-468 (#23226) taught `displayTime` that a notification offering the user a choice must not auto-hide, any of those pinned the popover open for good.

The user's close is now its own state, apart from `open`, which keeps its meaning of whether the expired are still listed. A notification arriving clears it, that being how the popover shows itself unasked and the alternative being that the first thing the user closes is the last they see. Only an id we haven't seen counts, so an activity reporting progress won't reopen what they just shut.

The modern header's popover was never affected: it gates its panel on the Headless UI open state and opens itself by clicking its own button, which is the arrangement arrived at here.

### DIFF
--- a/src/renderer/src/views/NotificationButton.test.tsx
+++ b/src/renderer/src/views/NotificationButton.test.tsx
@@ -1,0 +1,143 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+import type * as ReactReduxTypes from "react-redux";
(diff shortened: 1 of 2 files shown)


Aragas (approved):

Aragas:
Test code is failing @r1chm8

PR #24028, opened 2026-08-26
Report which UI the session is running on app_launched (LAZ-964)

r1chm8:
Gives us visibility into how many users are still on the legacy UI, so the deprecation timing can be a data-backed call.                      

Adds an `is_legacy_ui` property to the existing `app_launched` Mixpanel event, read from `settings.window.useModernLayout` when analytics starts. Because the setting is persisted, this reports each user's *current* mode from their next launch onward — so we get unique users and session share for the legacy UI immediately, with no backfill needed. Segmentable by date in Mixpanel.

No new events and no super property: the launch event already fires once per session, which is the granularity the ticket asks for.

### Notes                                              

- The argument is optional, so existing three-arg construction of `AppLaunchedEvent` still compiles and an absent value reads as unknown rather than reporting a mode.
- `startAnalytics()` runs before `migrate()`, so the event is built from pre-migration state. This is fine: the only migration touching this setting is `enableModernLayout_2_0`, which only runs for users coming from 1.x where the setting didn't exist and defaults to modern — the same value it then writes.
- Property documented in the analytics README for the data team.

### Testing
                                                         
Three tests in `MixpanelAnalytics.test.ts` cover the value reaching `mixpanel.track` in both directions and the omitted case. Full workspace `build`, `lint`, `typecheck` and `test` pass.

### DIFF
--- a/extensions/theme-switcher/src/SettingsTheme.tsx
+++ b/extensions/theme-switcher/src/SettingsTheme.tsx
@@ -270,7 +270,7 @@ class SettingsTheme extends ComponentEx<IProps, IComponentState> {
   };
 
   private toggleLegacyUI = (useLegacy: boolean) => {
-    this.context.api.events.emit("analytics-track-click-event", "Themes", "Toggle legacy UI");
+    this.context.api.events.emit("analytics-track-ui-mode-changed", useLegacy);
     this.props.onSetUseModernLayout(!useLegacy);
   };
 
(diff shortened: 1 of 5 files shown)


Aragas:
We could also add a theme switched event to see whether users look at the Modern UI and switch back and vice versa with Classic UI

r1chm8:
> We could also add a theme switched event to see whether users look at the Modern UI and switch back and vice versa with Classic UI

I've added this in on the second commit.

Aragas (approved):

PR #24042, opened 2026-08-27
ci: switch workflows to pnpm/setup@v2

joshedney:
## Goal

Swap the action reference in every workflow/action that installs `pnpm`, from `pnpm/action-setup@v6.x` to `pnpm/setup@v2`, keeping existing `with:` options (e.g. `cache: true`) unchanged.

Added the `install: false` option to prevent the setup action from running install due to issues with missing dependancies.

It handles the install of pnpm differently. Uses the binary rather than running `npm i -g pnpm`, I also believe that the old Action is being deprecated soon

## Testing

Checked to make sure that current CI actions work as expected


Aragas:
If you look at the action logs you'll see `libatomic1 is already the newest version (14.2.0-4ubuntu2~24.04.1).`
https://github.com/Nexus-Mods/Vortex/actions/runs/33052819097/job/98452301383?pr=24042#step:5:50

So it looks like manual installation is not needed?

Aragas:
To be more precise, we need it in the E2E.yml file, since that runner is our own and it seems to be missing the library

.github/actions/package/action.yml
```diff
@@ -86,7 +86,7 @@ runs:
         echo "Using Node.js version: $nodeVersion"
 
     - name: Install pnpm
-      uses: pnpm/action-setup@v6
+      uses: pnpm/setup@v2
```
Aragas, line 80: Don't we also need `install: false` here?

joshedney:
> If you look at the action logs you'll see `libatomic1 is already the newest version (14.2.0-4ubuntu2~24.04.1).` https://github.com/Nexus-Mods/Vortex/actions/runs/33052819097/job/98452301383?pr=24042#step:5:50
> 
> So it looks like manual installation is not needed?

Correct, I thought we were having issues with that step too, but it is redundant.

.github/actions/package/action.yml
joshedney: I've added it as it was missing. Since we install further down it is needed.

Aragas:
Also, we were previously pinning to a specific version and were updating it manually, but now we're using a generic v2 version. @erri120 any opinion on this?

.github/workflows/e2e.yml
```diff
@@ -52,10 +52,17 @@ jobs:
       - name: Install volta, Node.js and Yarn
         uses: volta-cli/action@v5
 
+      # NOTE: pnpm's standalone binary requires libatomic1, which is missing
+      # by default on the ubuntu-latest runner image
```
Aragas:
```suggestion
      # by default on our custom runner images
```

.github/workflows/format.yml
```diff
@@ -35,9 +19,10 @@ jobs:
       - uses: actions/checkout@v7
       - name: Install volta, Node.js
         uses: volta-cli/action@v5
-      - uses: pnpm/action-setup@v6.0.9
+      - uses: pnpm/setup@v2
         with:
           cache: true
+          install: false
       - name: Fix pnpm node-gyp permissions
         run: find "$PNPM_HOME" -path '*/node-gyp/gyp/gyp_main.py' -exec chmod +x {} \;
```
Aragas: Since we switched to a different way of working with pnpm now I believe that chmod is not required anymore. Same in `main.yml` and `actions-check.yml`

Aragas:
Based on the docs, `pnpm/setup` replaces `actions/setup-node`, so we should do several things:
1. Remove `volta-cli/action`
2. Remove `actions/setup-node` (don't forget about `cache: true`). I also don't think we need `registry-url`? And `Read Node.js version from package.json`

Most likely we'll only use `devEngines.runtime` now and `engines.node` with `volta.node` will be obsolete, but it's not necessary to do in this PR @erri120 am I right here?

.github/workflows/format.yml
joshedney: This has now been removed.

Aragas:
I've updated the loot and bsatk packages, since our custom runner didn't supply npm and they required it previously.
I also introduced a shared composite action so we won't need to duplicate the custom setup logic

erri120 (approved):

PR #24045, opened 2026-08-27
Replace `DownloadError` and `UploadError` with `VortexError`

erri120:
Closes LAZ-747 for good.

I don't like the IPC hacks, but LAZ-749 will handle a proper IPC rework for errors.


src/main/src/downloading/manager.ts
```diff
@@ -251,11 +251,11 @@ export class DownloadManager {
     // that callers awaiting only pause() don't see unhandled rejections.
     // Non-cancellation errors still reject.
     const promise = rawPromise.catch((err) => {
-      if (err instanceof DownloadError && err.code === "cancellation") return;
+      if (isVortexError(err) && err.data.kind === "user-canceled") return;
```
Aragas: I missed initially the context in IDCs PR, but I remember I suggested `isVortexError(err, "user-canceled")`, which seem to be replaced with `parseError`
Should we still use `isVortexError` with my suggested signature then?

src/renderer/src/extensions/mod_management/InstallManager.ts
```diff
@@ -5432,7 +5433,7 @@ class InstallManager {
                     return resolve(id);
                   } else if (error instanceof AlreadyDownloaded) {
                     return resolve(error.downloadId);
-                  } else if (error instanceof DownloadIsHTML) {
+                  } else if (parseError(error).data.kind === "download:is-html") {
```
Aragas, line 5436: Similar here - `parseError(error, "download:is-html")` or something like this

src/main/src/downloading/manager.ts
erri120: Ended up removing `isVortexError` after all and replaced it with an assertion method for tests only.

src/renderer/src/extensions/mod_management/InstallManager.ts
Aragas: Continuing here - always accessing `data.kind` for checking looks not ideal, a second function for VortexErrorKind would be preferable for readability

erri120: Maybe later, I want to get all the old errors moved other first.

IDCs (approved):

PR #24062, opened 2026-08-29
Stop the updater analytics reporting offers that never happened

insomnious:
Two corrections to the updater funnel, both to events added on this branch and not yet released.

All three What's New buttons reported the same source, so there was no telling whether people read the notes before downloading or before restarting. They are now offer, staged and error_retry.

The check outcome was worked out by elimination, so a check landing back on an already staged update counted as a new offer, and background checks were not suppressed either. Anyone leaving an update staged emitted a false offer every four hours. It now reports already_staged, and background checks stay quiet for it as they do for up_to_date.

Tested by driving the updater against a local mock feed and watching the dev log. The events are confirmed arriving in the dev Mixpanel project.

Checked live:

- all four check outcomes, including the new already_staged on a manual check against a staged update
- offered, download_started, download_completed and install_started across all three kinds (patch, update, downgrade)
- both the manual and background branches of every event carrying that flag
- download_failed with retry_offered set, and a check failure reported separately from a download failure
- cancelling a download mid transfer emits nothing, so cancels do not inflate the failure rate
- a background check that finds nothing emits nothing
- events emitted before login are queued and flushed in order once analytics starts

The three new release notes sources are covered by unit tests but were not clicked through after the rename. app_updated and the post_update source still need a packaged install, since justUpdatedFrom is compiled out of dev builds.

### DIFF
--- a/src/renderer/src/extensions/updater/updaterAnalytics.ts
+++ b/src/renderer/src/extensions/updater/updaterAnalytics.ts
@@ -58,20 +58,25 @@ export function createUpdaterAnalytics(deps: UpdaterAnalyticsDeps): UpdaterAnaly
 
   const onTransition = (prev: UpdaterState | null, next: UpdaterState) => {
     // a check settled: report every manual outcome, and background failures
-    // and offers; the background "nothing new" is left out on purpose
+    // and offers; the background "nothing new" is left out on purpose, which
+    // covers landing back on an already-staged update as well as finding
+    // nothing (checks run 4-hourly, so neither is worth repeating forever)
     if (prev?.type === "checking" && next.type !== "checking") {
       const outcome =
         next.type === "error"
           ? "failed"
           : next.type === "idle" || next.type === "disabled"
             ? "up_to_date"
-            : "offered";
-      if (prev.manual || outcome !== "up_to_date") {
+            : next.type === "staged"
+              ? "already_staged"
+              : "offered";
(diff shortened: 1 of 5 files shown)


erri120 (approved):

src/renderer/src/extensions/updater/updaterAnalytics.ts
```diff
@@ -58,20 +58,25 @@ export function createUpdaterAnalytics(deps: UpdaterAnalyticsDeps): UpdaterAnaly
...
+              ? "already_staged"
+              : "offered";
+      const nothingNew = outcome === "up_to_date" || outcome === "already_staged";
+      if (prev.manual || !nothingNew) {
         emit(
           new AppUpdateCheckCompletedEvent({
             manual: prev.manual,
             outcome,
-            error_message: next.type === "error" ? next.message : undefined,
+            error_message: next.type === "error" ? next.message.slice(0, 200) : undefined,
```
erri120: Slice should be done inside the constructor so that callers don't forget.

erri120 (changes requested):

src/renderer/src/extensions/updater/updaterAnalytics.ts
insomnious: Good catch, done in 2ab1dc6ee. Both constructors now call a shared truncateErrorMessage, and the two call sites pass the message through untouched.

erri120 (approved):

PR #24070, opened 2026-08-31
Address Health Check install failure reports 

Al12rs:
- Addresses the post-GA spike in `data_invalid`/`aggregate_error` health-check install failures reported in mixpanel.
- Caps concurrency on the bulk "install all" actions, which previously fired every download+install at once unlike every other bulk-install path in the app. 
- Also stops discarding the real error behind a failed file-access check and reports enough extra detail on `health_check_install_failed` (underlying OS code, AggregateError sub-error codes/count) to diagnose what's left from real data instead of guessing.


erri120 (changes requested):
Best wait on #24045 as it'll conflict. The mixpanel error classification will also change massively as part of the error rework to the point where I'd suggest not pushing those changes at all and only keeping the `MAX_SIMULTANEOUS_INSTALLS` change in this PR.

Al12rs:
Makes sense. I'll remove the two extra commits then

Al12rs:
@erri120 Changed PR to only limit concurrent installs

erri120 (approved):

PR #24071, opened 2026-08-31
docs: move project documentation into docs/

insomnious:
The six `AGENTS-*.md` files at the repo root are gone. Everything in them that was project or coding documentation is now in `docs/`, indexed by [docs/README.md](docs/README.md), and `AGENTS.md` covers only how to operate in this repo: commands, tools, skills, and which doc to read for a task.

They moved because a file only agents read never gets reviewed the way a doc a human opens does, so it rots. `AGENTS-DIRECTORIES.md` had eight wrong paths in sixty lines by the time I moved it. Auditing the rest of `docs/` then turned up another 40 citations pointing at files that do not exist, mostly the packages having moved to `src/<pkg>/src/` without the docs following.

The layout now:

- `docs/` owns all project and coding documentation, whoever the reader is
- `docs/README.md` indexes it, and every new doc gets a line there
- `docs/writing-documentation.md` holds the conventions: naming, structure, tone, and one owner per fact
- The root keeps only the files convention puts there: README, CONTRIBUTING, CODESTYLE, CHANGELOG, RELEASES, SECURITY, SUPPORT, CODE_OF_CONDUCT

Worth looking at first: `docs/I18N_STATUS.md` carries a new warning rather than a fix. It documented the i18next separators as `:` and `.` while the code sets `:::` and `::`, and the migration guide, the locale files and the one migrated example all assume the documented format. Someone who knows the i18n work should decide which side is wrong before more strings are migrated.

The only non-markdown file is `InstallManager.ts`, where three comments pointed at `AGENTS-COLLECTIONS.md`. Comment text only, no behaviour change.

Untested beyond `pnpm run format` and an audit that every cited source path and relative link resolves. It is all markdown apart from those comments.


AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+running Vortex on `127.0.0.1:9222`. Vortex has to be running with remote
+debugging first: `pnpm run dev`, or the VS Code "Debug Electron" profile. Use it
+to inspect live renderer state, console and network rather than adding temporary
+logging, which tends to get committed by accident.
+
+**Linear**, if available in your environment (it's user-scoped, not configured
+by this repo): fetch ticket context by ID instead of asking for a paste. Don't
+create or transition issues unless asked.
+
+## Skills
```
Sewer56:
```
## Skills

`.claude/skills/` holds repo skills. Prefer them over improvising:

- `watch-log` - tailing and investigating Vortex logs
- `changelog` - drafting a `CHANGELOG.md` entry
- `e2e-test` - scaffolding a Playwright E2E spec
```

This will be injected into system prompt already by harness.
Specifically a list of skills.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+`pnpm run test` **excludes the E2E suite** (`@vortex/e2e`), which needs a
+packaged app and a real game install. Passing it is not evidence that E2E
+passes, so say which suite you ran. Run E2E only when asked.
+
+Formatting, import order and Tailwind class order are owned by oxfmt and oxlint.
+Don't hand-fix them; run `pnpm run format` and let it win.
+
+## Committing
+
+A pre-commit hook runs `oxfmt` over every staged file, markdown included, so
```
Sewer56: LLM does not need to know about this.
Wastes tokens.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+- `watch-log` - tailing and investigating Vortex logs
+- `changelog` - drafting a `CHANGELOG.md` entry
+- `e2e-test` - scaffolding a Playwright E2E spec
+
+## Which doc to read
+
+Read the doc for the area you're touching rather than the whole tree.
+[docs/README.md](docs/README.md) is the full index; this table is the shortcut.
+
+| Working on                       | Read                                                                     |
```
Sewer56: This is actually fine.
Just remember there's distinction between README for LLMs and for humans.
For humans, usually much more verbose.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+- Don't write a throwaway script to do what an edit tool does.
+- The default shell is PowerShell on Windows. Anything committed to `scripts/`
+  should be cross-platform Node.
+- Never run the signed `pnpm package` locally; signing secrets are CI-only. See
+  [docs/packaging/windows.md](docs/packaging/windows.md).
+- When explaining code, cite `path:line` so the claim can be checked.
+
+## MCP servers
+
+**chrome-devtools** is configured for this repo (`.mcp.json`) and attaches to a
```
Sewer56: MCP Servers will be injected into system prompt by harness. Only include info LLM does not already know here. e.g. port if not stated.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+
+The hook may also trigger a dependency install mid-commit, so a commit can take
+about 30 seconds and print install output. That's normal.
+
+Don't commit, push, or open a PR unless asked. Leave changes in the working tree
+for review. If you're on `master`, branch first.
+
+## Toolchain
+
+Node and pnpm are pinned exactly: `packageManager` pins pnpm 11.10.0 with an
```
Sewer56: Should not be injected into every single LLM session no matter what user does.
Not relevant >99% of the time

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+Project and coding documentation lives in [docs/](docs/README.md). This file
+covers only how to operate in this repo as an agent: commands, tools, and which
+doc to load for a task. It deliberately duplicates nothing from `docs/`.
+
+## Verification
+
+Run these from the repo root. They fan out through Nx to every affected package,
+so there is no need to change directory into one.
+
+- `pnpm run build` - **already runs lint and typecheck**, so you don't need to
```
Sewer56: Make a single `PowerShell` and `shell` script for verify.
Then ask LLM to run that.
More efficient time and token wise. Saves calls.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+about 30 seconds and print install output. That's normal.
+
+Don't commit, push, or open a PR unless asked. Leave changes in the working tree
+for review. If you're on `master`, branch first.
+
+## Toolchain
+
+Node and pnpm are pinned exactly: `packageManager` pins pnpm 11.10.0 with an
+integrity hash, and `devEngines.runtime` plus the `node` catalog entry pin node
+24.17.0, which `pnpm install` downloads into `node_modules`. That pinned node is
```
Sewer56: This will most likely get stale after a node pin. Also not something LLM needs to know.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+what scripts run under, not whatever is on `PATH` — check with
+`pnpm exec node --version`. Use `pnpm`. Never `npm install` or `yarn`, and don't
+bump those versions unless asked.
+
+## Tool use
+
+- Prefer harness file tools (read, edit, glob, grep) over shelling out to
+  `cat`, `sed` or `Get-Content`. Faster, and the edits are reviewable.
+- Don't write a throwaway script to do what an edit tool does.
```
Sewer56: Consequence of LLM forgetting instructions over long context.
Someone's in the >200k token territory; which in itself is a bit unoptimal.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+Node and pnpm are pinned exactly: `packageManager` pins pnpm 11.10.0 with an
+integrity hash, and `devEngines.runtime` plus the `node` catalog entry pin node
+24.17.0, which `pnpm install` downloads into `node_modules`. That pinned node is
+what scripts run under, not whatever is on `PATH` — check with
+`pnpm exec node --version`. Use `pnpm`. Never `npm install` or `yarn`, and don't
+bump those versions unless asked.
+
+## Tool use
+
+- Prefer harness file tools (read, edit, glob, grep) over shelling out to
```
Sewer56: [Part of Claude Code's system prompt already.](
https://github.com/Piebald-AI/claude-code-system-prompts/blob/738bccbb279db7024b9a41f921b473d31ddc421a/system-prompts/system-prompt-harness-instructions.md?plain=1#L20)

And most harnesses'

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+  `cat`, `sed` or `Get-Content`. Faster, and the edits are reviewable.
+- Don't write a throwaway script to do what an edit tool does.
+- The default shell is PowerShell on Windows. Anything committed to `scripts/`
+  should be cross-platform Node.
+- Never run the signed `pnpm package` locally; signing secrets are CI-only. See
+  [docs/packaging/windows.md](docs/packaging/windows.md).
+- When explaining code, cite `path:line` so the claim can be checked.
```
Sewer56: [Part of Claude Code's System Prompt already](<https://github.com/Piebald-AI/claude-code-system-prompts/blob/738bccbb279db7024b9a41f921b473d31ddc421a/system-prompts/system-prompt-harness-instructions.md?plain=1#L3>)

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+what scripts run under, not whatever is on `PATH` — check with
+`pnpm exec node --version`. Use `pnpm`. Never `npm install` or `yarn`, and don't
+bump those versions unless asked.
+
+## Tool use
+
+- Prefer harness file tools (read, edit, glob, grep) over shelling out to
+  `cat`, `sed` or `Get-Content`. Faster, and the edits are reviewable.
+- Don't write a throwaway script to do what an edit tool does.
+- The default shell is PowerShell on Windows. Anything committed to `scripts/`
```
Sewer56: On Windows, this will be in Shell tool's description already. Keep only second part, i.e. keep scripts cross plat.

AGENTS.md
```diff
@@ -1,14 +1,95 @@
...
+- `e2e-test` - scaffolding a Playwright E2E spec
+
+## Which doc to read
+
+Read the doc for the area you're touching rather than the whole tree.
+[docs/README.md](docs/README.md) is the full index; this table is the shortcut.
+
+| Working on                       | Read                                                                     |
+| -------------------------------- | ------------------------------------------------------------------------ |
+| Finding your way around the repo | [docs/repo-layout.md](docs/repo-layout.md)                               |
```
Sewer56: Save tokens by changing this table into flat list. Also don't need links.

AGENTS.md
insomnious: Where should it go, if it all?

AGENTS.md
insomnious: Confirmed against the live system prompt, it injects names and descriptions. Section cut.

AGENTS.md
insomnious: Cut, moved to CONTRIBUTING.md.

AGENTS.md
insomnious: Agreed. docs/README.md is the verbose one for humans, this stays a routing shortcut. No change.

AGENTS.md
insomnious: Right, the harness gives the server name but not the port. Keeping only that Vortex has to already be running on 127.0.0.1:9222, dropping the rest.

AGENTS.md
insomnious: Nowhere. It was already in docs/install-instructions/shared.md, which names `devEngines.runtime` and the catalog entry rather than hardcoding the versions. Cut.

AGENTS.md
insomnious: Going with `pnpm run verify` = format && build && test. One call, same saving, no per-OS script to keep in sync.

AGENTS.md
insomnious: Cut for that reason. shared.md names `devEngines.runtime` rather than the version number, so a bump does not touch the docs.

AGENTS.md
insomnious: Fair. Cut.

AGENTS.md
insomnious: Checked, it is in there. Cut.

AGENTS.md
insomnious: Same, confirmed present. Cut.

AGENTS.md
insomnious: Done, kept only the cross-platform Node half.

AGENTS.md
insomnious: Done, flat list of label plus plain path.

AGENTS.md
```diff
@@ -1,14 +1,56 @@
...
-- For collections or phased installation: read `AGENTS-COLLECTIONS.md`
-- For the auto-updater (update flow, channels, downgrades): read `docs/updater.md`
-- For testing updates locally (mock feed, test builds, env vars): read `docs/updater-testing.md`
+Project and coding documentation lives in [docs/](docs/README.md). This file
+covers only how to operate in this repo as an agent: verification, and which doc
+to load for a task. It deliberately duplicates nothing from `docs/`.
+
+## Verification
+
+`pnpm run verify` from the repo root runs format, build and test in one pass.
```
Sewer56: Here's a trick you can do.
Make the command print out the underlying commands performed.

That way you don't need to state 'it performs lint, typecheck', etc.
The LLM will know from context.

Another one of my sneaky tricks.

```
Build covers lint and typecheck, so there is no need to run those separately, and
everything fans out through Nx to the affected packages, so there is no need to
change directory into one.
```

Will allow you to trim all of this :p

docs/writing-documentation.md
```diff
@@ -0,0 +1,99 @@
...
+regardless of whether a human or an agent is the expected reader. The two
+audiences want the same things: what this is, how it works, where the code is.
+
+Do not put project or coding documentation in `AGENTS.md` or `CLAUDE.md`.
+`AGENTS.md` covers only how to operate as an agent in this repo: tool
+preferences, MCP servers, skills, and which doc to load for a task. If you find
+yourself writing "the renderer is structured as..." into `AGENTS.md`, it belongs
+in `docs/`.
+
+We used to keep this content in `AGENTS-DIRECTORIES.md`, `AGENTS-FRONTEND.md`
```
Sewer56: This can be yanked out. Better describe only the current behaviour.

AGENTS.md
Sewer56:
```
"verify": "echo pnpm run format && pnpm run format && echo pnpm run build && pnpm run build && echo pnpm run test && pnpm run test"
```

Or just state exact commands ran, rather than saying it in text/prose.
Otherwise we risk LLM repeating commands.

Sewer56 (approved):
Note that I didn't go through the frontend stuff too deeply, as I've not touched that part of the codebase.

All's good otherwise 👌

PR #24075, opened 2026-09-01
Keep persistence alive past a single fault

IDCs:
One fault in the diff pipeline cost the rest of the session, silently: a diff past the engine's argument-spread limit threw out of computeStateDiff, the throw escaped dispatch and froze previousState so every later diff grew from the same stale snapshot, and a send failure left the flush timer handle set so no flush was ever scheduled again.

Diff operations are now appended without spread, a diff-computation fault is caught and logged with previousState held so the changes ride the next diff, and a hive's queue is only cleared once its send succeeded.

fixes LAZ-1007

### DIFF
--- a/src/renderer/src/store/persistDiffMiddleware.ts
+++ b/src/renderer/src/store/persistDiffMiddleware.ts
@@ -164,9 +164,10 @@ export function createPersistDiffMiddleware(
   let flushTimeout: ReturnType<typeof setTimeout> | null = null;
 
   /**
-   * Flush all pending diffs to main process
+   * Flush all pending diffs to main process.
    */
   const flushDiffs = () => {
+    flushTimeout = null;
     const persistApi = getApi();
     if (persistApi == null) {
       return;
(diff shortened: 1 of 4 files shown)


src/renderer/src/store/persistDiffMiddleware.ts
```diff
@@ -176,14 +177,24 @@ export function createPersistDiffMiddleware(
...
         persistApi.sendDiff(hive, serializedOps);
+      } catch (err) {
+        log("error", "failed to send state diff for persistence", {
+          hive,
+          operationCount: serializedOps.length,
+          error: getErrorMessageOrDefault(err),
+        });
+        continue;
       }
+      delete pendingDiffs[hive];
```
erri120: Should this be in a `finally` block so it's always removed? Otherwise this would be better in the `try` block.

src/renderer/src/store/stateDiff.ts
```diff
@@ -16,6 +16,16 @@ function isObject(state: unknown): state is Record<string, unknown> {
...
+/**
+ * Append without spread: `push(...items)` passes every element as an argument and throws past
+ * the engine's argument limit (~125k); one subtree's diff can exceed that.
+ */
+function appendOperations(target: DiffOperation[], items: DiffOperation[]): void {
+  for (const item of items) {
+    target.push(item);
+  }
+}
+
```
erri120: Why do we need this?

IDCs: We don't really, it can be an anonymous function too. As long as we don't spread the items in a push.

src/renderer/src/store/persistDiffMiddleware.ts
IDCs: Good point.

IDCs: Placed in `try` block as we do not want to lose the queue.

src/renderer/src/store/stateDiff.ts
IDCs: Opted for a recursive accumulator instead - we already use the same shape for state flattening which works quite well.

erri120 (approved):

PR #24087, opened 2026-09-02
test: QA-98 user can log back in after signing out

RedRanger14:
Linear ticket: https://linear.app/nexus-mods/issue/QA-98/account-57-if-i-sign-out-then-i-can-trigger-the-login-process-again-by

Adds a logout-then-login test to the sign-out spec: after signing out, the header Log in button reopens the OAuth flow on the website and the user authenticates back in. Signing in stays setup (the nexusUser fixture), so each test owns its app instance and is order-independent.

Extracts the logout flow into helpers/account so QA-97 and QA-98 share one implementation, and swaps loginToNexus' inline openExternal stub for the existing installExternalOpenSpy so the opened URL is assertable. seededAuthStatePath moves to helpers/authState, shared by the snapshot fixture and the new test.


packages/e2e/src/tests/account-logout.spec.ts
```diff
@@ -25,8 +25,42 @@ test.describe("Account - Sign out", () => {
...
+  test("[QA-98] user can log back in after signing out", async ({
+    vortexApp,
+    vortexWindow,
+  }, testInfo) => {
+    const profileMenu = new ProfileMenu(vortexWindow);
+
+    await test.step("Sign out from the account menu", async () => {
+      await expect(profileMenu.avatarButton).toBeVisible({ timeout: Timeouts.NETWORK });
+      await profileMenu.avatarButton.click();
+      await profileMenu.logoutItem.dispatchEvent("click");
```
chnexusmods, line 49: Is it worth splitting these actions into separate steps? Are we ever likely to hit a race condition where it tries to click the logout button before it's available?

RedRanger14: More test steps, now thats feedback I can get behind

chnexusmods (approved):

PR #24089, opened 2026-09-02
Fix dot path game installation having spurious conflicts

In-line:
If the Skyrim Special Edition is installed in a path containing `.`, like `Z:\.local\share\Steam\...` Vortex was reporting spurious conflicts, because `isBlacklisted` wasn't functioning correctly with paths containing a dot.

This fix is relevant to Wine/Proton usecase and to the Windows users, who may install the game to path having `.` in it.


In-line:
@Aragas @erri120 @r1chm8 ping

IDCs (changes requested):
Hey @In-line - Vortex uses `minimatch` 3.1.5; the `nodot` option does not exist on that version. I think what you're looking for is `dot: true`

In-line:
@IDCs thanks

IDCs (approved):

PR #24091, opened 2026-09-02
Migrate IPC error serialization to use VortexError

erri120:
Closes LAZ-749.


src/preload/src/ipc.ts
```diff
@@ -0,0 +1,122 @@
...
+  channel: C,
+  ...args: SerializableArgs<Parameters<InvokeChannels[C]>>
+): Promise<AssertSerializable<Awaited<ReturnType<InvokeChannels[C]>>>> {
+  const reply: WireReply<AssertSerializable<Awaited<ReturnType<InvokeChannels[C]>>>> =
+    await ipcRenderer.invoke(channel, ...args);
+  if (reply.error) {
+    throw deserializeVortexError(reply.error, errorOriginTracker);
+  } else {
+    return reply.data;
+  }
```
IDCs, line 60: We're now expecting every reply from main to have `data` or `error` defined, but `getInitMetadata` will return the metadata object directly. Which means unwrapping gives `undefined` and the app will fail to start.

IDCs:
```
2026-09-03T05:36:06.347Z [ERRO] [RENDERER] error setting up renderer {"message":"Cannot read properties of undefined (reading 'commandLine')","name":"TypeError","stack":"TypeError: Cannot read properties of undefined (reading 'commandLine')\n    at applyAppMetadata (C:\\Projects\\wtf\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\renderer.tsx:504:16)\n    at init (C:\\Projects\\wtf\\src\\main\\build\\webpack:\\@vortex\\renderer\\src\\renderer.tsx:639:5)"}
```

IDCs (approved):

PR #24093, opened 2026-09-02
Add docs for game support, releasing and extensions

insomnious:
Three docs for onboarding: how a game comes to be supported, what happens to a release after it is packaged, and what an extension is. Game support and extensions had nothing at all. Releasing had six accurate docs and no map, so that one is mostly links plus the post-release automation, signing operations and announcing.

The announcing section is reconstructed from Slack rather than from the repo, so it needs a check from Vitalii and comms before anyone relies on it.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,6 +26,8 @@ Project and coding documentation. To set up and build Vortex, start with
 
 ## Subsystems
 
+- [extensions.md](extensions.md) - What an extension is, how core, bundled and user extensions load, the lifecycle and the registration surface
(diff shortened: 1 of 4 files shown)


iqraut:
That sounds like a useful way to structure the onboarding, especially if each doc answers one clear part of the release lifecycle. The announcing section is the one I’d be most cautious with since Slack discussions can capture context that never became an official process. Getting Vitalii and comms to validate it before treating it as documentation makes sense, and it might also be worth marking that section as provisional until that review is done.
For checking Stardew Valley items, locations, crops, and other gameplay details, https://stardewvalleyapk.app/ can be a useful reference too.

PR #24104, opened 2026-09-07
pin plugin persistor and load order reducer behaviour

IDCs:
Pre-move tranche of LAZ-1025: characterization tests pinning current plugin management behaviour
ahead of the gamebryo-plugin-management core move (LAZ-1026).

- PluginPersistor: missing plugins.txt, the original plugin format, blueprint exclusion, native
  load order offsets.
- loadOrder reducer: setPluginOrder clean slate, updatePluginOrder semantics; tests drive the
  real action creators.

part of LAZ-1025

### DIFF
--- a/extensions/gamebryo-plugin-management/test-utils/vortex-api.ts
+++ b/extensions/gamebryo-plugin-management/test-utils/vortex-api.ts
@@ -21,14 +21,19 @@ import { batchDispatch, makeOverlayableDictionary } from "../../../src/renderer/
  */
 
 // real node-fs behavior wrapped in Bluebird (production code relies on Bluebird's
-// predicate .catch), except readdirAsync which tests arrange per-case
+// predicate .catch and chains .filter on readdirAsync). readdirAsync is a vi.fn whose
+// default reads the real directory; suites arrange it per-case to override, and
+// mockReset restores this default. utimesAsync returns a native promise: its only
+// consumer uses a plain .catch.
 export const fs = {
-  readdirAsync: vi.fn(),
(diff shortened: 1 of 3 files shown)


extensions/gamebryo-plugin-management/test-utils/vortex-api.ts
```diff
@@ -21,14 +21,19 @@ import { batchDispatch, makeOverlayableDictionary } from "../../../src/renderer/
  */
 
 // real node-fs behavior wrapped in Bluebird (production code relies on Bluebird's
-// predicate .catch), except readdirAsync which tests arrange per-case
+// predicate .catch and chains .filter on readdirAsync). readdirAsync is a vi.fn whose
+// default reads the real directory; suites arrange it per-case to override, and
+// mockReset restores this default. utimesAsync returns a native promise: its only
+// consumer uses a plain .catch.
 export const fs = {
```
erri120: `export const fs: Partial<typeof import("../../../src/renderer/src/util/fs")>`

Ugly but will give you correct types and link this FS declaration with the real FS declaration. Remove the explicit types and rely on implicit types with this.

erri120 (changes requested):

extensions/gamebryo-plugin-management/test-utils/vortex-api.ts
IDCs: @erri120 - TBH these fixtures are not going to survive the move. `test-utils` is basically a stand-in because we couldn't reach the proper fixtures defined on renderer from within a bundled extension. Still worth adding it I guess.

erri120 (approved):

PR #24109, opened 2026-09-07
Error and OTeL Fixes

Aragas:
Fixed instanceof checks for VortexError
Added a telemetry self check E2E test
Fixed the creash report process not being launched in dev builds Fixed environmental error detection
Added read-only fs error detection
Report the real stack for the error instead of the synthetic


erri120 (changes requested):
- Change `instanceof`-fix to be hardcoded map on the deserialization function, mark it as deprecated
- revert the `getPackagePath` changes
- explicit test for `adoptCauseFrames` with real stackframes
- revert `src/renderer/src/extensions/analytics/mixpanel/error-code.ts`
- remove `packages/e2e/telemetry-selftest.ts`

src/main/src/getVortexPath.ts
```diff
@@ -84,7 +84,10 @@ function getLocalesPath(): string {
  */
 function getPackagePath(unpacked: boolean): string {
   if (isDevelopment) {
-    return basePath;
+    // basePath is <app>/build, which holds the bundle but no package.json, so it
+    // can't be handed to electron.exe as an app path. The unpacked variant has to
+    // stay on build, that's where LICENSE.md and the bundled binaries live.
+    return unpacked ? basePath : applicationPath;
```
Aragas, line 90: Needed like this. Otherwise on a dev build it runs `electron.exe %PATH%/src/main/build --report <file>` instead of `electron.exe %PATH%/src/main --report <file>`
We need to supply the root folder, not the build folder

src/shared/src/errors.ts
```diff
@@ -170,6 +178,43 @@ const INSTALL_PATH_RE = new RegExp(String.raw`(?:${WIN}|${UNIX})${SEGS}(?=${ANCH
...
+ * at the classifier rather than at the failure — the error it wraps carries the
+ * real throw site. Since frames are what {@link computeErrorFingerprint} hashes,
+ * reporting the classifier's would collapse every error it touches into one
+ * fingerprint. The header stays ours so the first line still names the error
+ * actually being reported.
+ *
+ * Errors are read, never mutated: each error's own `stack` keeps describing
+ * where that error was constructed, which is the truth.
+ */
+export const resolveReportedStack = (error: Error): string | undefined => {
```
erri120: Missing tests for this method.

src/shared/src/errors.ts
```diff
@@ -170,6 +178,43 @@ const INSTALL_PATH_RE = new RegExp(String.raw`(?:${WIN}|${UNIX})${SEGS}(?=${ANCH
...
+  let frames = framesOf(error.stack);
+  let cause: unknown = error.cause;
+  for (let depth = 0; depth < MAX_CAUSE_DEPTH && cause instanceof Error; depth += 1) {
+    frames = framesOf(cause.stack) ?? frames;
+    cause = cause.cause;
+  }
+
+  return frames === undefined ? error.stack : `${error.name}: ${error.message}\n${frames}`;
+};
+
```
Aragas: Here's an example of what it fixes

* Renderer calls app:extractFileIcon; writeFile into System32 throws. Raw error:
```
Error: EPERM: operation not permitted, open 'C:\Windows\System32\icon.png'
    at open (node:internal/fs/promises:640:25)
    at writeFile (node:internal/fs/promises:1260:14)
    at src/main/src/ipcHandlers.ts:237:7
```
* Renderer calls a different handler; an fs.stat on a missing mod folder throws. Raw error:
```
Error: ENOENT: no such file or directory, stat 'D:\Games\Skyrim\Data'
    at stat (node:internal/fs/promises:1032:18)
    at src/main/src/filesystem/filesystem-impl.ts:212:5
```

Without the fix, both errors will have a stack like this
```
VortexError: …
    at Object.parseError (src/shared/src/errors/parser.ts:39:10)
    at toWireError (src/shared/src/errors/serialization.ts:141:56)
    at src/main/src/ipc.ts:173:18
    at Session.<anonymous> (node:electron/js2c/browser_init:2:114924)
```

With the fix, here's what the stack will look like
```
VortexError: EPERM: operation not permitted, open 'C:\Windows\System32\icon.png'
    at open (node:internal/fs/promises:640:25)
    at writeFile (node:internal/fs/promises:1260:14)
    at src/main/src/ipcHandlers.ts:237:7
```
```
VortexError: File or directory does not exist at 'D:\Games\Skyrim\Data'
    at stat (node:internal/fs/promises:1032:18)
    at src/main/src/filesystem/filesystem-impl.ts:212:5
```

erri120: Right, put it in a test.

src/shared/src/errors.ts
Aragas: Done

src/shared/src/errors.test.ts
```diff
@@ -468,3 +474,103 @@ describe("isEnvironmentalError", () => {
...
+    expect(cause.stack).toBe(causeBefore);
+  });
+
+  it("walks to the deepest cause that has frames", () => {
+    const root = withFrames(new Error("root"), "at root (src/root.ts:1:1)");
+    const middle = withFrames(new Error("middle", { cause: root }), "at middle (src/mid.ts:1:1)");
+    const err = withFrames(new Error("top", { cause: middle }), OWN_FRAME);
+
+    expect(resolveReportedStack(err)).toBe("Error: top\n    at root (src/root.ts:1:1)");
+  });
```
erri120: So instead of

```
Error 1: src/top.ts
Error 2: src/mid.ts
Error 3: src/bottom.ts
```

we get

```
Error 1: src/bottom.ts
```

I'd prefer if we didn't remove information. Maybe something recursive where the stack for error `n` is a combination of it's own stack + error `n - 1`. Then you get full visibility into the stack throughout all errors for better debugging.

Aragas: The only sane way would be to do something like this, any other suggestions?

```
VortexError: EPERM: operation not permitted, open 'C:\Windows\System32\icon.png'
    at Object.parseError (src/shared/src/errors/parser.ts:39:10)
    at toWireError (src/shared/src/errors/serialization.ts:141:56)
    at src/main/src/ipc.ts:173:18
    at Session.<anonymous> (node:electron/js2c/browser_init:2:114924)
Caused by: Error: EPERM: operation not permitted, open 'C:\Windows\System32\icon.png'
    at open (node:internal/fs/promises:640:25)
    at writeFile (node:internal/fs/promises:1260:14)
    at src/main/src/ipcHandlers.ts:237:7
```

```
VortexError: File or directory does not exist at 'D:\Games\Skyrim\Data'
    at Object.parseError (src/shared/src/errors/parser.ts:39:10)
    at toWireError (src/shared/src/errors/serialization.ts:141:56)
    at src/main/src/ipc.ts:173:18
    at Session.<anonymous> (node:electron/js2c/browser_init:2:114924)
Caused by: Error: ENOENT: no such file or directory, stat 'D:\Games\Skyrim\Data'
    at stat (node:internal/fs/promises:1032:18)
```

Aragas: Done. The codebase looks simpler now

erri120 (approved):

PR #24122, opened 2026-09-08
Simplify classifyErrorCode

erri120:
Closes LAZ-748


IDCs:
@insomnious - do you have any funnels set up for the error kind strings on mixpanel?

PR is good - but it could break some stuff. `user_canceled` -> `user-canceled`, `data_invalid` -> `data-invalid` stuff like that.

IDCs (approved):

erri120:
> @insomnious - do you have any funnels set up for the error kind strings on mixpanel?
> 
> PR is good - but it could break some stuff. `user_canceled` -> `user-canceled`, `data_invalid` -> `data-invalid` stuff like that.

I've checked mixpanel and only found breakdowns so far, no explicit filters.

PR #24128, opened 2026-09-08
Fixed Error Classification

Aragas:
HTTPError now survives IPC transition
Fixed instanceof HTTPError check
Nexus-API HTTPError is now mapped into our HTTPError RequestError is now http:generic

Closes fingerprints 158ae649, 4b4edc1d, f13a338b, 3fbb1069, d9b612cc, 4bf4c70a
Closes LAZ-1083

### DIFF
--- a/src/main/src/downloading/errors.ts
+++ b/src/main/src/downloading/errors.ts
@@ -46,12 +46,12 @@ export function toNetworkError(endpoint: URL | ResolvedEndpoint, err: unknown): 
     );
     if (parsed.data.kind !== "unknown") return parsed;
 
-    // got's own codes (ERR_READING_RESPONSE_STREAM, ...) have no POSIX mapping: the connection
-    // dropped mid-request, which is worth retrying
+    // got's own codes (ERR_READING_RESPONSE_STREAM, ...) have no POSIX mapping; the retry
+    // strategy decides by `originalCode`
     return new VortexError(
-      "Network request failed (transient)",
+      "Network request failed",
       { kind: "http:generic", url: urlString, originalCode: err.code },
-      { cause: err, isTransient: true },
+      { cause: err },
     );
   }
 
--- a/src/main/src/downloading/errors.ts
+++ b/src/main/src/downloading/errors.ts
@@ -39,10 +39,19 @@ export function toNetworkError(endpoint: URL | ResolvedEndpoint, err: unknown): 
   }
 
   if (err instanceof RequestError) {
-    return parseError(err, { url: urlString }, ({ data, isTransient }) =>
+    const parsed = parseError(err, { url: urlString }, ({ data, isTransient }) =>
       data.kind === "http:generic" || data.kind === "os:generic"
         ? `Network request failed${isTransient ? " (transient)" : ""}`
         : undefined,
+    );
+    if (parsed.data.kind !== "unknown") return parsed;
+
+    // got's own codes (ERR_READING_RESPONSE_STREAM, ...) have no POSIX mapping: the connection
+    // dropped mid-request, which is worth retrying
+    return new VortexError(
+      "Network request failed (transient)",
+      { kind: "http:generic", url: urlString, originalCode: err.code },
+      { cause: err, isTransient: true },
     );
   }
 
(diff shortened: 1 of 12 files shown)


src/main/src/downloading/errors.ts
```diff
@@ -39,11 +39,20 @@ export function toNetworkError(endpoint: URL | ResolvedEndpoint, err: unknown):
...
     );
+    if (parsed.data.kind !== "unknown") return parsed;
+
+    // got's own codes (ERR_READING_RESPONSE_STREAM, ...) have no POSIX mapping: the connection
+    // dropped mid-request, which is worth retrying
+    return new VortexError(
+      "Network request failed (transient)",
+      { kind: "http:generic", url: urlString, originalCode: err.code },
+      { cause: err, isTransient: true },
+    );
```
erri120: How do you know this is transient?

src/shared/src/errors/parser.ts
```diff
@@ -56,7 +56,11 @@ function parseNodeSystemError(
 ): { message: string; data: VortexError["data"]; isTransient?: boolean } | undefined {
   const data = parseNodeSystemErrorData(cause);
   if (!data) {
-    return undefined;
+    // "socket hang up" and TLS disconnects carry a POSIX code but no errno/syscall
+    const code = "code" in cause ? cause.code : undefined;
+    return typeof code === "string" && NETWORK_POSIX_CODES.has(code)
+      ? networkError(code, { originalCode: code }, cause.message, context)
+      : undefined;
   }
```
erri120: You're saying that ""socket hang up" and TLS disconnects carry a POSIX code" yet you don't specify what those are. The comment doesn't match the actual logic here, the branch is running against ALL codes we have in the `NETWORK_POSIX_CODES` set.

I'm assuming you're getting this:

https://github.com/nodejs/node/blob/51c992f99bc0ec6dc1cafdc9f00defa28eead3ea/lib/internal/tls/wrap.js#L1361-L1366

which does indeed not set `errno` or `syscall`: https://github.com/nodejs/node/blob/51c992f99bc0ec6dc1cafdc9f00defa28eead3ea/lib/internal/errors.js#L847-L856

as verified by Node.js tests: https://github.com/nodejs/node/blob/51c992f99bc0ec6dc1cafdc9f00defa28eead3ea/test/parallel/test-tls-wrap-econnreset.js#L19

So instead of checking all network POSIX codes, this branch should just be for `ECONNRESET` with a comment linking to the code like I did.

src/renderer/src/extensions/nexus_integration/nxmProtocol.ts
```diff
@@ -547,7 +552,12 @@ export class NxmProtocol {
     if (error instanceof HTTPError && error.statusCode === 401) {
       throw new ProcessCanceled("You are not logged in to Nexus Mods!");
     }
-    throw error;
+    // The error crosses IPC to the downloader, which only understands VortexError kinds; a raw
+    // socket failure from the api client would otherwise arrive as "Unknown error thrown".
+    const parsed = parseError(error, undefined, ({ data }) =>
+      data.kind === "os:generic" ? "Network request failed" : undefined,
+    );
+    throw parsed.data.kind === "unknown" ? error : parsed;
```
erri120: Walk me through the stack, how do we end up here in this branch?

erri120 (changes requested):

src/renderer/src/extensions/nexus_integration/nxmProtocol.ts
Aragas: This was a false-positive detection. Renderer doesn't need any handling, nexus-api already handles it

src/shared/src/errors/parser.ts
Aragas: Thanks, corrected. Should have looked deeper into it

src/main/src/downloading/errors.ts
Aragas: You don't. Initially I was assuming that we could just classify them in like this generally, but no, that's not a good idea

Aragas: Also, what's the point of this flag? It's not used atm anywhere

erri120: `isTransient` is mostly useful for file operations but since we aren't checking for FS errors yet it's unused right now.

erri120 (approved):

PR #24137, opened 2026-09-09
Upgrade Electron to 44.2.0

erri120:
Closes LAZ-729.

As far as breaking changes go, the previously deprecated `clipboard` API was now changed to align with the W3C API.

Otherwise should be a simple drop-in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,17 +6,22 @@
   "type": "commonjs",
   "packageManager": "pnpm@11.10.0+sha512.0b7f8b98060031904c017e3a41eb187a16d40eeb829b95c4f8cb03681761fc4ab53dd219115b9b447f4dce1a05a214764461e7d3703392a9f32f9511ce8c86c8",
   "engines": {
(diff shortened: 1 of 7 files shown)


Aragas:
So we now depend on both a specific node version and a specific `node-abi` version?

erri120:
> So we now depend on both a specific node version and a specific `node-abi` version?

`node-abi` is a detection library used by `@electron/rebuild` to figure out the ABI version of electron. It's pretty whack but it means we need a recent version of `node-abi` that knows about our version of electron.

Aragas:
> `node-abi` is a detection library used by `@electron/rebuild` to figure out the ABI version of electron. It's pretty whack but it means we need a recent version of `node-abi` that knows about our version of electron.

Makes sense. So we'll need to keep it up to date if we want to use the latest electron version. Might be worth a comment, but sounds good

Aragas (approved):

PR #24140, opened 2026-09-09
Fixed a signed-in session with no account and no login button

r1chm8:
Reports of users who are logged out and offline seeing no login button. They are not logged out — the app has thrown their account details away and kept their credentials.

`setOAuthCredentials` stores the credentials it is handed and then looks the avatar up over the network, so most of the ways that call can fail say nothing about the session, an offline start among them. `updateToken` read any failure as a rejected login and cleared `userInfo`, leaving the credentials in state. Those still satisfy `isLoggedIn`, so the header had nothing to show in either slot: no account menu, because there were no details, and no "Log in" button, because as far as the store was concerned the user was already signed in — while the toast told them to log in again.

Only the site refusing the credentials clears the account now: 401, which nexus-api rethrows once its own token refresh hasn't helped and which `errorFromNexusError` already reads as a refused login, and 403. Every other failure — offline, a timeout, a 500 — keeps the last known account. A session with nothing persisted, a first run offline or one an older build already wiped, is seeded from the claims the access token carries, which needs no request; the avatar and email only exist server-side, so the header falls back to its generic icon until the next successful refresh.

The profile menu also opens on a generic account when details are missing, so logging out — the way back to a login button — stays reachable however a session ends up in that state.

Reading a status code off an `unknown` error is mechanical, and the shared error helpers already do it for `code` and `systemCode`, so `getErrorStatusCode` joins them in `src/shared/src/errors.ts`; what stays in `nexus_integration` is the part only it can judge, that 401 and 403 end the session. The other 21 bare `statusCode` reads in that extension are left for a follow-up.

One thing left alone deliberately: a genuinely rejected session still keeps its dead credentials, so *"please log in again"* does not produce a login button by itself — the user has to log out first. Clearing credentials on a real rejection is the fix for that, and is a separate change.

Verified with `pnpm run verify`; the E2E suite was not run.

### DIFF
--- a/src/renderer/src/extensions/nexus_integration/util.ts
+++ b/src/renderer/src/extensions/nexus_integration/util.ts
@@ -1955,6 +1955,52 @@ function onJWTTokenRefresh(api: IExtensionApi, credentials: IOAuthCredentials, n
   //Promise.resolve(getUserInfo(api, nexus));
 }
 
+/**
+ * Codes for "we never reached the site", as opposed to the site telling us the credentials
+ * are no good. While the network is down the Disableable proxy stands in for every request
+ * and rejects with ProcessCanceled, and the calls that bypass it fail with a socket error.
+ */
+const CONNECTION_ERROR_CODES = [
+  "EAI_AGAIN",
+  "ECONNABORTED",
+  "ECONNREFUSED",
+  "ECONNRESET",
+  "EHOSTUNREACH",
+  "ENETDOWN",
+  "ENETUNREACH",
+  "ENOTFOUND",
+  "EPIPE",
+  "ESOCKETTIMEDOUT",
+  "ETIMEDOUT",
+];
+
+const isConnectionError = (err: Error): boolean =>
+  err instanceof ProcessCanceled ||
+  err instanceof TimeoutError ||
+  CONNECTION_ERROR_CODES.includes(getErrorCode(err) ?? "") ||
+  getErrorMessage(err).includes("getaddrinfo");
+
+/**
+ * The account the access token describes on its own. All of this is signed into the token, so
+ * reading it needs no request, which makes it the fallback for a session we hold credentials
+ * for but can't reach the site to flesh out. The avatar and email only exist server-side and
+ * come out empty; the header falls back to the generic account icon for an empty avatar.
+ */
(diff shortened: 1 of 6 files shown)


src/renderer/src/extensions/nexus_integration/util.ts
```diff
@@ -1981,6 +2027,27 @@ export function updateToken(
...
+          // already wiped - so fall back to what the token itself says
+          const fromToken = userInfoFromToken(credentials.token);
+          if (fromToken !== undefined) {
+            api.store.dispatch(setUserInfo(fromToken));
+          }
+        }
+        api.events.emit("did-login", err);
+        return false;
+      }
+
```
erri120, line 2025: It shouldn't matter whether the error was a connection error or not. In fact there is only one error that we should have an explicit branch for: 403 forbidden or 401 unauthorized. This is the only condition we should check for because that's what the code below is actually assuming:

> Authentication failed, please log in again

erri120 (changes requested):

src/renderer/src/extensions/nexus_integration/util.ts
r1chm8: I've included 401 too as I believe that's what nexus-api surfaces

src/renderer/src/extensions/nexus_integration/util.ts
```diff
@@ -1981,10 +2011,31 @@ export function updateToken(
     .then(() => getUserInfo(api, nexus)) // update userinfo as we've set some new nexus credentials, either by launch, login or token refresh
     .then(() => true)
     .catch((err) => {
-      api.showErrorNotification("Authentication failed, please log in again", err, {
-        allowReport: false,
+      if (isLoginRefused(err)) {
```
erri120, line 2017: `err` is `unknown` and shouldn't get implicitly cast. `isLoginRefused` should do proper type checks on `unknown`.

erri120 (changes requested):

src/renderer/src/extensions/nexus_integration/util.ts
r1chm8: good spot, added a new error helper to narrow this type down.

erri120 (approved):

PR #24148, opened 2026-09-09
CI Deadlock fix

Aragas:
Cherry-pick of #24147 into `master`.

---

Fixed a deadlock where we require the main.yml workflow to be run and be successful and paths-ignore actually skipping the workflow execution

We now check whether the workflow should be skipped inside the workflow execution.

See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks
> Avoid requiring workflows that can be skipped.

Since we can't skip it, lets just mark it as successful 🤷‍♂️

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,40 +3,8 @@ name: Main
 on:
   push:
     branches: [master, release/*]
-    paths-ignore:
-      - "**.md"
-      - "docs/**"
-      - "locales/**"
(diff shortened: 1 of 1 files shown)


Aragas:
Cherry pick failed because of the lack of workflow permissions, to be fixed

PR #24150, opened 2026-09-09
Compute a fingerprint for a native crash

Aragas:
We didn't provide any fingerprint to a native crash, which is definitely not correct.
Now we'll be computing the fingerprint based on the app version (always), title, process and the offset and anything else that is available

### DIFF
--- a/src/main/src/errorReporting.ts
+++ b/src/main/src/errorReporting.ts
@@ -11,7 +11,7 @@ import {
   getErrorMessageOrDefault,
   sanitizeFramePath,
 } from "@vortex/shared";
-import type { ReportableError } from "@vortex/shared/errors";
+import { type CrashType, type ReportableError, CrashTypeTitle } from "@vortex/shared/errors";
 import { recordErrorOnSpan, SanitizingSpanExporter } from "@vortex/shared/telemetry";
 import { app } from "electron";
 
@@ -57,7 +57,7 @@ export function errorToReportableError(error: Error): ReportableError {
 }
 
 interface ICrashInfo {
-  type: string;
+  type: CrashType;
   error: ReportableError;
   context?: Record<string, string>;
   reportProcess?: string;
@@ -321,14 +321,6 @@ async function collectCrashDumps(): Promise<IDumpFile[]> {
   return found;
 }
 
-const CRASH_TITLES: Record<string, string> = {
-  Crash: "Unrecoverable error",
-  EarlyCrash: "Crash during startup",
-  PreviousSessionCrash: "Native crash",
-  ChildProcessGone: "Child process crashed",
-  RenderProcessGone: "Renderer process crashed",
-};
-
 /** What identifies a crash site when there is no JavaScript stack to hash. */
 const CRASH_IDENTITY_ATTRIBUTES = [
   "crash.sourceProcess",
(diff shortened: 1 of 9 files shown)


src/main/src/errorReporting.ts
```diff
@@ -312,6 +321,41 @@ async function collectCrashDumps(): Promise<IDumpFile[]> {
   return found;
 }
 
+const CRASH_TITLES: Record<string, string> = {
+  Crash: "Unrecoverable error",
+  EarlyCrash: "Crash during startup",
+  PreviousSessionCrash: "Native crash",
+  ChildProcessGone: "Child process crashed",
+  RenderProcessGone: "Renderer process crashed",
+};
```
erri120: Why do we need this?

```typescript
"error.title": error.title ?? CRASH_TITLES[type] ?? type
```

it just seems unnecessary when we can fallback to `type` always.

Aragas: To have a human-readable title in our telemetry view.

erri120: 1) `RenderProcessGone` is pretty human-readable
2) It's falling back to `type` for everything else anyways

I'd rather have the original in telemetry and not have to do a translation from our custom titles to the original type for debugging.

Aragas: RenderProcessGone and the other types are our own custom codes we use for identification of errors, those are not externally sourced
I do see the point here and will see how better it could be done

Aragas: Is this approach acceptable or would you still remove the human-readable part? Is your view with errors and their classifications to have only codes without a human-readable mapping?

src/shared/src/types/errors.ts
```diff
@@ -1,5 +1,25 @@
...
+  "Installer failed": "Installer failed",
+  "Unknown error": "Unknown error",
+} as const;
+
+/**
+ * Catalog of every crash kind Vortex knows how to produce and classify.
+ * @public
+ */
+export type CrashType = keyof typeof CrashTypeTitle;
+
```
Aragas, line 22: Wasn't sure whether that belongs to base.ts or to a separate crash.ts

src/main/src/errorReporting.ts
erri120: I simply disagree that error analysis and classification should be on the client. How we interpret codes, types and whatnot is for the analysis step on our telemetry service.

Aragas: For that we would need an infrastructure that doesn't exists yet. You have a good point with this, but I don't see a reason to not have more information right now and then move this logic to infra, once it's ready. I don't believe creating such infrastructure would be easy, plus we still need to figure out _what_ we would want the client to send us. We will not know unless we have created a working flow

src/main/src/errorReporting.ts
```diff
@@ -395,20 +431,21 @@ export async function reportCrash(
...
+      ...attributes,
+    };
+    const span = tracer.startSpan("crash.report", { attributes: spanAttributes });
 
     const errorObj = new Error(error.message);
     errorObj.stack = error.stack;
     recordErrorOnSpan(span, errorObj, app.getVersion(), context, {
-      "error.title": error.title ?? "",
-      ...attributes,
+      "error.title": error.title ?? CrashTypeTitle[type],
```
erri120:
```suggestion
      "error.title": error.title ?? `Native crash: ${type}`,
```

Aragas: Done

erri120 (approved):

PR #24153, opened 2026-09-09
More crash loops to fix

Aragas:
Continuation of https://github.com/Nexus-Mods/Vortex/pull/24151
More infinite crash loops to fix

| Category | Exit code | Spans | Sessions | Median per session | Max in one session |
|---|---|---|---|---|---|
| Utility process crashed | -1073741205 (0xC000026B, DLL init failed at logoff) | 429,097 | 5,816 | 13 | 6,244 |
| GPU process crashed | 34 | 15,334 | 8,827 | 1 | 27 |
| Utility process crashed | -2147483645 | 4,341 | 997 | 3 | 67 |
| Utility process crashed | -1 | 2,357 | 545 | 1 | 634 |
| Utility process crashed | -1073741819 | 1,064 | 401 | 2 | 45 |

### DIFF
--- a/src/main/src/errorReporting.ts
+++ b/src/main/src/errorReporting.ts
@@ -312,15 +312,6 @@ async function collectCrashDumps(): Promise<IDumpFile[]> {
   return found;
 }
 
-// STATUS_DLL_INIT_FAILED_LOGOFF: a helper relaunched during logoff fails to start with this.
-// Fallback for when no window is alive to receive the session-end event
-const WINDOWS_SESSION_END_EXIT_CODE = -1073741205;
-
-/** Whether a process that died with `exitCode` faulted, rather than never starting at logoff. */
-export function isReportableExit(exitCode: number): boolean {
-  return exitCode !== WINDOWS_SESSION_END_EXIT_CODE;
-}
-
 /**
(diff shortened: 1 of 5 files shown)


src/main/src/errorReporting.ts
```diff
@@ -312,6 +312,15 @@ async function collectCrashDumps(): Promise<IDumpFile[]> {
...
 }
 
+// STATUS_DLL_INIT_FAILED_LOGOFF: Windows kills every process of a session with this at logoff,
+// and Chromium keeps relaunching its helpers into it until the main process is gone
+const WINDOWS_SESSION_END_EXIT_CODE = -1073741205;
+
+/** Whether a process that died with `exitCode` faulted, rather than being torn down by the OS. */
+export function isReportableExit(exitCode: number): boolean {
+  return exitCode !== WINDOWS_SESSION_END_EXIT_CODE;
+}
```
erri120: Do we really need this? It seems like something that should be fixed upstream: preventing Electron from restarting the utility process because the session is ending.

Aragas: Good point. I'm using a session end listener instead for the renderer process, that should be a better alternative

erri120 (approved):

PR #24164, opened 2026-09-10
fix: select a game's default primary tool when discovered later

### LINKED ISSUE #19543: [BUG ALPHA] Collections - Installing the New Vegas Script Extender did not automatically install as the primary tool
**Your Nexus Mods username**
VeryLastKiss

**Describe the bug**
When downloading a collection (I downloaded [NNV Non-Premium](https://www.nexusmods.com/games/newvegas/collections/vuadnp)) which contains the New Vegas Script Extender, then the script extender did not automatically set itself as the primary tool in Vortex. This means if a user were to click the "Play" button at the top left, then it will launch the vanilla game launcher, which will in turn also undo a lot of the collection's pre-configured base game .ini tweaks, which can cause various gameplay issues including crashes.

The New Vegas Script Extender is included into the NNV Non-Premium collection by default without any special attributes being assigned to it (it is not set for replicate, it is not optional, it is not binary patched, or etc.).

As a special note, I have never seen this bug occur on the 1.15.2 stable version of Vortex.

**To Reproduce**
1. Be on version 1.16.0-BETA.5 of Vortex.
2. Download the [NNV Non-Premium](https://www.nexusmods.com/games/newvegas/collections/vuadnp) collection. Alternatively, you could probably download Essential Foundation, which also contains the New Vegas Script Extender.
3. You do not need to download the optional mods in the collection, although if you do, there is still another bug which occurs where optional mods do not seem to apply mod overwrites (**NVR and ENB AO Hair Transparency Fix** is set to load before **New Vegas Mesh Improvement Mod (NVMIM)**, but the overwrite is not applied for the end-user).
4. Go to the **"Dashboard"** tab. The New Vegas Script Extender (xNVSE) will not be set as the primary tool.

**Expected behavior**
The New Vegas Script Extender should automatically be set as the primary tool in Vortex when the mod is downloaded through a collection (or just in general).

**Screenshots**
I did not wind up taking a screenshot for this; I apologize.

 - OS: **Windows 11**
 - Vortex Version: **1.16.0-BETA.5**

[network.log](https://github.com/user-attachments/files/25055555/network.log)
[vortex.log](https://github.com/user-attachments/files/25055553/vortex.log)
[vortex1.log](https://github.com/user-attachments/files/25055554/vortex1.log)
[vortex2.log](https://github.com/user-attachments/files/25055551/vortex2.log)
[vortex3.log](https://github.com/user-attachments/files/25055550/vortex3.log)
[vortex4.log](https://github.com/user-attachments/files/25055552/vortex4.log)

doodlum:
## Problem

Installing a collection that ships a script extender leaves the script extender unselected as the primary tool, so **Play** launches the vanilla executable. For collections that also apply base-game `.ini` tweaks, the vanilla launcher then reverts them.

Reported in #19543 (Windows 11).

## Cause

`defaultPrimary` is only ever consulted in the game-activation path in `setupGameMode`:

https://github.com/Nexus-Mods/Vortex/blob/c348fd013ad191261583968369add76b906865ab/src/renderer/src/extensions/gamemode_management/GameModeManager.ts#L157-L170

That block reads `discovery.tools` once, at activation time. `onDiscoveredTool` — which also fires for tools discovered after deployment and while the game is already active — registers the tool but never promotes it.

So a script extender arriving after activation (the collection-install case) is listed under Tools with nothing selecting it. `defaultPrimary` does reach the callback: `testApplicationDirValid` spreads the whole tool object into it (`{ ...application, path, hidden, custom }`).

## Change

Select a declared default in `onDiscoveredTool`, as soon as it becomes available.

Existing guards are preserved:

- only for the active profile's game
- only when `settings.interface.primaryTool[gameId]` is unset, so an explicit user choice is never overridden
- customised tools are still not overwritten

## How to reproduce manually

Fallout New Vegas is used here because its `nvse` tool declares `defaultPrimary: true`; the Skyrim/Fallout script extenders and SMAPI behave the same way.

1. Start from a Fallout New Vegas install with **no script extender present** — no `nvse_loader.exe` in the game folder. Manage the game.
2. Go to **Tools**. The **Default launcher** panel shows its placeholder, "Runs instead of the game when you hit Play." Nothing is selected, which is correct at this point: the tool does not exist yet.
3. Now make the script extender appear, which is what a collection install does when it delivers xNVSE: install xNVSE, or just drop `nvse_loader.exe` into the game folder.
4. Trigger a deployment — **Mods → ⋯ → Deploy Mods**. This fires `did-deploy`, which re-runs tool discovery on the already-active game, so xNVSE is discovered here.
5. Go back to **Tools**.

**Actual:** the Default launcher panel still shows the placeholder. xNVSE is listed under Tools but is not the default launcher, so **Play** starts the vanilla executable.

**Expected:** xNVSE is the Default launcher.

The original report reaches the same state via a collection install rather than steps 3-4; the mechanism is the same — the extender is discovered after the game was already activated.

To confirm the tool really was discovered (rather than simply not found), the log shows `discovering relative tools` followed by `potential match ...\nvse_loader.exe` after the deployment in step 4.

Fixes #19543

### DIFF
--- a/src/renderer/src/extensions/gamemode_management/GameModeManager.ts
+++ b/src/renderer/src/extensions/gamemode_management/GameModeManager.ts
@@ -503,9 +503,8 @@ class GameModeManager {
   }
 
   private onDiscoveredTool = (gameId: string, result: IDiscoveredTool) => {
-    const state = this.mStore.getState();
     const existing = getSafe(
-      state,
+      this.mStore.getState(),
       ["settings", "gameMode", "discovered", gameId, "tools", result.id],
       undefined,
     );
@@ -519,12 +518,9 @@ class GameModeManager {
     // appeared later was shown in Tools but Quick Launch kept starting the vanilla executable.
     // Select any game's declared default as soon as it becomes available, while preserving every
     // explicit user choice.
-    const active = activeProfile(this.mStore.getState());
-    const primary = getSafe(
-      this.mStore.getState(),
-      ["settings", "interface", "primaryTool", gameId],
-      undefined,
-    );
+    const state = this.mStore.getState();
+    const active = activeProfile(state);
+    const primary = state.settings.interface.primaryTool?.[gameId];
     if (active?.gameId === gameId && primary === undefined && result.defaultPrimary === true) {
       this.mStore.dispatch(setPrimaryTool(gameId, result.id));
     }
--- a/src/renderer/src/extensions/gamemode_management/GameModeManager.ts
+++ b/src/renderer/src/extensions/gamemode_management/GameModeManager.ts
@@ -503,8 +503,9 @@ class GameModeManager {
   }
 
   private onDiscoveredTool = (gameId: string, result: IDiscoveredTool) => {
+    const state = this.mStore.getState();
     const existing = getSafe(
-      this.mStore.getState(),
+      state,
       ["settings", "gameMode", "discovered", gameId, "tools", result.id],
       undefined,
     );
@@ -512,6 +513,20 @@ class GameModeManager {
     if (existing === undefined || !existing.custom) {
       delete result.executable;
       this.mStore.dispatch(addDiscoveredTool(gameId, result.id, result, false));
+    }
+    // Tool discovery also runs after deployment and on an already-active game. Previously the
+    // default was selected only during the narrow game-activation path, so a script extender that
+    // appeared later was shown in Tools but Quick Launch kept starting the vanilla executable.
+    // Select any game's declared default as soon as it becomes available, while preserving every
+    // explicit user choice.
+    const active = activeProfile(this.mStore.getState());
+    const primary = getSafe(
+      this.mStore.getState(),
+      ["settings", "interface", "primaryTool", gameId],
+      undefined,
+    );
+    if (active?.gameId === gameId && primary === undefined && result.defaultPrimary === true) {
+      this.mStore.dispatch(setPrimaryTool(gameId, result.id));
     }
   };
 


src/renderer/src/extensions/gamemode_management/GameModeManager.ts
```diff
@@ -513,6 +514,20 @@ class GameModeManager {
...
+    // explicit user choice.
+    const active = activeProfile(this.mStore.getState());
+    const primary = getSafe(
+      this.mStore.getState(),
+      ["settings", "interface", "primaryTool", gameId],
+      undefined,
+    );
+    if (active?.gameId === gameId && primary === undefined && result.defaultPrimary === true) {
+      this.mStore.dispatch(setPrimaryTool(gameId, result.id));
+    }
```
erri120, line 526: `getSafe` is deprecated, new code shouldn't use it.

erri120 (changes requested):

src/renderer/src/extensions/gamemode_management/GameModeManager.ts
doodlum: Replaced with optional chaining: `state.settings.interface.primaryTool?.[gameId]` (`primaryTool` is typed as an optional record, so this is equivalent). I also reverted my incidental edit to the adjacent pre-existing `getSafe` call so the diff is now purely additive and that deprecated call is left exactly as it is on master.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

IDCs:
I'd like to see a regression test for this fix. `makeApiHarness` in `test-utils/builders` will give you a store with real state shape which you can use.

IDCs (dismissed):

src/renderer/src/extensions/gamemode_management/GameModeManager.ts
```diff
@@ -594,6 +594,17 @@ class GameModeManager {
...
     }
+    // Tool discovery also runs after deployment and on an already-active game. Previously the
+    // default was selected only during the narrow game-activation path, so a script extender that
+    // appeared later was shown in Tools but Quick Launch kept starting the vanilla executable.
+    // Select any game's declared default as soon as it becomes available, while preserving every
+    // explicit user choice.
+    const state = this.mStore.getState();
+    const active = activeProfile(state);
+    const primary = state.settings.interface.primaryTool?.[gameId];
+    if (active?.gameId === gameId && primary === undefined && result.defaultPrimary === true) {
```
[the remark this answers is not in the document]
doodlum: Half right, so I addressed the coverage gap but not the guard.

You are correct that `setPrimaryTool` still fires when the existing record has `custom: true`, and that no test covered it — that gap is real and is now closed by "still selects a customised tool as the default launcher" (42b342de8).

I did not add `custom` to the condition, because `custom` protects the tool *record* from being overwritten by discovery rather than expressing which tool should launch, and `setGameMode` already promotes any `discovered.tools` entry with `defaultPrimary` without consulting it (GameModeManager.ts L160-173). Gating only this path would make the two halves of one feature disagree: a game whose extender the user customised would get no default launcher here, but would get one on activation. Happy to flip it if a maintainer prefers the stricter reading — leaving the thread open for that.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #24170, opened 2026-09-10
fix(ui): don't take focus when the notification tray opens itself (LAZ-1065)

r1chm8:
Typing in a table filter or search box and having a notification arrive took the cursor out of the field. [LAZ-1065](https://linear.app/nexus-mods/issue/LAZ-1065/keep-text-fields-focused-when-popups-appear)

## Cause

The spine tray auto-opens by driving its own trigger — `buttonRef.current.click()` — and Headless UI's `PopoverButton` click handler ends by focusing the button it toggled (`(i = t.state.button) == null || i.focus()` in `popover.js`). So the bell took focus every time a notification arrived.


src/renderer/src/views/components/Spine/notifications/Notifications.tsx
```diff
@@ -23,6 +23,17 @@ import { useNotificationItems } from "./hooks/useNotificationItems.hook";
...
+const openWithoutTakingFocus = (button: HTMLButtonElement) => {
+  Object.defineProperty(button, "focus", { configurable: true, value: () => undefined });
+
+  try {
+    button.click();
+  } finally {
+    delete (button as unknown as { focus?: () => void }).focus;
+  }
+};
+
```
erri120, line 37:
```suggestion
type Focusable<T extends HTMLElement> = T & { focus?: () => undefined };

/** Opens the tray as its button would, minus the focus Headless UI's handler takes. */
const openWithoutTakingFocus = (button: Focusable<HTMLButtonElement>) => {
  Object.defineProperty(button, "focus", { configurable: true, value: () => undefined });

  try {
    button.click();
  } finally {
    delete button.focus;
  }
};

```

Keeping `Object.defineProperty` because it does have a different runtime characteristic than normal property assignment `button.focus = ...`

erri120 (changes requested):

src/renderer/src/views/components/Spine/notifications/Notifications.tsx
r1chm8: The suggest change was throwing a type error, I've used `Reflect.deleteProperty` instead of extending the `HTMLButtonElement` type.

erri120 (approved):

PR #24171, opened 2026-09-10
Steam discovery crash fix

Aragas:
Real fix was this

```
const code = getErrorMessageOrDefault(err);
return ["EPERM", "ENOENT"].includes(code)
    ? PromiseBB.resolve(steamPaths)
    : PromiseBB.reject(err);
```
to
```
const code = getErrorCode(err);
return code !== null && ["EPERM", "ENOENT"].includes(code)
    ? PromiseBB.resolve(steamPaths)
    : PromiseBB.reject(err);
```

Decided to switch the function to use proper async, converted `fs.readFileAsync` to `fsOG.readFile`, added proper type support for `simple-vdf`, replaced getSafeCI
Added tests to confirm that the changes are safe

Closes LAZ-1108


src/renderer/src/util/Steam.ts
```diff
@@ -210,57 +212,62 @@ class Steam implements IGameStore {
...
+      ++counter;
+      lib = asBlock(libObj[`${counter}`]);
+    }
+    log("debug", "found steam install folders", { steamPaths });
+    return steamPaths;
   }
 
   private parseManifests(): PromiseBB<ISteamEntry[]> {
-    return this.resolveSteamPaths().then((steamPaths: string[]) =>
+    return PromiseBB.resolve(this.resolveSteamPaths()).then((steamPaths: string[]) =>
```
erri120, line 270: If you remember from the presentation:

❌ `Bluebird.resolve(nativePromise);`
✔️ `Promise.resolve(bluebirdPromise);`

I'm currently mapping out a possible upgrade path for our game store handlers to convert all of them from Bluebird to native Promises.

Aragas: Good catch! The problem is, this change would trigger a big chain of changes further, so I decided not to do that here in the initial PR

Aragas: Do you want me to do it? I followed the principle of only refactoring functions that are part of the relevant change, not refactoring the whole class

erri120: I'd keep this PR focused on the bug fix, not the Bluebird to Promise conversion. I know it sucks to keep writing Bluebird but the game stores are too interwoven with other stuff.

Aragas: @erri120 anything else pending for this PR?

erri120: No, just change it back to Bluebird and only have the fix in the PR.

Aragas: I can't do it unless I'll get rid of the `async/await` code in `resolveSteamPaths`. Unless I'm missing something?

erri120: Yes, that's what I'm saying. Defer the native conversion for later, just have the fix in this PR. No native Promises or async/await.

Aragas: Are we sure about this? Because it looks like an overreaction to Bluebird stuff. We do want to improve the codebase which we touch usually, I think having a proper function wrapped in Bluebird.resolve gives more than it takes

Aragas: @erri120 restored the bb chains

Aragas: Sec, push failed xD

Aragas: Okay, done

erri120 (approved):

PR #24172, opened 2026-09-10
Feat/metadata assignment fix - Fixes Issue #21979

### LINKED ISSUE #21979: [BUG] - Incorrect Mod MetaData Assignment for Packages Downloaded from GitHub - Breaks Updates
**Your Nexus Mods username**
ChemBoy1

**Describe the bug**
Vortex metadata is not assigning correctly, and will find BepInEx and ConfigManager packages from other games on the site (or mods that have nothing to do with BepInEx at all) and assign them as the mod. This causes Vortex to check for updates from those mod pages, potentially installing incorrect mods.

In this example with Silksong, the BepInEx package (5.4.23.5, downloaded by modtype-bepinex extension) metadata is assigned from this package on another game domain (it is an unaltered 5.4.23.5 package): https://www.nexusmods.com/villagemerchant/mods/1. At least its the correct mod!

For the ConfigManger (downloaded by Silksong extension) the mod assigned has nothing to do with BepInEx at all. This is obviously more problematic.
https://www.nexusmods.com/hollowknightsilksong/mods/7 

I cannot see the file archive for that page, but perhaps there is an unaltered 18.4.1 ConfigManager in the archives. That would make more sense.

Note that this behavior does not change whether or not I pass an empty object in for `dlInfo `. Indicating the metadata is being fully handled by the Vortex app.

**To Reproduce**
1. Manage a Unity BepInEx game, like HK Silksong
2. The extension downloads BepInEx and the ConfigManager from GitHub
3. See that Vortex fetched metadata for a package from another game (BepInEx) or a completely unrelated mod from the same game domain (ConfigManager).
4. Running an update check will cause Vortex to download the incorrect mod for the update to the ConfigManger.

**Expected behavior**
Vortex should not assign the incorrect mod metadata.

**Screenshots**

Image

Image

Image

Image

 - OS: Windows 11 25H2
 - Vortex Version 1.16.7

**Additional context**
I thought this had been patched previously, but noticing it happening again.

ChemGuy1611:
Fixes #21979.

This PR provides a means to avoid the incorrect assignment of Nexus Mods mod metadata to mod packages downloaded from external sources. This is important because incorrect metadata assignment can lead to the user's mod environment being broken by updates that have nothing to do with the intended mod package (i.e. BepInEx).

The extensions modtype-bepinex and baldursgate3 were given a single-line update of `source: website` in the dlInfo that is required for Vortex to skip trying to assign metadata to the mod.

Tests are included, and changes were built and tested live in Vortex dev environment. 

## Problem

Every finished download is md5-matched against the Nexus meta database, and that lookup keys on
the hash alone — it is game-agnostic. Extensions that fetch third-party dependencies from GitHub
(BepInEx, ConfigurationManager, MelonLoader, UE4SS, …) get files that are byte-identical to copies
other authors re-uploaded to Nexus for _their_ games, so the lookup hits the wrong mod on the wrong
game. `metaLookupMatch` can't filter it out — the filename doesn't match, the game doesn't match,
and the last rung of the ladder is an unguarded `filtered[0]`.

That foreign identity is then written onto the download and the installed mod, and
`checkModVersion` runs an update check against it. Verified against live Nexus GraphQL:

| Dependency                   | md5 resolves to                       | Update check under `hollowknightsilksong` queries        |
| ---------------------------- | ------------------------------------- | -------------------------------------------------------- |
| ConfigurationManager v18.4.1 | `megastoresimulator/mods/7`           | `hollowknightsilksong/mods/7` → "No-HUD-No-Effects"      |
| BepInEx 5.4.23.3             | `mytimeatsandrock/mods/129` (+3 more) | `hollowknightsilksong/mods/129` → "bell beast save file" |

Accepting the offered update overwrites the dependency with an unrelated mod.

## Core changes

Three conservative changes. None discards the `meta` blob the lookup returns — description, author
and picture stay useful. They stop an unrelated mod's _identity_ being written onto a file and then
acted on.

**A — `queryDLInfo.ts`, `queryInfoInternal`: don't overwrite an extension-declared source.**
There is already a guard for real Nexus downloads (a `nexus.ids.fileId` that disagrees with the
lookup means bad server data), but it's a no-op for a GitHub download with no `fileId` to compare.
Added a sibling guard: if the download's `modInfo.source` is set and isn't `'nexus'`, return early —
skipping the `source: 'nexus'` stamp, the three `nexus.ids.*` writes, and the `set-download-games`
move. An extension that stated where a file came from outranks an md5 guess. `source: undefined`
still takes the normal path, so legacy downloads are unaffected.

**B — `processAttributes`: derive `downloadGame` from whatever supplied `modId`.**
The extractor already resolves `modId`/`fileId` from `nexus.ids ?? meta.details`, but `downloadGame`
came from a different extractor reading `download.game` — so the id and the game it is queried
against could describe different mods. New `downloadGameForNexusId` helper pins them together, and
branches on which source won:

- id from the download's own nexus record → that record's `gameId`, or `undefined` if it has none,
  which leaves `download.game` in place. Deliberately _not_ falling through to `meta.domainName`:
  an nxm-protocol resolve stamps `modId`/`fileId` without a `gameId` (this is how collection members
  are fetched), so a collided md5 would otherwise re-point a download whose ids are perfectly good.
- id from the md5 lookup → that same lookup's `domainName`, so the two agree.

**C — `checkModsVersion.ts`, `checkModVersion`: source gate on the update check.**
After the existing `isNaN(modId)` bail, return early when `attributes.source` is set and isn't
`'nexus'`. Not `=== 'nexus'` — mods predating the attribute must keep updating. The bulk path
already pre-filters on source, so this covers the single-mod callers: the Version column's per-row
check and the `fixIds` repair flow.

## Extension-side changes

Change A only helps once an extension states where a file came from. Two bundled extensions fetch a
dependency straight from GitHub and never declared a source, so the md5 collision hit them
unopposed:

| Extension           | Dependency        | Fetched from                 |
| ------------------- | ----------------- | ---------------------------- |
| `modtype-bepinex`   | BepInEx           | `github.com/BepInEx/BepInEx` |
| `game-baldursgate3` | LSLib/Divine Tool | `github.com/Norbyte/lslib`   |

Both now set `source: "website"` on the `start-download` payload. `'website'` is a registered mod
source (`mod_management/index.ts` registers `user-generated`, `website`, `unsupported`;
`nexus_integration` adds `nexus`) — it renders "Website" in the mod list's Source column and is the
value `InstallManager` writes for its own web downloads. `modtype-bepinex` already declares
`source: "nexus"` on its genuinely-Nexus route (`bepInExDownloader.ts`); this is the equivalent for
the GitHub route, extracted into a small `buildGithubDownloadInfo` helper so it can be unit-tested.

Only these two. The declaration helps only where an md5 hit can be nothing but the _wrong_ mod — a
dependency with no single canonical Nexus page, re-uploaded piecemeal across unrelated games.
`script-extender-installer` (SKSE/F4SE/etc. each have one canonical Nexus page — the md5 match there
is correct and drives a useful update check) and `modtype-umm` (Unity Mod Manager's primary route
is already a `nexus-download`) are deliberately left as they are.

Neither extension's own update path is affected: `checkForUpdates` in both queries the GitHub
releases API directly, so BepInEx and LSLib still update.

## Not included

Requiring corroboration before `metaLookupMatch` returns `filtered[0]` would also fix this, but it
changes behaviour for legitimately renamed archives — which is the case that rung exists for.

## Tests

Colocated vitest, following the existing `extractors.test.ts` pattern.

- `download_management/util/queryDLInfo.test.ts` (5) — drives the now-exported `queryInfoInternal`
  with a fake api: an extension-set source survives an md5 hit; `undefined` still gets stamped; the
  pre-existing fileId guard still holds; a collection's nexus dependency keeps its own identity.
- `nexus_integration/util/convertGameId.test.ts` (9) — both provenance branches of
  `downloadGameForNexusId`, including that a record with a `modId` but no `gameId` yields `undefined`
  rather than the md5 domain.
- `mod_management/util/metaLookupMatch.test.ts` (7) — characterization of the match ladder including
  the `filtered[0]` fallback, so any later hardening shows up as a diff here.
- `extensions/modtype-bepinex/src/downloadInfo.test.ts` (3, new — `modtype-bepinex` had no test
  infrastructure, so this stands up `vitest.config.ts` + a `test` script mirroring
  `extensions/feedback`): `buildGithubDownloadInfo` returns a `source` that is set and not
  `'nexus'` (the exact predicate change A gates on), is `'website'` specifically, and files the
  download under the requested game.

The behaviour-change tests were confirmed to fail against the unpatched code (the `modtype-bepinex`
one: 2 of 3 red with `source` removed from the builder).

`nx typecheck` / `lint` / `test` and `oxfmt --check` are clean on `@vortex/renderer` (every core
change) and on `modtype-bepinex` + `game-baldursgate3` (typecheck/build/test on the former, build/test
on the latter — it has no `typecheck` target).

## Versioning

Bumped `modtype-bepinex` 0.2.10 → 0.2.11 and `game-baldursgate3` 1.5.10 → 1.5.11, one CHANGELOG
bullet each. Both `CHANGELOG.md` files already trail `package.json`; drop the bumps if that does not
fit your release process.

## Note for maintainers

Change A is a cooperation point: once an extension states provenance, Vortex respects it over an
md5 guess. The two bundled extensions above are fixed here; non-bundled game extensions that pull a
shared dependency from GitHub (there are several) will want the same one-line `source` declaration.


Aragas:
Hey @ChemGuy1611!
We'll most likely check the PR no sooner than Monday!

Aragas:
Hey @ChemGuy1611!
We had an internal conversation regarding the issue.
We believe the main pain point is that Vortex assigns a Nexus mod from a different game domain, so we're going to filter md5 hits by the download's game domains before matching. Your approach with the source guard only protects extensions that opt in, this way it should be a universal fix.
We'll cherry-pick your `metaLookupMatch` and `queryDLInfo` tests though, with your authorship kept!
Here's the PR https://github.com/Nexus-Mods/Vortex/pull/24195

If you have any feedback regarding the approach, happy to continue the conversation!

IDCs:
Closing as https://github.com/Nexus-Mods/Vortex/pull/24195 supersedes this PR.

PR #24194, opened 2026-09-15
feat(ui): pin Deploy and Purge to the mods toolbar by default

r1chm8:
Pins Deploy and Purge to the mods toolbar by default. Only reaches users who never pinned or unpinned them.

[LAZ-1155](https://linear.app/nexus-mods/issue/LAZ-1155/add-deploy-and-purge-as-default-toolbar-buttons)


erri120 (dismissed):

r1chm8:
Updated this to point to master instead of the beta release

erri120 (approved):

PR #24196, opened 2026-09-15
load the plugins into libloot before every sort

IDCs:
Every plugin sort now runs libloot's pre-sort sequence itself, as the LOOT application does: refresh the load-order state, hold the game's main master headers-only, fully load every other plugin being sorted, then `SortPlugins`. Sorts no longer depend on a `plugin-details` pass having run first, so the Sort Now button, the ESL conversion, the health-test sort, the cycle-recovery re-sort, the VR extensions' own sorts and the sort deferred behind a deployment all stop failing with "Plugin not loaded". The `lootSortAsync` API drops its plugin-details pass. [ADR 0001](https://linear.app/nexus-mods/document/adr-0001-the-plugin-sort-loads-the-plugins-it-sorts-313577ff37b0) records the decision and its costs.

fixes [LAZ-1141](https://linear.app/nexus-mods/issue/LAZ-1141)


docs/adr/0001-plugin-sort-loads-its-plugins.md
erri120, line 1: Do we really need this? ADRs are usually for major decisions, things like "what coding language to use" less "why I changed this one method". I don't think this is the right place for the doc, maybe attached to the relevant tickets but not checked into source.

IDCs: Sure I'll take it out

Aragas: Counter point - this is an important change that extension developers might be interested in. Easier to have a doc than to search the code changes

IDCs: I attached the ADR to the linear ticket so we can reference it when needed. Git history will also hold the doc so it's still viewable by the public.

erri120 (approved):

PR #24206, opened 2026-09-16
tell the user which master a plugin is missing and why

IDCs:
The Missing Masters health check now says why each master is unavailable, in LOOT's words: `not-installed` and `not-enabled` are the user's install state and stay warnings, while `removed-externally`, `unscanned` and `not-deployed` once nothing is left to rescan mean the plugin list disagrees with the game folder, and are errors. Those cases go to telemetry once per master per state, with the mod and collection on both sides of the dependency, so a broken collection revision is identifiable from one span. A sort that fails on an unloaded master names it the same way, and the check cannot be hidden.

Stacked on #24196.

part of [LAZ-1030](https://linear.app/nexus-mods/issue/LAZ-1030)
part of [LAZ-1039](https://linear.app/nexus-mods/issue/LAZ-1039)


src/renderer/src/extensions/gamebryo_plugin_management/util/missingMasters.ts
```diff
@@ -0,0 +1,395 @@
...
+ * install state; removed-externally and unscanned are the plugin list disagreeing with the game
+ * folder, as is not-deployed once nothing is left to rescan.
+ */
+type MasterState =
+  | "not-installed"
+  | "not-deployed"
+  | "not-enabled"
+  | "removed-externally"
+  | "unscanned";
+
```
erri120, line 40: I strongly recommend using a const object + `typeof`:

```typescript
const MasterStates = {
  notInstalled: "not-installed",
  notDeployed: "not-deployed"
} as const;

type MasterState = (typeof MasterStates)[keyof typeof MasterStates]
```

This combines the runtime system with the type system for safety in both. Instead of doing something like `if (state === "not-installed")` you'd do `if (state === MasterStates.notInstalled)` to prevent accidental drifts and easier usage analysis.

Aragas: Out of curiosity, would that apply also to the error system? We have string-based stuff there

erri120: We can't build our error kind union from data because it maps strings to types and needs to be an open interface. But we can have constant values instead of literal strings.

The openness of the error system makes this a less than ideal choice. The const data + typeof combination works great for the situation we have here: a closed controlled set of values. You'll find similar in our path providers.

erri120 (approved):

PR #24214, opened 2026-09-16
Remove `IGameStore` API export

erri120:
Closes LAZ-1194.

This allows us to change `IGameStore` however we want without having to worry about extensions. No extension implements a custom store, nor should they, so `registerGameStore` is also gone.


Aragas (approved):
I don't necessarily like that we're removing the ability to provide a custom game store

PR #24215, opened 2026-09-16
load libloot's metadata lists when they change on disk

IDCs:
libloot holds the masterlist and userlist in memory and watches neither file, so a fresh masterlist never reached it and plugin details answered from the lists loaded at the last sort. One owner now decides whether the lists are current, used by the sort path, the details path, the masterlist download and init, and callers that just rewrote a list say so rather than trusting the file clock.

fixes [LAZ-1142](https://linear.app/nexus-mods/issue/LAZ-1142)


src/renderer/src/extensions/gamebryo_plugin_management/autosort.ts
```diff
@@ -435,7 +442,7 @@ class LootInterface {
       this.mExtensionApi.dismissNotification("loot-cycle-warning");
       const timeBefore = Date.now();
       store.dispatch(startActivity("plugins", "sorting"));
-      this.mSortPromise = this.readLists(gameMode, loot)
+      this.mSortPromise = Bluebird.resolve(this.ensureLists(gameMode, loot))
```
erri120: 🔫 no `Bluebird.resolve`

IDCs:
Err - I swear I thought I saw this as "approved" via the PR overview screen 🤦

PR #24228, opened 2026-09-17
test: clear the warnings the renderer suite prints

r1chm8:
CI logs carry warnings from the renderer suite that no test fails on and that none of us see locally. This clears all of them, and one turns out to be a real bug.

> Struck-through text below is what the first commits did and later commits replaced. Kept visible so the commit-by-commit read makes sense.

## `An update to ... inside a test was not wrapped in act(...)` — 37 across 7 files

`ToolsSection` renders `<Transition appear show>`. headless-ui drives the enter off `requestAnimationFrame` — `useTransition` calls `nextFrame(run)`, and `nextFrame` is `rAF(rAF(cb))` — so mounting one queues React state updates about three frames out.

Those tests are synchronous, so the frames cannot fire *during* a test; the stack never yields. They fire in the gap after the test body, before React Testing Library's auto-cleanup unmounts and cancels them. Locally that gap is shorter than a frame and cleanup always wins. On a loaded runner the timer is already due, the update lands with no `act()` around it, and it warns.

A shared `settleTransitions()` helper (`src/renderer/src/test-utils/transitions.ts`) runs the pending frames under `act()`, at the *end* of a test, after the assertions — settling completes the enter, which strips the enter-from classes some of these tests check.

~~It waits four animation frames: `run()` lands on the second via `nextFrame`, `done()` on the third, and the fourth is slack.~~ **Superseded in `822cc728e`.** headless-ui marks an element mid-transition with `data-transition` — the attribute it gives you to style against — so the helper waits for that to clear instead. Same effect, no arithmetic to justify:

```ts
export const settleTransitions = (): Promise<void> =>
  waitFor(() => {
    expect(document.querySelector("[data-transition]")).toBeNull();
  });
```

Tests that already `await` something (`userEvent`, `findBy*`) never needed it; the await flushes the frames under `act()` already. That is why, in `Modal.test.tsx`, only the three synchronous tests warned and the one with `await userEvent.click` did not.

| file | tests | source |
| --- | --- | --- |
| `ToolsSection.test.tsx` | ~~5~~ 3 | `Transition appear show` |
| `Modal.test.tsx` | 3 | `ModalWrapper`'s Transition |
| `EntryActions.test.tsx` | 2 | the reasons modal, plus its Tooltip |

Four more were the same warning from their own causes, each a bare call that updates state outside `act()`: `Tooltip.test.tsx` (`trigger.focus()`), `PopoverMenu.test.tsx` (two `.focus()` calls on pin buttons), `DownloadGraph.test.tsx` (the ResizeObserver callback invoked directly) and `NotificationButton.test.tsx` (a plain `setTimeout` wait).

~~All four wrap the offending call in `act()`.~~ Three do. `PopoverMenu` instead reaches the pin button with `await userEvent.click`, which leaves focus there and drops the `act` import from that file entirely — it is also how someone would actually land on that control. The other three have no user event to express: one is a keyboard-focus path a click would defeat, one is a ResizeObserver callback, one is a timer.

## Two ToolsSection tests deleted

`"holds the row back before it slides up"` and `"gives the row nothing to animate when motion is turned down"` asserted `delay-150`, `translate-y-6`, `reduce-motion:opacity-100` and friends — string literals in `ToolsSection.tsx`, read back. happy-dom runs no CSS, so nothing is observable; they could only fail on an edit already visible in the diff.

What is left covers what the component decides: the width-dependent classes, the keyed remount, and rendering nothing when the game has no tools.

## `react-i18next:: You will need to pass in an i18next instance` — 19 files

Nothing initialises i18next in tests, so react-i18next warns once per file and falls back to returning the key. The source strings in this repo *are* the keys, so that fallback is already what the assertions read. `test-setup.ts` now states it rather than leaving it as a side effect, and `useGameCardActions.test.tsx` drops its own copy of the same stub.

The stub keeps react-i18next's real return shape — an array that also carries `t`/`i18n`/`ready` as properties. The repo uses **both** `const { t } = useTranslation()` and `const [t] = useTranslation()` (`AddToolButton`, `FilterBox`, `Instructions`, `DuplicatesDialog`), and a plain `{ t }` stub would break the array form with "not iterable" the first time someone wrote a test for one of those.

## `Headless UI has polyfilled Element.prototype.getAnimations`

happy-dom has no Web Animations API, and `Transition` asks an element for its running animations to know a transition has finished. Left alone headless-ui installs its own polyfill and says so. `test-setup.ts` defines the same empty answer first, next to the existing `matchMedia` stub, so behaviour is unchanged and the notice goes away.

This one only surfaces *because* of the `act` fix above — until transitions ran to completion, nothing reached `getAnimations`.

## `[re-reselect] Invalid cache key "undefined"` — a real bug

`lastActiveProfileForGame` keyed its cache on the `gameId` it was handed:

```ts
)((state, gameId) => gameId);
```

Callers pass `activeGameId`, which is `undefined` whenever no game is active — `useModToolbarActions.hook.tsx:155`, `InstallManager.ts:597` and ~26 other sites. re-reselect refuses an undefined key, warns, and **skips memoisation for that call**. That happens in the running app, not just under a test fixture.

Now `gameId ?? "___empty"`, matching the guard `enabledModCountForProfile` directly above it already uses.

Three siblings have the same shape and are left alone for now — `needToDeployForGame`, `gameById`, `discoveryByGame` — plus `installPathForGame`, whose key selector *explicitly* returns `undefined` for an undefined gameId, which reads like someone expected that to skip the cache when it is exactly what triggers the warning.

## Verification

239 files / 2616 tests pass (~~2618~~, two deleted), with all four warning classes at zero. Run it with `--reporter=verbose --silent=false` — the default reporter hides console output, which is why these went unnoticed for so long. `tsc --noEmit` and oxlint clean.

The `getAnimations` stub was checked with a negative control: commenting out that one line brings the warning back.

Only one production file changes (`profile_management/selectors.ts`); the rest is test scaffolding.


src/renderer/src/test-utils/transitions.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Tests that already await something (userEvent, findBy*) don't need it — the await flushes the
+ * frames under act() already.
+ *
+ * Test-only: nothing in the production tree imports this module.
+ */
+import { act } from "@testing-library/react";
+
+export const settleTransitions = (): Promise<void> =>
+  act(async () => {
+    // enter is queued behind nextFrame(), which is two frames, then one more before it completes.
```
erri120: Source?

r1chm8: `@headlessui/react` 2.2.10, two files in the installed package.

`dist/utils/disposables.js` — `nextFrame` is a rAF nested inside a rAF:

```js
nextFrame(...e){ return r.requestAnimationFrame(() => r.requestAnimationFrame(...e)) }
```

`dist/hooks/use-transition.js` — the enter is started through it, with one more rAF around the completion check:

```js
function C(e, {prepare: r, run: t, done: n, inFlight: i}) {
  let a = m();
  return j(e, {prepare: r, inFlight: i}),      // prepare(), synchronous
    a.nextFrame(() => {
      t();                                      // run()   → frame 2
      a.requestAnimationFrame(() => {
        a.add(M(e, n))                          // done()  → frame 3
      })
    }), a.dispose
}
```

So `prepare()` is synchronous in the layout effect, `run()` lands two frames in via `nextFrame`, and `done()` one frame after that — three frames total. The loop awaits four, which is one frame of slack.

`done()` fires on that third frame rather than later because `M` short-circuits when the element reports no animations:

```js
let i = e.getAnimations?.().filter(o => o instanceof CSSTransition) ?? [];
return i.length === 0 ? (r(), t.dispose) : (Promise.allSettled(...).then(...), t.dispose)
```

Under happy-dom there is no Web Animations API, so that list is always empty and `done()` runs synchronously.

Unminified upstream, same code: `packages/@headlessui-react/src/hooks/use-transition.ts` and `src/utils/disposables.ts` in tailwindlabs/headlessui.

Worth saying that the count isn't load-bearing: too few frames and the warning comes back, which is exactly what this PR is fixing, so it fails loudly. Too many just costs a frame. It's a lower bound with margin rather than a magic number.

I've put the citation in the comment itself in 2db0daea4 so the next person doesn't have to ask.

r1chm8: Removed this now for wait on the data property.

erri120 (approved):

PR #24229, opened 2026-09-17
Clean up the `GameStoreHelper` API surface

erri120:
Closes LAZ-1205.

See the ticket for details, this removes zero-usage members from the public API for easier migration.


Aragas (approved):
Api file update btw?

PR #24230, opened 2026-09-17
test: quiet the noise the renderer suite prints

r1chm8:
Three unrelated warnings account for every stderr line the renderer suite prints that isn't a test's own output. None fails a test; all of them make a CI log harder to read, and one is a real bug.

## `react-i18next:: You will need to pass in an i18next instance` — 19 files

Nothing initialises i18next in tests, so react-i18next warns once per file and falls back to returning the key. The source strings in this repo *are* the keys, so that fallback is already what the assertions read. `test-setup.ts` now states it rather than leaving it as a side effect.

Four test files carried their own copy of the same stub and drop it (`AddedGames`, `SupportedGames` and `DetectedGames` do too, but those files only exist on the LAZ-1112 branch, so that part travels with it).

Note the stub keeps react-i18next's real return shape — an array that also carries `t`/`i18n`/`ready` as properties. The repo uses **both** `const { t } = useTranslation()` and `const [t] = useTranslation()` (`AddToolButton`, `FilterBox`, `Instructions`, `DuplicatesDialog`), and a plain `{ t }` stub would break the array form with "not iterable" the first time someone wrote a test for one of those.

## `Headless UI has polyfilled Element.prototype.getAnimations`

happy-dom has no Web Animations API, and headless-ui's `Transition` asks an element for its running animations to know a transition has finished. Left alone it installs its own polyfill and says so. `test-setup.ts` defines the same empty answer first, next to the existing `matchMedia` stub, so behaviour is unchanged and the notice goes away.

This one is latent on `master`: it only fires once a transition actually runs to completion, which is what #24228 introduces. Harmless here either way.

## `[re-reselect] Invalid cache key "undefined"` — a real bug

`lastActiveProfileForGame` keyed its cache on the `gameId` it was handed:

```ts
)((state, gameId) => gameId);
```

Callers pass `activeGameId`, which is `undefined` whenever no game is active — `useModToolbarActions.hook.tsx:155`, `InstallManager.ts:597` and ~26 other sites. re-reselect refuses an undefined key, warns, and **skips memoisation for that call**. That happens in the running app, not just under a test fixture.

Now `gameId ?? "___empty"`, matching the guard `enabledModCountForProfile` directly above it already uses.

Three siblings have the same shape and are left alone for now — `needToDeployForGame`, `gameById`, `discoveryByGame` — plus `installPathForGame`, whose key selector *explicitly* returns `undefined` for an undefined gameId, which reads like someone expected that to skip the cache when it is exactly what triggers the warning.

## Verification

239 files / 2618 tests pass, with all four warning classes at zero (`--reporter=verbose --silent=false`; the default reporter hides them, which is why these went unnoticed). `tsc --noEmit` and oxlint clean.

The `getAnimations` stub was checked with a negative control — commenting out that one line brings the warning back.


r1chm8:
Folded into #24228 as a third commit — it is the same cleanup and the `getAnimations` stub only has an effect once that PR's changes land.

PR #24251, opened 2026-09-21
Add a support bundle to the Help menu

insomnious:
Close LAZ-1062

Adds "Create support bundle" to the Help menu. It packs the logs, the current app state and any recent crash dumps into a single `.7z`, so instead of walking someone through three different folders we can just ask them for one file.

The dialog builds the archive as soon as it opens, then points at the forums or the Vortex channel on Discord, and gives them a button to show the file in its folder. Nothing's uploaded anywhere yet. The dialog will likely grow a form later, so there's a versioned `bundle.json` in the archive for that to hang off.

A few things worth a look:

- `prepareSupportBundle.ts` does the actual work. node-7z has some sharp edges: `mx` has to be a string or it throws, the progress callback's fourth argument is the kill function, and `add` resolves instead of rejecting when 7-Zip fails.
- `redact.ts` swaps Windows usernames in paths for `<USER>` and blanks the access tokens on signed download links. Hosts, filenames and expiry times stay, since that's the bit that tells you what actually failed. The account email comes out of the state export too, in case one of these ends up attached to a public post.
- `store.ts` gains `serializeState`, pulled out of `createFullStateBackup`. The Settings backup behaves exactly as before; the bundle just asks for `session` on top of the usual hives.

`pnpm run verify` passes. I've pulled apart a couple of real bundles looking for anything sensitive and came up clean, and the new event is landing in the dev Mixpanel project. It hasn't been run from a packaged build yet, so that's worth doing before this ships.


Aragas (approved):

insomnious:
Some shared sanitization utils should be created so that telemetry and this don't have code that is similar but not quite the same. Backlog ticket is LAZ-1230

PR #24262, opened 2026-09-22
Handle API rate limits instead of surfacing them

Aragas:
added universal isRateLimited to correctly handle 429 across our API
a rate-limited membership read now waits the ordinary cadence rather than the 30s share the login's account read at startup instead of fetching it twice


IDCs:
`util/checkModsVersion.ts:387` and `nxmProtocol.ts:529` were not converted to use `isRateLimited`/`notifyRateLimited` - not a big deal but you can potentially get double notifications for the same error.

IDCs (approved):

PR #24266, opened 2026-09-22
Fixed package flow

Aragas:
**Fixes:** Package workflow failing on `release/v2.8` ([run](https://github.com/Nexus-Mods/Vortex/actions/runs/35732975023/job/106762957840))

**What broke:** the rebuild step switched from `npx electron-rebuild` to `pnpm exec electron-rebuild`. pnpm 11 defaults `verifyDepsBeforeRun` to `install`, so `pnpm exec` runs `pnpm install` before the binary. That install happens in `src/main/dist`, a `pnpm deploy` artifact that can't be installed into:

- deploy's lockfile carries all 261 catalog entries, deploy's `pnpm-workspace.yaml` carries none → `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` under CI's frozen install
- outside CI it fails instead on deploy's peer-suffixed specs (`@electron/rebuild@4.1.0(supports-color@10.2.2)`)
- `Build Installer` and `Validate Package Creation` never ran

**Fix:** `PNPM_CONFIG_VERIFY_DEPS_BEFORE_RUN: "false"` on that one workflow step. Deploy-only config, scoped to the deploy step, nothing written into the workspace.

**Considered and dropped:**

- writing `verifyDepsBeforeRun: false` into the deployed `pnpm-workspace.yaml` — puts deploy-only config in the source tree
- moving the call into a `@vortex/main` script — the outer `pnpm run` triggers the root deps check, which does a full install pass and re-runs `src/main postinstall` (a dev-tree electron-rebuild) before it ever reaches `dist`

**Not affected:** `package` / `package:nosign` — pnpm sets `pnpm_config_verify_deps_before_run=false` for anything spawned from a pnpm script, so their nested `cd ./dist && pnpm electron-builder` never checked.

**Verified:** in `src/main/dist`, `pnpm exec node -v` fails with the catalogs error; with the env var set it prints the version.


src/main/prepare-dist-package.mjs
```diff
@@ -8,6 +8,7 @@ const MAIN_DIR = resolve(import.meta.dirname);
 const MAIN_PACKAGE_PATH = resolve(MAIN_DIR, "package.json");
 const DIST_DIR = resolve(MAIN_DIR, "build");
 const DIST_PACKAGE_PATH = resolve(DIST_DIR, "package.json");
+const WORKSPACE_PATH = resolve(MAIN_DIR, "pnpm-workspace.yaml");
```
erri120: This would be `src/main/pnpm-workspace.yaml` which makes this script can't be run locally without messing everything up and requiring you to manually clean up the file.

erri120 (changes requested):

src/main/prepare-dist-package.mjs
Aragas: I really don't like this workaround, but how about this?

erri120 (approved):

